PR #1, opened 2020-12-23
Add detect/build API, examples

Malax:
This is a quick port of an API for `bin/detect` and `bin/build` based on my libcnb prototype. It compiles against the new data types and I removed more advanced features like layer handling for now. This PR also includes an example directory with one buildpack that uses the preliminary API.

There a multiple issues with the code here, there are `unwraps`, placeholder variable names and there is no good reasoning behind the module structure yet. This PR is intended as a conversation starter only.


src/detect.rs
```diff
@@ -0,0 +1,75 @@
+use crate::data::build_plan::BuildPlan;
+use crate::shared::{write_toml_file, BuildFromPath, Platform};
+use std::env;
+use std::path::PathBuf;
+use std::process;
+
+pub fn cnb_runtime_detect<P: Platform + BuildFromPath>(
+    detect_fn: impl Fn(DetectContext<P>) -> DetectResult,
```
hone, line 8: This never actually returns a `DetectResult` since this method will always `process::exit`. I keep going back and forth on this, but I wonder if we should return the enum instead of process exit.

PR #2, opened 2020-12-27
Overhaul platform, build/detect API

Malax:
- `Platform` now resides in its own file and was overhauled to make implementation of new `Platform`s easier. 
- Introduced `GenericDetectContext` and `GenericBuildContext` for the default use-case. This removes generics for the standard use-case.
- `BuildFromPath` has been removed since we don't need the additional complexity it adds.
- `platform.rs` now has some rustdoc
- Streamlined `detect` and `build` function API, both return a `Result<_, BuildpackError>` now
- Introduced `BuildpackError` trait with implementation for all `std::err::Error` traits. We can add support for error handling frameworks by adding trait implementations for those frameworks.

### DIFF
--- a/examples/example-01-basics/src/bin/bp_detect.rs
+++ b/examples/example-01-basics/src/bin/bp_detect.rs
@@ -1,16 +1,17 @@
 use libcnb;
-use libcnb::detect::{DetectResult, DetectContext};
 use libcnb::data::build_plan::BuildPlan;
+use libcnb::detect::DetectOutcome;
+use libcnb::detect::GenericDetectContext;
+use libcnb::platform::Platform;
 use libcnb::shared;
-use libcnb::shared::GenericPlatform;
-use libcnb::shared::Platform;
+use std::error::Error;
 
 fn main() {
     libcnb::detect::cnb_runtime_detect(detect)
 }
 
-fn detect(_context: DetectContext<GenericPlatform>) -> DetectResult {
+fn detect(_context: GenericDetectContext) -> Result<DetectOutcome, std::io::Error> {
     let mut buildplan = BuildPlan::new();
 
-    DetectResult::Pass(buildplan)
+    Ok(DetectOutcome::Pass(buildplan))
 }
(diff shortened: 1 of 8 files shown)


examples/example-01-basics/src/bin/bp_detect.rs
```diff
@@ -1,16 +1,17 @@
...
-use libcnb::shared::GenericPlatform;
-use libcnb::shared::Platform;
+use std::error::Error;
 
 fn main() {
     libcnb::detect::cnb_runtime_detect(detect)
 }
 
-fn detect(_context: DetectContext<GenericPlatform>) -> DetectResult {
+fn detect(_context: GenericDetectContext) -> Result<DetectOutcome, std::io::Error> {
```
hone, line 13: Should we return our own Error type here?

src/platform.rs
```diff
@@ -0,0 +1,98 @@
...
+        })
+    }
+}
+
+/// Provides access to platform environment variables.
+pub struct PlatformEnv {
+    vars: HashMap<OsString, String>,
+}
+
+impl PlatformEnv {
```
hone, line 56: should we implement a `fn vars` or do you think people won't want an iterator?

hone (approved):

src/platform.rs
Malax: I think I stopped there since I couldn't properly emulate the `std::env` API and we discussed we wanted that. I'll leave this unimplemented for now.

examples/example-01-basics/src/bin/bp_detect.rs
Malax: It is a custom error type defined here: https://github.com/Malax/libcnb/pull/2/files#diff-c11e72174294250d95d68429cbb5e9ec58c16ac64087d428736be92bd22cd688R24-R26

I wanted it to be flexible so buildpack authors don't need to know anything about our error types. We can also implement the `BuildpackError` for other Error libraries if we need to. Using `io::Error` here was basically just a test if that works as I expected.

Do you want me to change the example?

hone: nah, I don't care about custom error types on the example. I expect a buildpack author to be "sloppy" and do convenience things to start.

PR #3, opened 2020-12-30
add base layer.rs

hone:
Add the start to a `Layer` struct with core functionality. We'll still want to add an `add_layer` fn to `Build`.

### DIFF
--- a/src/data/layer.rs
+++ b/src/data/layer.rs
@@ -12,3 +12,14 @@ pub struct Layer {
     pub cache: bool,
     pub metadata: Table,
 }
+
+impl Layer {
+    pub fn new() -> Self {
+        Layer {
+            launch: defaults::r#false(),
+            build: defaults::r#false(),
+            cache: defaults::r#false(),
+            metadata: Table::new(),
+        }
+    }
+}
--- a/src/data/layer.rs
+++ b/src/data/layer.rs
@@ -16,9 +16,9 @@ pub struct Layer {
 impl Layer {
     pub fn new() -> Self {
         Layer {
-            launch: defaults::r#false(),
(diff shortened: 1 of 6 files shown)


src/data/layer.rs
```diff
@@ -12,3 +12,14 @@ pub struct Layer {
     pub cache: bool,
     pub metadata: Table,
 }
+
+impl Layer {
+    pub fn new() -> Self {
+        Layer {
+            launch: defaults::r#false(),
+            build: defaults::r#false(),
+            cache: defaults::r#false(),
```
Malax, line 21: Shouldn't we just use `bool` literals there? I think we have those `fn`'s for serde which does not support defaults from literals.

src/error.rs
```diff
@@ -12,4 +12,10 @@ pub enum Error {
         "Found `{0}` but value MUST only contain numbers, letters, and the characters ., /, and -."
     )]
     InvalidStackId(String),
+    #[error("could not serialize into TOML")]
+    TomlSerError(#[from] toml::ser::Error),
+    #[error("could not deserialize from TOML")]
+    TomlDeError(#[from] toml::de::Error),
+    #[error("I/O Error")]
+    IoError(#[from] std::io::Error),
```
Malax, line 20: Will this make the `Display` of that `io::Error` appear somewhere? If not, that would complicate debugging.

Malax (approved):
👍🏻 , especially for all the rustdoc :)

src/data/layer.rs
hone: yeah, I was thinking about that. I picked this in case we ever change the default but happy to switch to false.

PR #4, opened 2020-12-30
Tidying up the Build

hone:
Removed all the warnings from `cargo build`.

* standardize `use` with `{}`
* remove unused imports
* `read_toml_file` and `write_toml_file` now use custom error type (also clean up `unwrap`)
* add accessors on `DetectContext`
* add `layer` fn on `BuildContext`

### DIFF
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -1,15 +1,16 @@
+use crate::Error as CrateError;
 use serde::{de::DeserializeOwned, Serialize};
-use std::{error::Error, fmt::Display, fs, io, path::Path};
+use std::{error::Error, fmt::Display, fs, path::Path};
 
-pub fn write_toml_file(value: &impl Serialize, path: impl AsRef<Path>) -> io::Result<()> {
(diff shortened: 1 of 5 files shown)


src/shared.rs
```diff
@@ -1,24 +1,18 @@
...
-use std::fs;
-use std::io;
-use std::path::{Iter, Path};
+use crate::Error as CrateError;
+use serde::{de::DeserializeOwned, Serialize};
+use std::{error::Error, fmt::Display, fs, path::Path};
 
-pub fn write_toml_file(value: &impl Serialize, path: impl AsRef<Path>) -> io::Result<()> {
+pub fn write_toml_file(value: &impl Serialize, path: impl AsRef<Path>) -> Result<(), CrateError> {
     // TODO: Fix Result type, remove unwrap
```
Malax, line 6: It looks like those TODO comments can go now. :)

Malax (approved):

src/shared.rs
hone: good catch!

PR #5, opened 2020-12-31
ruby sample buildpack from buildpacks docs

hone:
Here's my WIP spike of adding the ruby samples from the [buildpacks.io docs](https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/)


hone:
Changing `cnb_runtime_detect` and `cnb_runtime_build` to `fmt::Display` let me use both `std::io::Error` and `anyhow` in my example. Since we are only doing `eprintln!` we only depend on `display()` being implemented.

Malax (approved):

hone:
Sadly, I'm still trying to get it working with pack. :( Running into
buildpack plan TOML parsing issues.

hone:
@Malax I've now updated this to implement all the features from https://buildpacks.io/docs/buildpack-author-guide/create-buildpack/caching/

Malax (approved):

PR #22, opened 2021-05-05
add ability to write to content metadata with a closure

hone:


hone:
I couldn't think of a good fn name, so appreciate any bikeshed comments. :) I want this for the jvm-invoker-buildpack. The use case is you want to validate metadata first before writing a new one.

Malax:
I'm unsure about the naming. Since the metadata is passed as a mutable reference (which seems to be the Rust'y way to do it), a classic `map_metadata` naming is not fitting since I'd assume functor semantics which don't hold here. We should either research for prior art wrt to naming such functions in the Rust ecosystem, or merge this as-is to unblock `jvm-invoker-buildpack` for now.

We're pre-1.0 and can break API if we want to later. My 💯  bikesheddy comment would be: "What about `modify_content_metadata`?

Malax (approved):

hone:
I found some [Rust naming conventions](https://rust-lang.github.io/api-guidelines/naming.html) but sadly it doesn't cover this use case. It looks like the `..with..` is used in the [`FnOnce`](https://doc.rust-lang.org/std/ops/trait.FnOnce.html) documentation though.

PR #28, opened 2021-06-23
0.2.0

Malax:
Ongoing work on a `0.2.0` version which is more opinionated, contains various QOL changes, error handling improvements and better typing.


src/layer_lifecycle.rs
```diff
@@ -0,0 +1,362 @@
...
+        #[allow(unused_variables)] layer_path: &Path,
+        #[allow(unused_variables)] layer_content_metadata: LayerContentMetadata<LM>,
+    ) -> Result<O, E> {
+        Ok(O::default())
+    }
+
+    fn on_lifecycle_start(&self) {}
+    fn on_keep(&self) {}
+    fn on_update(&self) {}
+    fn on_create(&self) {}
```
edmorley, line 95: The `on_update` and `on_create` hooks here seem like they will always be called at the same time as `create` and `update`, thereby duplicating them. I'm presuming the reason these exist is because `on_keep` doesn't have an equivalent lifecycle method, and it would be inconsistent to just have `on_keep` and not the others? Or am I missing additional use-cases why `on_update` and `on_create` exist? :-)

src/layer_lifecycle.rs
```diff
@@ -0,0 +1,362 @@
...
+    #[error("Could not replace metadata after invalid_metadata_recovery: {0}")]
+    MetadataReplaceFailed(TomlFileError),
+
+    #[error("Could not read untyped metadata for invalid_metadata_recovery: {0}")]
+    CouldNotReadUntypedLayerMetadata(TomlFileError),
+
+    #[error("Could not write layer content metadata after creating new layer: {0}")]
+    CouldNotWriteLayerContentMetadataAfterCreate(TomlFileError),
+
+    #[error("Could create layer directory before layer lifecycle create: {0}")]
```
edmorley, line 140:
```suggestion
    #[error("Could not create layer directory before layer lifecycle create: {0}")]
```

src/platform.rs
```diff
@@ -27,28 +27,12 @@ where
     /// ```no_run
     ///use libcnb::platform::Platform;
     ///use libcnb::platform::GenericPlatform;
+    /// use libcnb::generic::GenericPlatform;
```
edmorley, line 30: It seems like the `libcnb::platform` lines can now be removed? (The indentation is also inconsistent)

edmorley (commented):
Thank you for working on this! I only had a quick look through but thought I'd comment now with the things I noticed, before I forgot :-)

src/platform.rs
Malax: Yeah, this has been removed. :)

src/layer_lifecycle.rs
Malax: Fixed, thanks!

src/layer_lifecycle.rs
```diff
@@ -0,0 +1,367 @@
...
+        #[allow(unused_variables)] layer_content_metadata: LayerContentMetadata<LM>,
+        #[allow(unused_variables)] build_context: &BuildContext<P, BM>,
+    ) -> Result<LayerContentMetadata<LM>, E> {
+        // Default implementation is a no-op
+        Ok(layer_content_metadata)
+    }
+
+    fn layer_lifecycle_data(
+        &self,
+        #[allow(unused_variables)] layer_path: &Path,
```
hone, line 86:
```suggestion
        _layer_path: &Path,
```

src/layer_lifecycle.rs
```diff
@@ -0,0 +1,367 @@
...
+        #[allow(unused_variables)] layer_content_metadata: LayerContentMetadata<LM>,
```
hone, line 87:
```suggestion
        _layer_content_metadata: LayerContentMetadata<LM>,
```

src/layer_lifecycle.rs
Malax: I use the attributes here to ensure that both the docs and generated code from IDEs (or other tooling) does not create parameters prefixed with `_`.

src/layer_lifecycle.rs
Malax: I use the attributes here to ensure that both the docs and generated code from IDEs (or other tooling) does not create parameters prefixed with _.

hone (approved):
I have a bunch of stuff that's hard to capture in single section changes. Let's merge this for now and wait on release `0.2.0`. I'll submit some PR or we can work through some stuff. The areas I want us to look into:

* The number of generic arguments for `LayerLifecycle`. There's 5 and I think it's a lot of cognitive overhead there. I wonder if we can use things like associated types to make this easier.
* We should loosen the trait restrictions on the Error for these closures except maybe for `GenericErrorHandler`.
* For some of these `Generic*` things, I wonder if we can just have default generic arguments.
* We should be exposing `toml::value::Table` if we expect users to use this as part of the crate.
* Naming should be a bit more standardized. We have `DetectOutcome` as an enum but then `ValidationResult`.

src/layer_lifecycle.rs
edmorley: @Malax ^

edmorley:
> I have a bunch of stuff that's hard to capture in single section changes.

Are any of these suggested changed things we still want to do? Would be good to file issues for them if so, otherwise this review feedback is likely to get forgotten :-)

edmorley:
> I have a bunch of stuff that's hard to capture in single section changes. Let's merge this for now and wait on releasing `0.2.0`. I'll submit some PRs and/or we can work through some stuff. The areas I want us to look into:

I've filed #68, #69, #70, #71 and #72 to track these five items.

@hone Could you check those issues accurately capture the intent of the suggestions (I had to improvise on some of the issue titles)? Also, if there was any other feedback not captured above, please could you open additional issues? :-)

PR #30, opened 2021-07-07
Add higher level environment handling

Malax:
Adds higher level environment handling, decoupling layer environment variables from the filesystem. This is the first step towards a declarative API for layer environment variables where buildpack authors don't have to write environment variables to disk, greatly improving testability due to no side-effects.

In a nutshell, this implements the layer environment variable handling that is done in the CNB lifecycle at runtime. I ported over the relevant tests from the CNB lifecycle source to ensure testing parity.

This will be integrated into layer lifecycles in a dedicated PR (let's face it, this PR is already too large).

Please refer to the included RustDoc for an overview how this works and is used. I copied the RustDoc for `LayerEnv` below:

# LayerEnv RustDoc
Represents environment variable modifications of a Cloud Native Buildpack layer.

Cloud Native Buildpacks can add a special directory to their layer directories to modify the
environment of subsequent buildpacks, the running container or specific processes at launch.
The rules for these modifications are described in the [relevant section of the specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#provided-by-the-buildpacks).

This type decouples this information from the file system, providing a type-safe in-memory
representation of the environment delta that is specified in the `env/*` directories of a layer.
Using this type, libcnb can provide declarative APIs that enable buildpack authors to easily
test their layer environment variable logic since they no longer write them to disk manually.

One use-case are environment variables that are modified by a layer that are required by the
same buildpack in later stages of the build process. For example, a buildpack might install a
build tool (i.e. Apache Maven) in one layer and adding the main binary to `PATH` via the `env`
directory of that layer. The same buildpack then wants to execute Maven to download dependencies
to a different layer. By using `LayerEnv`, the buildpack can encode these changes in a
type and, in addition to passing it to libcnb which will persist it to disk, pass it to the
logic that uses the build tool to download dependencies. The download process does not need to
know the layer name or any logic how to construct `PATH`.

## Applying the delta
`LayerEnv` is not a static set of environment variables, but a delta. Layers can modify existing
variables by appending or prepending or setting new ones only conditionally. If you only need a
static set of environment variables, see [`Env`].

To apply a `LayerEnv` delta to a given `Env`, use [`LayerEnv::apply`] like so:
```
use libcnb::layer_env::{LayerEnv, TargetLifecycle, ModificationBehavior};
use libcnb::Env;

let mut layer_env = LayerEnv::empty();
layer_env.insert(TargetLifecycle::All, ModificationBehavior::Append, "VAR", "bar");
layer_env.insert(TargetLifecycle::All, ModificationBehavior::Default, "VAR2", "default");

let mut env = Env::empty();
env.insert("VAR", "foo");
env.insert("VAR2", "previous-value");

let modified_env = layer_env.apply(TargetLifecycle::Build, &env);
assert_eq!(modified_env.get("VAR").unwrap(), "foobar");
assert_eq!(modified_env.get("VAR2").unwrap(), "previous-value");
```

## Implicit Entries
Some directories in a layer directory are be implicitly added to the layer environment if they
exist. The prime example for this behaviour is the `bin` directory. If it exists, its path will
be automatically appended to `PATH` using the operating systems path delimiter as the delimiter.

A full list of these special directories can be found in the
[Cloud Native Buildpack specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#layer-paths).

libcnb supports these, including all precedence and lifecycle rules, when a `LayerEnv` is read
from disk:
```
use libcnb::layer_env::{LayerEnv, TargetLifecycle};
use tempfile::tempdir;
use libcnb::Env;
use std::fs;

// Create a bogus layer directory
let temp_dir = tempdir().unwrap();
let layer_dir = temp_dir.path();
fs::create_dir_all(layer_dir.join("bin")).unwrap();
fs::create_dir_all(layer_dir.join("include")).unwrap();

let layer_env = LayerEnv::read_from_layer_dir(&layer_dir).unwrap();

let env = Env::empty();
let modified_env = layer_env.apply(TargetLifecycle::Launch, &env);

assert_eq!(modified_env.get("PATH").unwrap(), layer_dir.join("bin"));
assert_eq!(modified_env.get("CPATH"), None); // None, because CPATH is only added during build
```


schneems:
Gus-W-9528533

src/env.rs
```diff
@@ -0,0 +1,115 @@
+use std::collections::HashMap;
+use std::env;
+use std::env::VarsOs;
+use std::ffi::{OsStr, OsString};
+/// Generic collection of environment variables.
```
edmorley, line 5:
```suggestion

/// Generic collection of environment variables.
```

src/env.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+mod test {
+    #[test]
+    #[cfg(target_family = "unix")]
+    fn test_into_iterator() {
```
edmorley, line 95: Is this test name correct? It doesn't seem to be testing `iter()` / `IntoIterator` related functionality?

src/layer_env.rs
```diff
@@ -0,0 +1,528 @@
...
+            // This should include parsing the contents with an assumed charset and later emitting
+            // the raw bytes of that encoding as it might change the actual data. Since this is not
+            // explicitly written in the spec, we read through the the reference implementation and
+            // determined that it also treats the file contents as raw bytes.
+            // See: https://github.com/buildpacks/lifecycle/blob/a7428a55c2a14d8a37e84285b95dc63192e3264e/env/env.go#L73-L106
+            use std::os::unix::ffi::OsStringExt;
+            let file_contents = OsString::from_vec(fs::read(&path)?);
+
+            if let Some(file_name_stem) = file_name_stem {
+                let r#type = match file_name_extension {
```
edmorley, line 116: The use of `r#type` confused me for a moment, until I wondered whether it was due to `type` being a reserved keyword, and thus I found the raw identifier syntax here:
https://doc.rust-lang.org/reference/identifiers.html

Should we use a different variable name to avoid this confusing others too? eg `behaviour_type` ?

src/layer_env.rs
```diff
@@ -0,0 +1,528 @@
...
+                    }
+                    Some(file_name_extension) => match file_name_extension.to_str() {
+                        Some("append") => Some(ModificationBehavior::Append),
+                        Some("default") => Some(ModificationBehavior::Default),
+                        Some("delim") => Some(ModificationBehavior::Delimiter),
+                        Some("override") => Some(ModificationBehavior::Override),
+                        Some("prepend") => Some(ModificationBehavior::Prepend),
+                        // Note: This IS NOT the case where we have no extension. This handles
+                        // the case of an unknown or non-UTF-8 extension.
+                        Some(_) | None => None,
```
edmorley, line 130: When do we think this "unknown extension" scenario can occur? libcnb.rs will only be parsing the env dirs created by the current buildpack invocation, right? ie: Both earlier buildpacks in the build (whose values get set in in the system env by lifecycle) and cached previous builds won't make any difference? Which would mean this function is only ever having to parse values it wrote itself, unless someone is manually editing the env dir before calling this libcnb function?

If so, then does returning an Error make more sense here? (Given this isn't going to affect backwards compatibility or anything)

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// Cloud Native Buildpacks can add a special directory to their layer directories to modify the
+/// environment of subsequent buildpacks, the running container or specific processes at launch.
+/// The rules for these modifications are described in the [relevant section of the specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#provided-by-the-buildpacks).
+///
+/// This type decouples this information from the file system, providing a type-safe in-memory
+/// representation of the environment delta that is specified in the `env/*` directories of a layer.
+/// Using this type, libcnb can provide declarative APIs that enable buildpack authors to easily
+/// test their layer environment variable logic since they no longer write them to disk manually.
+///
+/// One use-case are environment variables that are modified by a layer that are required by the
```
edmorley, line 23:
```suggestion
/// One use-case is environment variables that are modified by a layer that are required by the
```

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// same buildpack in later stages of the build process. For example, a buildpack might install a
+/// build tool (i.e. Apache Maven) in one layer and adding the main binary to `PATH` via the `env`
```
edmorley, line 25:
```suggestion
/// build tool (e.g. Apache Maven) in one layer and add the main binary to `PATH` via the `env`
```

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// directory of that layer. The same buildpack then wants to execute Maven to download dependencies
+/// to a different layer. By using `LayerEnv`, the buildpack can encode these changes in a
+/// type and, in addition to passing it to libcnb which will persist it to disk, pass it to the
+/// logic that uses the build tool to download dependencies. The download process does not need to
+/// know the layer name or any logic how to construct `PATH`.
```
edmorley, line 30:
```suggestion
/// know the layer name or any of the logic for constructing `PATH`.
```

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+///
+/// # Applying the delta
+///`LayerEnv` is not a static set of environment variables, but a delta. Layers can modify existing
+/// variables by appending or prepending or setting new ones only conditionally. If you only need a
```
edmorley, line 34:
```suggestion
/// variables by appending, prepending or setting variables only if they were not already defined. If you only need a
```

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// env.insert("VAR", "foo");
+/// env.insert("VAR2", "previous-value");
+///
+/// let modified_env = layer_env.apply(TargetLifecycle::Build, &env);
+/// assert_eq!(modified_env.get("VAR").unwrap(), "foobar");
+/// assert_eq!(modified_env.get("VAR2").unwrap(), "previous-value");
+/// ```
+///
+/// # Implicit Entries
+/// Some directories in a layer directory are be implicitly added to the layer environment if they
```
edmorley, line 56:
```suggestion
/// Some directories in a layer directory are implicitly added to the layer environment if they
```

src/layer_env.rs
edmorley: ^

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+            // the raw bytes of that encoding as it might change the actual data. Since this is not
+            // explicitly written in the spec, we read through the the reference implementation and
+            // determined that it also treats the file contents as raw bytes.
+            // See: https://github.com/buildpacks/lifecycle/blob/a7428a55c2a14d8a37e84285b95dc63192e3264e/env/env.go#L73-L106
+            use std::os::unix::ffi::OsStringExt;
+            let file_contents = OsString::from_vec(fs::read(&path)?);
+
+            if let Some(file_name_stem) = file_name_stem {
+                let modification_behavior = match file_name_extension {
+                    None => {
```
edmorley, line 474: Similar to the "unknown extension" question for the code below, I'm presuming we'll only ever be reading the files that libcnb wrote, and as `write_to_env_dir()` always uses an explicit file extension, it seems we should never be hitting the "no extension" case here either. This would at least mean we don't need to worry as much about the CNB API 0.5+ differences, right? Should we even go as far as to remove support for the no extension case (eg error/panic) given explicit seems preferable and it's effectively dead code?

edmorley (approved):

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+    /// use std::fs;
+    ///
+    /// // Create a bogus layer directory
+    /// let temp_dir = tempdir().unwrap();
+    /// let layer_dir = temp_dir.path();
+    /// fs::create_dir_all(layer_dir.join("bin")).unwrap();
+    ///
+    /// let layer_env_dir = layer_dir.join("env");
+    /// fs::create_dir_all(&layer_env_dir).unwrap();
+    /// fs::write(layer_env_dir.join("ZERO_WING.default"), "ALL_YOUR_BASE_ARE_BELONG_TO_US").unwrap();
```
schneems, line 232: Someone set us up the bomb!

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+            ("PATH", TargetLifecycle::Launch, &bin_path),
+            ("LD_LIBRARY_PATH", TargetLifecycle::Launch, &lib_path),
+        ];
+
+        for (name, target_lifecycle, path) in layer_path_specs {
+            if path.is_dir() {
+                let target_delta = match target_lifecycle {
+                    TargetLifecycle::Build => &mut result_layer_env.layer_paths_build,
+                    TargetLifecycle::Launch => &mut result_layer_env.layer_paths_launch,
+                    _ => panic!("Unexpected TargetLifecycle in read_from_layer_dir implementation. This is a libcnb implementation error!"),
```
schneems, line 265: This should be unreachable if I'm reading right because we control all the inputs. If not then I was about to suggest we format the target_lifecycle and put it in the error message.

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+    Launch,
+    Process(String),
+}
+
+#[derive(Eq, PartialEq, Debug)]
+struct LayerEnvDelta {
+    entries: BTreeMap<(ModificationBehavior, OsString), OsString>,
+}
+
+impl LayerEnvDelta {
```
schneems, line 388: Comments and doc tests stopped here 👀

schneems (approved):
This was a beast to review. I'm betting much more so to actually write. Great job!

It was easier to review than I was originally anticipating. Credit to you and Rust. 

I left some comments as I went. Nothing blocking except maybe for the case of mis-spelling a process. 

For style, I generally like a space after markdown headers here are some docs for diesel.rs that have an extra space after headers https://docs.diesel.rs/src/diesel/query_builder/functions.rs.html#80-82. If you want I could send you suggestions for the rest of them.

src/env.rs
```diff
@@ -0,0 +1,120 @@
...
+    fn from(vars_os: VarsOs) -> Self {
+        Env {
+            inner: vars_os.collect(),
+        }
+    }
+}
+
+impl<'a> IntoIterator for &'a Env {
+    type Item = (&'a OsString, &'a OsString);
+    type IntoIter = std::collections::hash_map::Iter<'a, OsString, OsString>;
```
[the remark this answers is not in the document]
Malax: Great question! I cheated a bit in this implementation. Since at this point, `Env` is just a `HashMap` with a different name and fixed key/value types, I re-use [std::collections::hash_map::Iter](https://doc.rust-lang.org/std/collections/hash_map/struct.Iter.html) as the iterator implementation to avoid implementing it myself.

If you look closely, the `Item` for this `IntoIterator` is a tuple of [`OsString`](https://doc.rust-lang.org/std/ffi/struct.OsString.html) references. But `std::collections::hash_map::Iter` takes dedicated type variables for the key and the value.

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// fs::create_dir_all(layer_dir.join("include")).unwrap();
+///
+/// let layer_env = LayerEnv::read_from_layer_dir(&layer_dir).unwrap();
+///
+/// let env = Env::new();
+/// let modified_env = layer_env.apply(TargetLifecycle::Launch, &env);
+///
+/// assert_eq!(modified_env.get("PATH").unwrap(), layer_dir.join("bin"));
+/// assert_eq!(modified_env.get("CPATH"), None); // None, because CPATH is only added during build
+/// ```
```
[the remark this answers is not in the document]
Malax: 🎉 ❤️

src/layer_env.rs
Malax: Main screen turn on!

src/layer_env.rs
Malax: Yeah, I only added those for the public API, `LayerEnvDelta` is the first type that is not exposed from this module.

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+    pub fn apply(&self, target: TargetLifecycle, env: &Env) -> Env {
+        let deltas = match target {
+            TargetLifecycle::All => vec![&self.all],
+            TargetLifecycle::Build => vec![&self.all, &self.build, &self.layer_paths_build],
+            TargetLifecycle::Launch => vec![&self.all, &self.launch, &self.layer_paths_launch],
+            TargetLifecycle::Process(process) => {
+                let mut process_deltas = vec![&self.all];
+                if let Some(process_specific_delta) = self.process.get(&process) {
+                    process_deltas.push(process_specific_delta);
+                }
```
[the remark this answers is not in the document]
Malax: > It looks like if that happens then the process would get ALL of the env vars

I can see why you think that, but that's not the case. `self.all` contains all env var deltas that apply to all target lifecycles, not all deltas across all lifecycles.

With that out of the way, I don't think panicking is a good idea here. There is nothing wrong to apply the deltas for a process that doesn't have specific environment variables. If the user of this library passes in the wrong process type string, this is nothing the library should attempt to fix.

src/layer_env.rs
Malax: Correct, that's why the panic says "This is a libcnb implementation error!". To be honest I'm not super happy about this code. But adding another enum just for this case felt excessive.

src/layer_env.rs
Malax: > I'm presuming we'll only ever be reading the files that libcnb wrote

We will also read files created by the CNB lifecycle when that layer was cached. The lifecycle will not always restore the exact files we wrote, as it will normalize some parts.

In general, I don't think this code should make any assumptions when reading data from disk. Buildpack authors could, even if we don't want them to, write arbitrary env files to disk. Another process could do the same. Maybe an older version of the same buildpack that didn't use libcnb wrote environment variables without an extension. I think staying close to the spec (or the reference implementation when the spec isn't sufficient) is very important here. Future versions of the spec/lifecycle might make assumptions about older version to ensure backwards compatibility. If we diverge from the spec, we might run into issues later.

edmorley: @Malax I had not realised the env dir was also restored when a layer was cached. If that's the case, then 100% agree.

src/layer_env.rs
Malax: Like I said in [another comment](https://github.com/Malax/libcnb.rs/pull/30#discussion_r699079356), I don't think we should make any assumptions about the file system contents for various reasons. Let me know if you want me to talk about this specific instance more or if the link to the other comment is enough. :)

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// One use-case are environment variables that are modified by a layer that are required by the
+/// same buildpack in later stages of the build process. For example, a buildpack might install a
+/// build tool (i.e. Apache Maven) in one layer and adding the main binary to `PATH` via the `env`
+/// directory of that layer. The same buildpack then wants to execute Maven to download dependencies
+/// to a different layer. By using `LayerEnv`, the buildpack can encode these changes in a
+/// type and, in addition to passing it to libcnb which will persist it to disk, pass it to the
+/// logic that uses the build tool to download dependencies. The download process does not need to
+/// know the layer name or any logic how to construct `PATH`.
+///
+/// # Applying the delta
```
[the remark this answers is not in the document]
Malax: I'm not against adding newlines after markdown headings, but I'm hesitant to do the changes to this PR as we have other headings outside of this PR as well. If you feel strongly about this, I suggest we make DR about this, figure out how to enforce this automatically via a lint and ship a dedicated PR with the lint and the fixes. Is that okay for you?

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+/// let mut env = Env::new();
+/// env.insert("VAR", "foo");
+/// env.insert("VAR2", "previous-value");
+///
+/// let modified_env = layer_env.apply(TargetLifecycle::Build, &env);
+/// assert_eq!(modified_env.get("VAR").unwrap(), "foobar");
+/// assert_eq!(modified_env.get("VAR2").unwrap(), "previous-value");
+/// ```
+///
+/// # Implicit Entries
```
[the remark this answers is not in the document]
Malax: https://github.com/Malax/libcnb.rs/pull/30#discussion_r699099890

Malax:
Thank you @edmorley for the spelling and sentence structure corrections!

src/layer_env.rs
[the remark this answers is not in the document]
edmorley: I was curious to see what rustdocs/markdown related lints Clippy had already - there are a few listed at https://rust-lang.github.io/rust-clippy/stable/index.html but none that cover this case.

I did find https://github.com/rust-lang/rust-clippy/issues/1007 which covers adding more docs related lints.

src/layer_env.rs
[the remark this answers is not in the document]
edmorley: I guess I'm struggling to see the use-case where someone would be using process-specific `.apply()` during the build, to understand what the user expectations might be? (I can see why we need to write out process-specific env files, but in my mind they mostly would be useful for runtime.)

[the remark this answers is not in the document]
Malax: If there is a use-case during build it's most likely an edge-case. This PR includes support for applying them because it is mostly free in terms of required code. If we'd remove the ability to apply process specific environment deltas, we'd need more code as `TargetLifecycle` would be split up into two: one for specifying, one for applying. I can imagine the code being harder to follow too.

src/env.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+mod test {
```
hone, line 97: shouldn't this be in `#[cfg(test)]` so it's not built for non test envs?

src/layer_env.rs
```diff
@@ -0,0 +1,882 @@
...
+    ///
+    /// let mut layer_env = LayerEnv::new();
+    /// layer_env.insert(TargetLifecycle::All, ModificationBehavior::Append, "VAR", "bar");
+    /// layer_env.insert(TargetLifecycle::All, ModificationBehavior::Default, "VAR2", "default");
+    ///
+    /// let mut env = Env::new();
+    /// env.insert("VAR", "foo");
+    /// env.insert("VAR2", "previous-value");
+    ///
+    /// let modified_env = layer_env.apply(TargetLifecycle::Build, &env);
```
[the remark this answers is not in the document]
edmorley: I've added this comment to #73

src/env.rs
```diff
@@ -0,0 +1,120 @@
...
+    /// Inserts a key-value pair into the environment, overriding the value if `key` was already
+    /// present.
+    pub fn insert(&mut self, key: impl Into<OsString>, value: impl Into<OsString>) -> &mut Self {
+        self.inner.insert(key.into(), value.into());
+        self
+    }
+
+    /// Returns a cloned value corresponding to the given key.
+    pub fn get(&self, key: impl AsRef<OsStr>) -> Option<OsString> {
+        self.inner.get(key.as_ref()).cloned()
```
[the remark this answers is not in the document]
edmorley: Filed #74

src/env.rs
edmorley: Filed #75

edmorley:
I've created GUS-W-9959585 to track adding the next part of this.

PR #37, opened 2021-07-30
Fix CI config so it runs on all branches

edmorley:
Previously it only ran for `master` (or PRs targeting `master`), but the default branch is now `main`.

I've switched this to run for all branches (rather than just `main`), since it's useful to be able to see whether tests pass without having to open a PR.

See:
https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#on

Fixes GUS-W-9679128.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,10 +1,6 @@
 name: Rust
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+on: [push, pull_request]
 
 env:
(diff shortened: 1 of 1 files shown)


edmorley:
Hmm looks like GitHub Actions doesn't de-dupe the PR and push variants of the CI jobs (unlike Circle CI).

I'm open to switching back to specifying "main or PRs that target main" if you prefer :-)

Malax (approved):
Let's do this for now, we can always change it later. :)

PR #40, opened 2021-08-05
be defensive about state of <platform>/env folder

hone:
* handle if the env folder contains directories like the k8s mount case
* handle if the env folder does not exist

This should fix the java functions incident from a few weeks ago. Once this is merged, we can cut `0.1.2`. I'll also cherry pick this into `main` as a separate PR.


Malax (approved):

Malax:
Great debugging and fix! 🎉

PR #48, opened 2021-09-03
Assert literal toml string

schneems:
This isn't ideal as it's sensitive to changes in the output that won't affect the parsing. For instance if it decided to put two spaces on each side of the equals it would be the same toml output.

Ideally I would like to re-serialize it into TOML and assert specific values of it similar to https://github.com/heroku/buildpacks-ruby/blob/7893817cf76b36718da82ab294f2c5c3b7da9fab/spec/unit/bash_functions_spec.rb#L193-L196. I played around with the toml library interface but couldn't figure out how to get something like that working.

### DIFF
--- a/src/data/launch.rs
+++ b/src/data/launch.rs
@@ -27,7 +27,15 @@ pub struct Launch {
 /// false).unwrap();
 ///
 /// launch_toml.processes.push(web);
-/// assert!(toml::to_string(&launch_toml).is_ok());
+/// let toml_string = toml::to_string(&launch_toml);
+/// assert!(toml_string.is_ok());
+/// assert_eq!(toml_string.unwrap(), r#"
+/// [[processes]]
+/// type = "web"
+/// command = "bundle"
+/// args = ["exec", "ruby", "app.rb"]
+/// direct = false
+/// "#.trim_start());
 /// ```
 impl Launch {
     pub fn new() -> Self {


edmorley (commented):
Should this be a test rather than a doctest?

edmorley:
Or more to say: From reading between the lines, it looks like you've added this after hitting #49, in order to verify things are working as expected. However (a) if there is doubt that things are working as expected, we should add more tests, (b) debugging #49 is probably the next best step, since it will inform what we need to improve in libcnb.rs.

schneems:
> debugging #49 is probably the next best step, since it will inform what we need to improve in libcnb.rs.

#49 is closed, it was due to incompat between spec versions, but there are UX improvement opportunities.  

> Should this be a test rather than a doctest?

Could be either. I don't have a strong sense for cases where we should do one or the other. The existing doc test doesn't do anything other than show there's no error. It wasn't 100% clear to me that it did in fact produce toml on disk which is why I went to the docs. I would like the examples in the docs to be demonstrating something "real" if at all possible (for at least one case).

> if there is doubt that things are working as expected, we should add more tests

100% we should add more tests. More docs too. There's not a lot overall. My bias coming from a dynamic lang is probably to write too many tests, but that would be a high-quality problem to have.

Malax:
I'm with @edmorley on this one. This looks more like a regular test to me. We [have similar ones](https://github.com/Malax/libcnb.rs/blob/7fe69939501ee20e8f9f1274896e2c586badda6c/src/data/buildpack_plan.rs#L17-L51) already that validate the string representations in the `data` module.

In addition, I think doctests are meant to quickly show how the documented component works or to show the intended use. The fact that the code in the docs is actually run is more so to ensure the documentation is correct and not to test the code itself.

schneems:
> I think doctests are meant to quickly show how the documented component works or to show the intended use

I was thinking about this. I think a good example should show impact. Not just what it does, but WHY it does it. The underlying (implementation) reason for this launch interface is to generate toml so it can be written to disk. The existing example shows generating a toml object, and coming from a non-strongly-typed language ecosystem that impact wasn't clear/obvious.

That lead me to think that this existing test is brittle (as I pointed out before, any changes to how the toml is serialized could break the test). So that's not idea. The second thought is that, we're showing it being serialized to a toml object successfully. Which does show the WHY, but I'm not confident enough in the type guarantees. 

All that's to say. I agree. It's worth testing this, but 1) It's brittle and 2) Doesn't need to be a doc test. I'll update the PR.

Issue #49, opened 2021-09-03
Cannot write a basic web process with libcnb


schneems:
## Reproduction

- Example app fails with rust: https://github.com/schneems/basic_rust_lol_cnb_buildpack
- Example app works in bash: https://github.com/schneems/basic_bash_lol_cnb_buildpack

## Description

This in rust:

```rust
let process = Process::new(
    "web",
    String::from("while true; do echo 'lol'; sleep 2; done"),
    vec![String::from("")],
    false,
)?;
let launch = Launch::default().process(process);

context
    .write_launch(launch).unwrap();
```

Should be the same as this in bash:

```bash
cat >> "${layers_dir}/launch.toml" <<EOL
[[processes]]
type = "web"
command = "while true; do echo 'lol'; sleep 2; done"
args = [""]
direct = false
EOL
```

But it's not. The bash version works, the rust version does not.

schneems:
Closing this, Ed pointed out the examples are using different specs of CNB (0.2 vs 0.4). When I switch the bash version to 0.4 it fails too. Ed pointed out this change https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.4 to arg parsing. If I boot with empty args (instead of a vector with an empty string) then it works.

While it looked like it was a bug in lifecycle, Go and Ruby both treat an empty arg the same:

```
irb(main):005:0> pid = Process.spawn("while true; do echo 'lol'; sleep 2; done")
=> 13083
irb(main):006:0> lol

^C
irb(main):006:0> pid = Process.spawn("while true; do echo 'lol'; sleep 2; done", "")
(irb):6:in `spawn': No such file or directory - while true; do echo 'lol'; sleep 2; done (Errno::ENOENT)
```

I think that while the lifecycle launcher is technically correct, the error is confusing and difficult to debug. 

Also looking at the output:

```
$ pack inspect cutlass_image_basic_bash | grep -B2 web
Processes:
  TYPE                 SHELL        COMMAND        ARGS
  web (default)        bash         while true; do echo 'lol'; sleep 2; done
```

It's unclear that it's supplying the command with empty args (versus no args). If you put an actual arg the formatting changes:

```
Processes:
  TYPE                 SHELL        COMMAND                                         ARGS
  web (default)        bash         while true; do echo 'lol'; sleep 2; done        foo
```

Or if you supply no/empty args:

```
Processes:
  TYPE                 SHELL        COMMAND        ARGS
[...]

Issue #29, opened 2021-06-25
Figure out how to support multiple buildpack API versions 


Malax:

Malax:
I'm currently leaning towards 1 since it's the most simple implementation both in terms of maintenance and usage. Unless we have a great use-case for 2 and 3 I'd hold off implementing them, especially pre `1.0` where our API will change often and drastically.

edmorley:
I'm leaning towards (1) too.

The only reason I can see for needing to support multiple buildpack API versions in the same libcnb.rs release is if we regularly encounter situations where we need to release a bug fix or feature for users who are somehow stuck on an older buildpack API version.

And since a build can use buildpacks with a mixture of buildpack API versions, updating an individual buildpack's API version should be something that's generally easy, since it's not blocked on updating everything at once. Users will obviously need to be on a recent Pack CLI/`lifecycle` release, however it's not like `libcnb.rs` will support a new buildpack API version the instant it's released, so there's going to be lag on both sides anyway.

For one-off issues we can always release libcnb.rs hotfixes off a release branch.

hone:
@edmorley I think buildpacks API are going to be generally compatible across groups, but I don't think lifecycle can actually make that guarantee always and it's always best effort.

I think step 1 might be to just get the current API up to snuff with `0.5` then support `0.6`.

Malax:
I agree that we should at first document and ensure we support one specific API version properly.

hone:
@malax this feels like a DR, we should capture it (maybe in this repo). Thoughts?

Malax:
IDK if we need a DR document in this repo, but we need to document the strategy and currently supported version of the CNB spec in here :100:. Maybe in the `README` for now? I don't have a strong opinion other than that it needs to be documented.

hone:
For some prior art, the paketo buildpacks used to have a RFCs folder in the
repo. Maybe it's worth doing something similar with a DR?
https://github.com/paketo-buildpacks/nodejs/tree/main/rfcs

schneems:
> Maybe it's worth doing something similar with a DR

I'm +1 on that. I would like to close out this issue (productively) and it seems to document this and saying "here's the convo, here's what we decided" would be a good way forward).

edmorley:
We've currently settled on supporting only one version of the buildpack API (which IMO is the best option).

The currently supported API version is documented in the README:
https://github.com/Malax/libcnb.rs/blame/main/README.md#L14

[...]

PR #50, opened 2021-09-07
Add runtime error message on CNB API version mismatch

Malax:
Looking at #48 and #49, we need to be more explicit about the supported CNB API versions. There is an issue about this already (#29) which looks at the problem at a higher level. This PR adds an error message if the API versions in the  `buildpack.toml` and the supported CNB API version of ´libcnb.rs` do not match.

"But what is the supported CNB API version?" you might ask. This is not really clear and currently undocumented. To determine which version we initially targeted when we started writing libcnb, I looked at the [CNB spec releases](https://github.com/buildpacks/spec/releases) and found out that the de-facto supported version is `0.5`.

The main giveaway is the implementation of [RFC 0070](https://github.com/buildpacks/rfcs/blob/main/text/0070-new-buildpack-toml-keys.md) that landed in `0.6` which is currently unsupported. This also matches up with timeline when we started working on this: around December 2020 (https://github.com/Malax/libcnb.rs/commit/74400926a47e9fbd38b2258f6d2e5011c505c056).

So this currently checks for CNB API version `0.5`.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -6,11 +6,12 @@ use std::process::exit;
 use serde::de::DeserializeOwned;
 
 use crate::build::BuildContext;
+use crate::data::buildpack::BuildpackToml;
 use crate::detect::{DetectContext, DetectOutcome};
 use crate::error::{Error, ErrorHandler};
 use crate::platform::Platform;
 use crate::toml_file::{read_toml_file, write_toml_file};
-use crate::Result;
+use crate::{Result, LIBCNB_SUPPORTED_BUILDPACK_API};
 use std::fmt::{Debug, Display};
 
 /// Main entry point for this framework.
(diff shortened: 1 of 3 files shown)


src/runtime.rs
```diff
@@ -63,6 +64,24 @@ pub fn cnb_runtime<P: Platform, BM: DeserializeOwned, E: Debug + Display>(
     if let Err(lib_cnb_error) = result {
         exit(error_handler.handle_error(lib_cnb_error));
     }
+
+    match read_buildpack_toml::<BM, E>() {
+        Ok(buildpack_toml) => {
+            if buildpack_toml.api != LIBCNB_SUPPORTED_BUILDPACK_API {
+                eprintln!("Error: Cloud Native Buildpack API mismatch");
+                eprintln!(
+                    "This buildpack uses Cloud Native Buildpacks API version {}, but the underlying libcnb.rs library requires CNB API {}.",
```
joshwlewis, line 73: `This buildpack` might not be that helpful if you've got a bunch of buildpacks. Maybe add `buildpack_toml.name` to this message?

Malax: Good point, I added this in 7cf02ed! 👍🏻

src/runtime.rs
```diff
@@ -63,6 +64,28 @@ pub fn cnb_runtime<P: Platform, BM: DeserializeOwned, E: Debug + Display>(
     if let Err(lib_cnb_error) = result {
         exit(error_handler.handle_error(lib_cnb_error));
     }
+
+    match read_buildpack_toml::<BM, E>() {
```
edmorley, line 68: If I'm reading this correctly, this new check runs after the buildpack's detect/build runs, rather than before. What are your thoughts on moving this before? (To fail early and also prevent confusing error messages from usages of libcnb within detect/build due to something buildpack API related [though struggling to think of examples of when that might occur?])

Malax: You're right, and this wasn't the intention. I'll fix this. :)

Malax: Landed in 67d4bc6

edmorley (approved):
It would be great to have some tests for `cnb_runtime` too (one of which for this) :-)

Malax:
> It would be great to have some tests for `cnb_runtime` too (one of which for this) :-)

Yeah, the fact that `cnb_runtime` is so dependent on the environment, executable name, etc. makes it really hard to _unit_ test it. Adding abstractions for all these things just for the tests feels wrong but not having tests also feels wrong. We need a good integration test story.

PR #77, opened 2021-09-08
Allow ProcessTypes to be compared

schneems:
While writing a build pack I wanted to test that an `Launch::r#type` was properly set and tried to use `assert_eq!`. To my surprise I found that it could not be compared:

```
error[E0369]: binary operation `==` cannot be applied to type `ProcessType`
   --> src/data/launch.rs:146:9
    |
146 |           assert_eq!(
    |  _________^
    | |_________|
    | |
147 | |             ProcessType::from_str("web").unwrap(),
148 | |             ProcessType::from_str("web").unwrap()
149 | |         )
    | |         ^
    | |_________|
    | |_________ProcessType
    |           ProcessType
```

This is fixed by setting the partial eq trait https://doc.rust-lang.org/std/cmp/trait.PartialEq.html.

### DIFF
--- a/src/data/launch.rs
+++ b/src/data/launch.rs
@@ -99,7 +99,7 @@ pub struct Slice {
 /// let invalid = ProcessType::from_str("!nv4lid");
 /// assert!(invalid.is_err());
 /// ```
-#[derive(Serialize, Debug)]
+#[derive(Serialize, Debug, PartialEq, Eq)]
 pub struct ProcessType(String);
 
 impl ProcessType {
@@ -131,4 +131,22 @@ pub enum ProcessTypeError {
         "Found `{0}` but value MUST only contain numbers, letters, and the characters ., _, and -."
     )]
     InvalidProcessType(String),
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::str::FromStr;
+
+    #[test]
+    fn test_process_type_eq() {
+        assert_eq!(
+            ProcessType::from_str("web").unwrap(),
+            ProcessType::from_str("web").unwrap()
+        );
+        assert_ne!(
+            ProcessType::from_str("web").unwrap(),
+            ProcessType::from_str("nope").unwrap()
+        )
+    }
 }


src/data/launch.rs
```diff
@@ -100,6 +100,7 @@ pub struct Slice {
 /// assert!(invalid.is_err());
 /// ```
 #[derive(Serialize, Debug)]
+#[derive(PartialEq)]
```
edmorley, line 103: Should `Eq` be added too?
https://doc.rust-lang.org/std/cmp/trait.Eq.html

Also, should the two `derive` usages be combined?

Malax: I think `Eq` is what we want here, based on definitions in the linked docs. Since `Eq` always requires `PartialEq`, we have to implement them both.

> Also, should the two derive usages be combined?
`rustfmt` says so.

Malax (changes requested):
See comment under @edmorley's review

src/data/launch.rs
schneems: If I'm understanding correctly. The difference is that `Eq` enforces `reflexive` while `PartialEq` does not. Using the example from PartialEq `Nan` cannot be reflexive(ly) equal. 

As a side note, I'm not sure why `PartialEq` implements default `==` but `Eq` doesn't.

```
error[E0277]: can't compare `launch::ProcessType` with `launch::ProcessType`
   --> src/data/launch.rs:102:28
    |
102 | #[derive(Serialize, Debug, Eq)]
    |                            ^^ no implementation for `launch::ProcessType == launch::ProcessType`
    |
   ::: /Users/rschneeman/.rustup/toolchains/stable-x86_64-apple-darwin/lib/rustlib/src/rust/library/core/src/cmp.rs:268:15
    |
268 | pub trait Eq: PartialEq<Self> {
    |               --------------- required by this bound in `Eq`
    |
    = help: the trait `PartialEq` is not implemented for `launch::ProcessType`
    = note: this error originates in the derive macro `Eq` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Another way: Why isn't `Eq` derive a superset of `PartialEq` ?

schneems: PR is updated

Malax (approved):
👍🏻

PR #80, opened 2021-09-09
Add missing #[cfg(test)] to env.rs

### LINKED ISSUE #75: Add missing `#[cfg(test)]` to `Env`'s tests
I've split this review comment out of #30 so it doesn't get lost, since the PR has already been merged.

_Originally posted by @hone in https://github.com/Malax/libcnb.rs/pull/30#discussion_r699634951_

---

shouldn't this be in `#[cfg(test)]` so it's not built for non test envs?

---

See also:
https://doc.rust-lang.org/stable/book/ch11-03-test-organization.html#the-tests-module-and-cfgtest

Malax:
Fixes #75

### DIFF
--- a/src/env.rs
+++ b/src/env.rs
@@ -94,6 +94,7 @@ impl<'a> IntoIterator for &'a Env {
     }
 }
 
+#[cfg(test)]
 mod test {
     #[test]
     #[cfg(target_family = "unix")]


edmorley (approved):
Thank you :-)

PR #82, opened 2021-09-11
allow deserializing launch.toml

hone:
I think we should make everything deserialize/serialize (some of these will be a bit non-trivial). I'm starting with `launch.toml` since I need it right now. :)

### DIFF
--- a/src/data/bom.rs
+++ b/src/data/bom.rs
@@ -1,9 +1,9 @@
-use serde::Serialize;
+use serde::{Deserialize, Serialize};
 use toml;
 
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Thank you :-)

hone:
@edmorley do you usually wait for @Malax to chime in or should I just merge?

edmorley:
It depends - your call! :-)

Malax:
FWIW: I don't think I should be a blocker for PRs. :) I'm happy to review, but can't promise swift reviews right now.

PR #87, opened 2021-09-14
example-02 fixes

dzuelke:
- bump buildpack API version
- move to vendored OpenSSL for easy cross-building from macOS
- add `Makefile.toml` and `package.toml` for easy builds
- fix `Gemfile.lock` checksumming (was looking in the wrong dir, always failing at runtime)

`layer_lifecycle_data` isn't working yet (none of the env vars end up in the built image), that somehow got broken in the 00e20e8ba5021923217b80d5d2fa4174d6f836b1 refactor

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .idea
 Cargo.lock
 examples/**/target
+examples/**/target-cargo
--- a/examples/example-02-ruby-sample/Cargo.lock
+++ b/examples/example-02-ruby-sample/Cargo.lock
@@ -10,18 +10,18 @@ checksum = "ee2a4ec343196209d6594e19543ae87a39f96d5534d7174822a3ad825dd6ed7e"
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.15"
+version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7404febffaa47dac81aa44dba71523c9d069b1bdc50a77db41195149e17f68e5"
+checksum = "1e37cfd5e7657ada45f742d6e99ca5788580b5c529dc78faf11ece6dc702656f"
 dependencies = [
  "memchr",
 ]
 
 [[package]]
 name = "anyhow"
-version = "1.0.41"
+version = "1.0.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15af2628f6890fe2609a3b91bef4c83450512802e59489f9c1cb1fa5df064a61"
+checksum = "61604a8f862e1d5c3229fdd78f8b02c68dcf73a4c4b05fd636d12240aaa242c1"
 
 [[package]]
 name = "autocfg"
@@ -136,7 +136,7 @@ dependencies = [
  "anyhow",
  "flate2",
  "libcnb",
- "openssl-sys",
+ "openssl",
  "reqwest",
  "serde",
  "sha2",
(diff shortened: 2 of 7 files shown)


Malax:
Hey David, thanks for the fixes!

I'm unsure if we want our Heroku build scripts in this repository. They make it much easier to get started with the examples but make them more opinionated. @hone, @schneems, @edmorley: Thoughts?

edmorley:
I think we need to figure out the example story overall - at the moment we have many examples in varying states of out of date-ness:
- 2x `examples` in this repo that don't build / aren't tested in CI
- https://github.com/Malax/rust-cnb-starter
- https://github.com/heroku/heroku-buildpacks-repo-template

And then these implementations:
- https://github.com/heroku/buildpacks-jvm/tree/main/buildpacks/jvm-function-invoker
- https://github.com/heroku/procfile-cnb-rust

The difficulty is that:
1. We likely want both monorepo and single-buildpack examples (so having just a single example isn't viable)
2. We might not want Heroku/libherokubuildpack usages in this repo
3. Repos that aren't this repo (once tested in CI as changes are made), or an actual production buildpack repo are likely going to get out of date

One option would be to say that this repo is the only location for examples, and that those examples are solely single buildpacks not using `libherokubuildpack`, but that we also link out to actual production monorepo Heroku buildpacks in `examples/README.md`. And then we delete https://github.com/Malax/rust-cnb-starter and eventually https://github.com/heroku/heroku-buildpacks-repo-template (once we've stood up the per-lanuagage monorepos, that will then be where further changes evolve).

schneems:
I feel like someone should be able to build a buildpack with just libcnb.rs. I think this is related to my prior comment of https://github.com/Malax/libcnb.rs/issues/92. 

If you agree then I think the next requirement is for someone to be able to build a buildpack correctly using our documentation. I would say a requirement of that documentation is it needs to stay in alignment with libcnb and to *that* end I would say it needs to at least be executed, ideally be tested, even more ideal would be to ensure that it works with pack. If it works with pack, then it needs to be able to compile to musl.

It's a bit of a give a mouse a cookie scenario. 

Some alternatives:

- We could make a "CNB on Rust book" similar to rustdoc book (and friends).
- We could have libcnb be ONLY the low level API and have another higher level lib that sits somewhere between libcnb and libherokubuildpack. (Not loving this, as I don't really know what would go in it).
- We could commit to a single, minimal example and then push other examples to another repo/library (like a book library). 
- We could have CI pull in one or more external example repos and build them.

Right now our own ecosystem is very fragmented. We've got random repos all over the place that may or may not be up-to-date and may or may not be maintained anymore.

Recap: We should have at least one working example in our repo. We should have at least one canonical/maintained example that actually compiles to the point where it can be run by pack (may or may not live in libcnb.rs). We should have automation that ensure these examples don't break via compilation/test.

schneems:
I think i'm in favor of merging this in now and figuring out the next right thing later. We can always take out the makefile.toml at a future date. It's not serving anyone having broken examples in the repo right now.

schneems (approved):
We talked about this in planning today quickly. I want to merge this so the examples work. The larger question/problem/philosophy can be re-visited and talked about in the future. 

Ideally, these examples would have integration tests via cutlass or replacement. Ideally also the examples in the readme would be exercised as well https://github.com/Malax/libcnb.rs/issues/92. 

We need some forward progress and momentum. We can revisit this example and our example strategy in the future. I'm merging this in.

PR #88, opened 2021-09-17
Fix detect/build mode detection when using symlinks

### LINKED ISSUE #51: Support detect/build mode detection when using symlinks
Currently libcnb.rs is designed to be used with a buildpack built as a single binary shared between both `bin/detect` and `bin/build` in order to reduce compile time and buildpack size. libcnb.rs determines what mode it is being run on based on the binary's filename:
https://github.com/Malax/libcnb.rs/blob/7fe69939501ee20e8f9f1274896e2c586badda6c/src/runtime.rs#L41-L50

However the current implementation resolves symlinks to their target, meaning that hardlinks have to be used instead.

Unfortunately, pack doesn't preserve the hardlinks in the packaged buildpack (either when packaged to a `.cnb` file or to an OCI image) - instead including two copies of the files, defeating the file size reduction.

I will file an issue against pack's hardlink support, however I think it would be good to add symlink support to libcnb.rs regardless, since (a) we won't have to wait for pack/lifecycle to be fixed/released, (b) symlinks are slightly more intuitive to understand when someone is inspecting a buildpack, compared to hardlinks (which are mostly invisible, unless paying close attention to file properties).

edmorley:
`cnb_runtime()` has to detect the currently running binary name in order to determine whether the binary should be run in detect or build mode.

Previously the current binary name was calculated using [std::env::current_exe()](https://doc.rust-lang.org/std/env/fn.current_exe.html).

Unfortunately that function's behaviour varies by platform, and on some platforms (such as Linux) if the current file was invoked via a symlink, then `current_exe()` returns the symlink target rather than the original symlink path itself.

As such until now, buildpacks using libcnb.rs had used hard links instead of symlinks, however Pack does not handle hard links properly (buildpacks/pack#1286), which results in the binary being bundled twice in the buildpack (doubling the size).

Even if that Pack issue ends up being fixed, we likely want to support both symlinks and hard links, which we can do by using [std::env::args()](https://doc.rust-lang.org/std/env/fn.args.html) instead.

There are some caveats with that approach:

> The first element is traditionally the path of the executable, but it can be set to arbitrary text, and may not even exist.
> This means this property should not be relied upon for security purposes.

...however for our use-case this is absolutely fine, since (a) the buildpack packagers control the naming of the file + Pack/`lifecycle` controls how it is run, (b) any invalid or missing filename arguments will be rejected, (c) this isn't a security sensitive usage.

I looked into whether this could be easily tested, however it would be non-trivial. For now, this has been tested locally with a test buildpack using both symlinks and hard links. In the future perhaps this can be tested via `examples/` (once they are fixed and run in CI).

Fixes #51.
GUS-W-9866676.

### DIFF
--- a/src/runtime.rs
+++ b/src/runtime.rs
@@ -1,5 +1,6 @@
 use std::env;
-use std::path::PathBuf;
+use std::ffi::OsStr;
+use std::path::{Path, PathBuf};
 use std::process;
 use std::process::exit;
 
@@ -15,6 +16,14 @@ use crate::{Result, LIBCNB_SUPPORTED_BUILDPACK_API};
 use std::fmt::{Debug, Display};
 
 /// Main entry point for this framework.
+///
+/// The Buildpack API requires us to have separate entry points for each of `bin/{detect,build}`.
+/// In order to save the compile time and buildpack size of having two very similar binaries, a
+/// single binary is built instead, with the filename by which it is invoked being used to determine
+/// the mode in which it is being run. The desired filenames are then created as symlinks or
+/// hard links to this single binary.
+///
+/// Currently symlinks are recommended over hard hard links due to [buildpacks/pack#1286](https://github.com/buildpacks/pack/issues/1286).
 ///
 /// # Example
 /// ```no_run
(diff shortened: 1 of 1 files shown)


dzuelke:
Won't the names be `build.exe` and `detect.exe` on Windows though?

edmorley:
@dzuelke Ah good point, I will switch to [Path.file_stem](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem).

Malax:
Great fix, I like it! 👍🏻 

> I will switch to [Path::file_stem](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem).

*puts on pedantic hat* That'd allow the usage of i.e. `build.sh` on Linux. We could have constants for each target OS that contain the expected file name. TBF, this is nothing that I would block this PR on, especially since Windows support wasn't there before this PR either.

edmorley:
Yeah I agree - I'll leave the `target_family` for now.

Malax (approved):

Issue #79, opened 2021-09-08
Implement `GenericPlatform::new`


hone:
For testing, `GenericPlatform` should implement a `new()` function for instantiating a blank platform.

Malax:
+1, I noticed the same thing when trying to implement tests for `detect` and `build`. In addition to just `new`, we probably want additional functions to add env vars programmatically as well.

edmorley:
Fixed by #157.

Issue #70, opened 2021-09-08
Use default generic arguments for some of the `Generic*` types


edmorley:
This is a review comment split out of https://github.com/Malax/libcnb.rs/pull/28#pullrequestreview-705465664.

If this suggestion ends up being wontfix we can always close this issue out, however I didn't want it to get lost given that the PR was already merged :-)

> * For some of these `Generic*` things, I wonder if we can just have default generic arguments.
> -- @hone

edmorley:
@Malax Is this now redundant/wontfix after #129?

Malax:
TBH, I'm not sure what @hone meant with his comment. It might be resolved by #129, but I'm not sure.

Malax:
I think I now know what this was about. This was a comment from the `LayerLifecycle` work where we potentially could've set a default for `P: Platform` to `GenericPlatform` for example. I think @edmorley is right and this no longer applicable until [Rust RFC2532](https://rust-lang.github.io/rfcs/2532-associated-type-defaults.html) is finalized.

Feel free to re-open if you think this is still valid.

hone:
yeah, not possible right now until that RFC is merged, but we weren't using associated types before.

PR #93, opened 2021-09-22
Introduce test_helper::TempContext

schneems:
When exercising `build()` and `detect()` in tests we need an easy way to generate a `GenericBuildContext` and `GenericDetectContext`. This commit adds a helper that creates these structs and associated temporary directories.

GUS-W-9909197

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,4 @@ serde = { version = "^1.0.126", features = ["derive"] }
 thiserror = "^1.0.26"
 toml = "^0.5.8"
 anyhow = { version = "^1.0.42", optional = true }
-
-[dev-dependencies]
 tempfile = "^3.2.0"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,6 +39,7 @@ pub mod data;
 pub mod layer_env;
 
 pub mod layer_lifecycle;
+pub mod test_helper;
 
 use crate::data::buildpack::BuildpackApi;
 pub use build::BuildContext;
--- a/src/test_helper.rs
+++ b/src/test_helper.rs
@@ -1,0 +1,2 @@
+pub mod temp_context;
+pub use temp_context::TempContext;
--- a/src/test_helper/temp_context.rs
+++ b/src/test_helper/temp_context.rs
@@ -1,0 +1,115 @@
+use crate::{
+    data::buildpack::BuildpackToml, data::buildpack_plan::BuildpackPlan,
+    data::buildpack_plan::Entry, BuildContext, DetectContext, GenericBuildContext,
+    GenericDetectContext, GenericMetadata, GenericPlatform, Platform,
+};
+
+/// Creates build and detect contexts with references to temp directories for testing
+///
+/// When the context goes out of scope, the directories are cleaned from disk.
+///
+///```
+/// use libcnb::test_helper::TempContext;
+/// use std::fs;
+///
+/// pub fn build(context: libcnb::GenericBuildContext) {}
+/// pub fn detect(context: libcnb::GenericDetectContext) {}
+///
+/// let temp = TempContext::new("heroku-20", include_str!("../../examples/example-02-ruby-sample/buildpack.toml"));
+/// let build_context = temp.build;
+/// let detect_context = temp.detect;
+/// let launch_toml_path = build_context.layers_dir.join("launch.toml");
+///
+/// std::fs::write(build_context.app_dir.join("Procfile"), "web: bundle exec rails s").unwrap();
+///
+/// // Pass to build to assert behavior
+/// build(build_context);
+///
+/// // Pass to detect to assert behavior
+/// detect(detect_context);
+///```
+pub struct TempContext {
+    pub detect: GenericDetectContext,
+    pub build: GenericBuildContext,
+    _tmp_dir: tempfile::TempDir,
+}
+
+impl TempContext {
+    pub fn new(stack_id: impl AsRef<str>, buildpack_toml_string: impl AsRef<str>) -> Self {
+        let buildpack_toml_string = buildpack_toml_string.as_ref();
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let app_dir = tmp_dir.path().join("app");
+        let layers_dir = tmp_dir.path().join("layers");
+        let platform_dir = tmp_dir.path().join("platform");
+        let buildpack_dir = tmp_dir.path().join("buildpack");
+
+        for dir in [&app_dir, &layers_dir, &buildpack_dir, &platform_dir] {
+            std::fs::create_dir_all(dir).unwrap();
+        }
+
+        let stack_id = String::from(stack_id.as_ref());
+        let platform = GenericPlatform::from_path(&platform_dir).unwrap();
+        let buildpack_descriptor: BuildpackToml<GenericMetadata> =
+            toml::from_str(buildpack_toml_string).unwrap();
+
+        let detect_context = DetectContext {
+            platform,
+            buildpack_descriptor,
+            app_dir: app_dir.clone(),
+            buildpack_dir: buildpack_dir.clone(),
+            stack_id: stack_id.clone(),
+        };
+
+        let platform = GenericPlatform::from_path(&platform_dir).unwrap();
+        let buildpack_descriptor: BuildpackToml<GenericMetadata> =
+            toml::from_str(buildpack_toml_string).unwrap();
+        let buildpack_plan = BuildpackPlan {
+            entries: Vec::<Entry>::new(),
+        };
+        let build_context = BuildContext {
+            layers_dir,
+            app_dir,
+            buildpack_dir,
+            stack_id,
+            platform,
+            buildpack_plan,
+            buildpack_descriptor,
+        };
+
+        TempContext {
+            detect: detect_context,
+            build: build_context,
+            _tmp_dir: tmp_dir,
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_directories_are_created() {
+        let temp = TempContext::new("heroku-20", fake_buildpack_toml());
+        assert!(temp.build.app_dir.exists());
+        assert!(temp.build.layers_dir.exists());
+        assert!(temp.build.buildpack_dir.exists());
+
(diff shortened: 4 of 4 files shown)


Cargo.toml
```diff
@@ -20,6 +20,6 @@ serde = { version = "^1.0.126", features = ["derive"] }
 thiserror = "^1.0.26"
 toml = "^0.5.8"
 anyhow = { version = "^1.0.42", optional = true }
+tempfile = "^3.2.0"
 
 [dev-dependencies]
```
edmorley, line 25: The whole `[dev-dependencies]` section can be removed now

edmorley (commented):
I wonder if we could reduce the amount of duplication here (with `cnb_runtime_detect`, `cnb_runtime_build` etc) by giving `GenericPlatform`, `DetectContext` and `BuildContext` constructors?

See also #79 and #70.

schneems:
> I wonder if we could reduce the amount of duplication here (with cnb_runtime_detect, cnb_runtime_build etc) by giving GenericPlatform, DetectContext and BuildContext constructors?

I looked into it and it seems that while there's many lines of code here, most are just passing values. Pulling the constructor stuff into a `::new()` only really abstracts away 2 lines those map 1:1 on inputs so we still need to pass just as many values:

```rs
impl GenericBuildContext {
    fn new(layers_dir: PathBuf, app_dir: PathBuf, buildpack_dir: PathBuf, platform_dir: PathBuf, stack_id: String, buildpack_plan: BuildpackPlan, buildpack_toml_string: &str) -> GenericBuildContext {
        let platform = GenericPlatform::from_path(&platform_dir).unwrap();
        let buildpack_descriptor: BuildpackToml<GenericMetadata> =
            toml::from_str(buildpack_toml_string).unwrap();

        BuildContext {
            layers_dir,
            app_dir,
            buildpack_dir,
            stack_id,
            platform,
            buildpack_plan,
            buildpack_descriptor
        }
    }
}
```

The `cnb_runtime_build` doesn't use the `GenericBuildContext` so this extraction wouldn't be re-used anywhere:

```rust
fn cnb_runtime_build<
    E: Debug + Display,
    F: Fn(BuildContext<P, BM>) -> Result<(), E>,
    BM: DeserializeOwned,
    P: Platform,
>(
    build_fn: F,
) -> Result<(), E> {
    let args = parse_build_args_or_exit();

    let layers_dir = args.layers_dir_path;

    let app_dir = env::current_dir().map_err(Error::CannotDetermineAppDirectory)?;

    let stack_id: String = env::var("CNB_STACK_ID").map_err(Error::CannotDetermineStackId)?;

    let platform =
        P::from_path(&args.platform_dir_path).map_err(Error::CannotCreatePlatformFromPath)?;

    let buildpack_plan =
        read_toml_file(&args.buildpack_plan_path).map_err(Error::CannotReadBuildpackPlan)?;

    let context = BuildContext {
        layers_dir,
        app_dir,
        stack_id,
        platform,
        buildpack_plan,
        buildpack_dir: read_buildpack_dir()?,
        buildpack_descriptor: read_buildpack_toml()?,
    };

    build_fn(context)
}
```

Cargo.toml
schneems: Updated and removed this section.

Malax:
Referencing https://github.com/Malax/libcnb.rs/issues/79

Malax (changes requested):
I'll write a detailed explanation why I think this shouldn't be merged in the current state ASAP. I wanted to get my "request changes" in now, since I'll need some time to write my thoughts down but don't want to risk that this is getting merged as-is.

schneems:
Sounds good. Looking forward to your review.

Malax:
Thanks for the PR @schneems!

After reading the source, I identified the following issues with the current implementation:
- No support for metadata other than `GenericMetadata`
- No support for setting environment variables
- No support for any other platform than `GenericPlatform`
- Using a string for buildpack.toml instead of `BuildpackToml<T>`
- Additional dependency at runtime because of a test helper (minor)

I think most of the issues are due to the design of this test helper and proper fixes require a slightly different approach.

My thoughts:

Starting with the most flexible approach that allows users to test every aspect of their detect/build implementations, we would have something like the code below. It allows for custom metadata, environment variables, platforms that are not `GenericPlatform`, etc. I think the first step should be enabling the below as our base.

```rust
let detect_context = DetectContext {
    app_dir: PathBuf::from(""),
    buildpack_dir: PathBuf::from(""),
    stack_id: String::from("heroku-20"),
    platform: GenericPlatform {
        env: {
            let mut env = Env::new();
            env.insert("", "");
            env
        }
    },
    buildpack_descriptor: BuildpackToml {
        api: BuildpackApi { major: 0, minor: 5 },
        buildpack: Buildpack {
            id: "heroku/foo".parse::<BuildpackId>().unwrap(),
            name: String::from("d"),
            version: semver::Version::parse("0.1.2").unwrap(),
            homepage: None,
            clear_env: false,
        },
        stacks: vec![Stack {
            id: "heroku-20".parse::<StackId>().unwrap(),
            mixins: vec![],
        }],
        order: vec![],
        metadata: JvmFunctionInvokerBuildpackMetadata {
            runtime: JvmFunctionInvokerBuildpackRuntimeMetadata {
                url: String::from("http://example.com/404.tar.gz"),
                sha256: String::from("bogus"),
            }
        },
    },
};
```

This mostly works today already, but there are some things that need to change. As outlined in https://github.com/Malax/libcnb.rs/issues/79, `GenericPlatform` currently cannot be constructed from user code. This is the first thing we need to make the above work. Even though the above works and is very flexible for testing, it is obviously unwieldy for simple test cases. But it allows us to implement convenience helpers on top.

A use-case that works well with the current implementation in this PR is testing with the `buildpack.toml` from the buildpack source instead of duplicating the values in code. Luckily, we can support the same without any extra implementation from the libcnb side:

```rust
use toml;

let detect_context = DetectContext {
    app_dir: PathBuf::from(""),
    buildpack_dir: PathBuf::from(""),
    stack_id: String::from("heroku-20"),
    platform: GenericPlatform {
        env: {
            let mut env = Env::new();
            env.insert("", "");
            env
        }
    },
    buildpack_descriptor: toml::from_str::<BuildpackToml<JvmFunctionInvokerBuildpackMetadata>>(include_str!("../buildpack.toml")).unwrap(),
};
```

Let's think about some convenience helpers. By using the builder pattern, we can provide sensible defaults but also allow users to override everything if they need to:

```rust
let detect_context = DetectContextBuilder::new("heroku-20")
    .buildpack_toml_from_file("../buildpack.toml")
    .app_dir(PathBuf::from("../test-assets/whatever/app-1"))
    .build()
    .unwrap();
```

In this super simple case, the resulting detect context will create temporary directories for anything except the app directory. The stack will be set to `heroku-20` and the buildpack_toml information will be loaded from disk.

Here are some additional ideas what we could support:

```rust
let detect_context = DetectContextBuilder::new("heroku-20")
    .buildpack_toml(BuildpackToml {
        api: BuildpackApi { major: 0, minor: 5 },
        buildpack: Buildpack {
            id: "heroku/foo".parse::<BuildpackId>().unwrap(),
            name: String::from("d"),
            version: semver::Version::parse("0.1.2").unwrap(),
            homepage: None,
            clear_env: false,
        },
        stacks: vec![Stack {
            id: "heroku-20".parse::<StackId>().unwrap(),
            mixins: vec![],
        }],
        order: vec![],
        metadata: JvmFunctionInvokerBuildpackMetadata {
            runtime: JvmFunctionInvokerBuildpackRuntimeMetadata {
                url: String::from("http://example.com/404.tar.gz"),
                sha256: String::from("bogus"),
            }
        },
    })
    .platform_env("HEROKU_DEBUG", "true")
    .platform_env("VERBOSE_ERRORS", "true")
    .build()
    .unwrap();
```

Using the builder, the same base configuration can be shared across tests:

```rust
#[cfg(test)]
mod test {
    #[test]
    fn test_unsupported_stack() {
        let detect_context = detect_context_builder().stack("cedar-14").build().unwrap();
        // ...
    }

    #[test]
    fn test_app_2() {
        let detect_context = detect_context_builder().app_dir(PathBuf::from("../test-assets/whatever/app-2")).build().unwrap();
        // ...
    }

    fn detect_context_builder() -> DetectContextBuilder {
        DetectContextBuilder::new("heroku-20")
            .buildpack_toml_from_file("../buildpack.toml")
            .app_dir(PathBuf::from("../test-assets/whatever/app-1"))
    }
}
```

The examples are just a small sample of what we could support. We need to put some work into the design of the builder, but together with the base support for creating `DetectContext` by hand (see beginning of this comment), we cover all use-cases while making it easy to use even in simple test cases.

schneems:
Great. I have been thinking about the problem of default or optional values recently and this will give me an opportunity to experiment a bit if you've got some time I would also appreciate some feedback on https://github.com/heroku/libherokubuildpack/pull/32.

schneems:
One issue with the proposed design though is: A hard requirement for this work for me in testing is supporting temporary directories (which is why this is a struct in the first place).

I see two options either the builder returns an intermediate struct that contains a reference to the tmpdir or we provide another struct that creates all the tempdirs and the builder consumes/references it.
External dirs:

```
let temp_dirs = TempDirs::new()
let detect_context = DetectContextBuilder::new("heroku-20")
    .buildpack_toml_from_file("../buildpack.toml")
    .dirs_from_object(temp_dirs)
    .build()
    .unwrap();
```

Internal dirs:

```

let builder = DetectContextBuilder::new("heroku-20")
    .buildpack_toml_from_file("../buildpack.toml")
    .with_temp_dirs()
    .build()
    .unwrap();
let detect_context = builder.context;
```

I think I like the first one better, it seems more explicit.

Malax:
> if you've got some time I would also appreciate some feedback on [heroku/libherokubuildpack#32](https://github.com/heroku/libherokubuildpack/pull/32).

I can't access this outside of work :(

edmorley:
> I can't access this outside of work :(

If it's of interest, Richard has cloned to a private repo under his account https://github.com/schneems/libherokubuildpack_copy and presumably sent an invite to you (since I got an invite).

schneems:
I made a builder for DefaultContext as a spike https://github.com/Malax/libcnb.rs/pull/96

edmorley:
@schneems Should this be closed in favour of #96? It feels like this PR is less likely to be merged?

PR #94, opened 2021-09-22
Add default key to launch.toml

schneems:
From https://github.com/buildpacks/spec/blob/6ae550e868508ed242eeac6685bda6bb99cae7f4/buildpack.md#launchtoml-toml

> MAY specify a default boolean that indicates that the process type should be selected as the buildpack-provided default during the export phase.


edmorley (commented):
The `default` key isn't supported until Buildpack API 0.6:
https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.6

Whereas libcnb currently only works with Buildpack API 0.5:
https://github.com/Malax/libcnb.rs/blob/3180fb553fce45c70740c06eb9b13a177493e486/src/lib.rs#L63

There is a GUS work item open for adding support for Buildpack API 0.6 ([GUS-W-9625362](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07AH000000RebNYAS/view)) however no GitHub issue - so I've filed one now: #95

I haven't tried this change locally, but my gut feeling is that making it without bumping the `buildpack.toml` API version might result in build failures. That is - I think this change will need to happen at the same time as #95 rather than before.

schneems:
I checked and it doesn't error on the undefined key. It may not behave like the developer thinks, but it won't error:

```
cd /tmp
mkdir default_launchtoml
cd default_launchtoml
mkdir bin
cat > bin/detect <<EOF
#!/usr/bin/env bash
exit 0
EOF

cat > bin/build <<EOF
#!/usr/bin/env bash

layers_dir=\$1

cat >> "\${layers_dir}/launch.toml" <<EOL
[[processes]]
type = "web"
command = "while true; do echo 'lol'; sleep 2; done"
args = []
direct = false
default = true
EOL
EOF

cat > buildpack.toml <<EOF
# Buildpack API version
api = "0.5"

# Buildpack ID and metadata
[buildpack]
id = "samples/bash-script"
version = "0.0.1"
name = "Bash Script Buildpack"

# Stacks that the buildpack will work with
[[stacks]]
id = "io.buildpacks.samples.stacks.bionic"

[[stacks]]
id = "io.buildpacks.samples.stacks.alpine"

[[stacks]]
id = "io.buildpacks.stacks.bionic"

[[stacks]]
id = "heroku-20"
EOF

chmod +x bin/detect
chmod +x bin/build

pack build cutlass_image_basic_bash --path . -B heroku/buildpacks:20 --buildpack ./  -v
$ docker run -it --rm --init cutlass_image_basic_bash
lol
lol
lol
```

Malax:
I'm with Ed on this one. We should avoid ambiguous CNB spec support if possible. However, this is a good PR for `0.6` support. Should we merge it into a `0.6.0` branch instead of main? Or are we worried that `main` evolves too fast and keeping up `0.6.0` is too much work?

schneems:
On https://github.com/Malax/libcnb.rs/issues/95#issuecomment-925453064. I see the todos as:

- You or Ed have a look at the exec.d stuff and let me know what need to be done. If neither of you can get to it I can go digging but I'll have to start building context from scratch.
- For "Adds types table to /.toml." if we want to be extra safe we can make it a requirement to support this for 0.6 but honestly it seems low risk.
- I **think** everything else either requires no work or is optional. Please review my analysis and contradict any claims I've made.

If that's true then we seem very close to supporting 0.6 in which case I think it's fine to target main for 0.6. 

If we do decide to do branch-based dev, one pattern I've seen that works well for Rails is to have the future version (0.6) on main and make a branch for the current version. PRs by default would target main (0.6) and we could decide what we want to backport to a v0.5 branch.

It has the added benefit of if we ever decide we want to do multi-version support based on libcnb versions it's easy to jump to older versions.

I went ahead and made a `v0.5` branch, we can decide later if we need to use it.

edmorley (approved):
I think we should just merge this to `main`, and plan on finishing adding support for 0.6 before the next libcnb.rs release (0.6 needs to be a priority IMO so this should be fine; see #95 for remaining tasks) :-)

schneems:
Sounds good. What do you think about shipping a builder for Process? It's got several optional components. Right now this update will force anyone who uses Process to update their code to explicitly specify if their process is default or not. If we shipped a builder, it would allow us to define a default value here.

edmorley:
@schneems Yeah open to that :-)

hone:
We should probably make it clear in the `README` what version is being supported I'd be in favor of a branch personally, unless we plan on rolling up all our releases into `0.6` pretty quickly.

hone:
> There's already 0.6 stuff on main now. I'm just trying to knock off as many issues in https://github.com/Malax/libcnb.rs/milestone/2 as I can. I want to get 0.6 out.

yeah, that was the stuff I wanted to avoid actually. I was hoping to cut a last actual `0.5` compatible release.

PR #96, opened 2021-09-24
Add a builder for Context

schneems:
This commit is an implementation of a "consuming builder" for DetectContext https://doc.rust-lang.org/1.0.0/style/ownership/builders.html.

I'm looking for feedback on the high-level API. i.e., These method names:

```rs
let context = DetectContextBuilder::new("heroku-20")
    .buildpack_toml_from_string(include_str!("../examples/example-02-ruby-sample/buildpack.toml"))
    .from_temp_dir(&temp_dir)
    .build::<std::io::Error>()
    .unwrap();

detect(context);
```

I'm also interested in ways to improve the following problems:

- The generation of errors and values inside of `build()` is gnarly. It works for now...but it's brute force via match, not very clean.
- Ambiguous multiple value errors ("call ONLY one") don't list the values that are set, just says the developer set multiple values.
- I had to add a method to toml_file.rs to get error handling to work as I wanted. Maybe we can do it without that modification.
- I wrote code this to be in line with the code inside runtime.rs, which uses generics and including generic error. That means Rust needs to be able to resolve the error type even when it's not used (due to `.unwrap()` hence the funky:

```rs
.build::<std::io::Error>()
```

I talked to Terence about this, and he said since we know all the possible errors coming from this code invocation, there might be another way to split this out, but it would require refactoring how we're doing error handling in the project.

- Build method is huge
- The error handling was very difficult. I ended up "cheating" and adding an "Other" error type to pass in a string. I'm not sure what's a better way to do this.
- app_dir and buildpack_dir don't have any existence checks. The PathBuf is not guaranteed to exist.

Thoughts:

- I don't think it's possible to do this as a non-consuming builder. We would not be allowed to take any values that cannot be cloned. Right now, we're taking Platform and BuildpackToml<BM>. We could maybe re-work to require these to be clonable, but I think that's a considerable change.
- I renamed the interface `buildpack_descriptor` to `buildpack_toml` as I think it's less vague. I didn't touch the DetectContext interface and left it as `buildpack_descriptor`.
- We could get more or less magical as we choose. For instance, we could check to see if a buildpack.toml file exists based in the `buildpack_dir` and use it as a fallback.


src/detect.rs
```diff
@@ -18,3 +23,221 @@ pub enum DetectOutcome {
...
+
+    pub fn build<E: Display + Debug>(self) -> crate::Result<DetectContext<P, BM>, E> {
+        match vec![self.platform.is_some(), self.platform_dir.is_some()]
+            .iter()
+            .filter(|&&x| x)
+            .count()
+        {
+            0 => {
+                return Err(crate::error::Error::Other(
+                    "Platform is missing, call platform() or platform_from_path()".to_string(),
```
hone, line 146: If we want to use this to replace code internally, I'm not sure this is a great error message to return. This feels very specific to testing.

src/detect.rs
```diff
@@ -18,3 +23,221 @@ pub enum DetectOutcome {
...
+impl<P: Platform, BM: DeserializeOwned> DetectContextBuilder<P, BM> {
+    pub fn new(stack_id: impl AsRef<str>) -> Self {
+        DetectContextBuilder {
+            app_dir: None,
+            buildpack_dir: None,
+            buildpack_toml_path: None,
+            buildpack_toml_string: None,
+            platform_dir: None,
+            stack_id: String::from(stack_id.as_ref()),
+            platform: None,
```
hone, line 46: Should we default this to `GenericPlatform` if none is defined?

src/detect.rs
```diff
@@ -18,3 +23,221 @@ pub enum DetectOutcome {
...
+    /// let temp_dir = tempfile::tempdir().unwrap();
+    /// let context = DetectContextBuilder::new("heroku-20")
+    ///     .buildpack_toml_from_string(include_str!("../examples/example-02-ruby-sample/buildpack.toml"))
+    ///     .from_temp_dir(&temp_dir)
+    ///     .build::<std::io::Error>()
+    ///     .unwrap();
+    ///
+    /// detect(context);
+    ///```
+    pub fn from_temp_dir(mut self, temp_dir: &tempfile::TempDir) -> DetectContextBuilder<P, BM> {
```
hone, line 68:
```suggestion
    pub fn from_temp_dir(mut self, temp_dir: impl AsRef<Path>) -> DetectContextBuilder<P, BM> {
```

Should we use `Path` here? We don't do anything special with `TempDir` here.

src/detect.rs
```diff
@@ -18,3 +23,221 @@ pub enum DetectOutcome {
...
+        self
+    }
+
+    pub fn buildpack_toml_from_file(
+        mut self,
+        path: impl AsRef<Path>,
+    ) -> DetectContextBuilder<P, BM> {
+        self.buildpack_toml_path = Some(path.as_ref().to_path_buf());
+        self
+    }
```
hone, line 105: I wonder if we could take an enum here and then match it on build:

```rust
enum BuildpackTomlArg {
  Path(PathBuf),
  Str(String)
}

pub fn buildpack_toml(input: BuildpackTomlArg) {
  // ...
}
```

hone:
There are a few ways we can reduce the `match` complexity in `build`:

* Do we want a constructor that forces users to input the required args instead as having them as methods?
* If we leverage `enum`s for multiple ways to do something but you need at least, we won't need to check multiple `is_some()` calls

src/detect.rs
schneems: We need to be lazy about it and set it in the `build()` otherwise we lose the programmer intent. I'm fine with defaulting more things. We could get rid of this error and replace it with a GenericPlatform instead.

```
                    return Err(crate::error::Error::Other(
                        "Platform is missing, call platform() or platform_from_path()".to_string(),
                    ))
```

It's ultimately a design question if that would be "unexpected" and we want the programmer to be more explicit/intentional or we want it to JUST WORK ™️. 

I guess as I'm typing this out, I'm realizing that if they weren't expecting a GenericPlatform, and got one then their types would likely get borked and they would find out that way. In that case...I'm in favor of handling stuff for the user where possible.

schneems: Actually, scratch that, it looks like we need a dir to build one `GenericPlatform::from_path(&platform_dir)`

src/detect.rs
schneems: I wanted to restrict it to make sure people were using this in a way I expected. I'm worried someone might point it to a their current directory and suddenly new "app" and "buildpack" directories are created and modified.

src/detect.rs
schneems: I like that.

src/detect.rs
schneems: It's a very internal error message. The flexibility that using a builder buys you means that it can also fail in ways not previously possible. When it fails I want to give the implementer more info on how to fix it. I'm not sure how to better guard to make sure that end buildpack authors don't see this (unless in testing).

src/detect.rs
schneems: This would make for a more verbose API if we expose it:

```
    /// let temp_dir = tempfile::tempdir().unwrap();
    /// let context = DetectContextBuilder::new("heroku-20")
    ///     .buildpack_toml(BuildpackTomlArg::Str(include_str!("../examples/example-02-ruby-sample/buildpack.toml")))
    ///     .from_temp_dir(&temp_dir)
    ///     .build::<std::io::Error>()
    ///     .unwrap();
```

But if we don't expose it, and set it as a value internally it solves the problem of more than one source of truth, in this case the last one wins. Which might be confusing but seems fine. I tried it out and it does make some of the match stuff cleaner too reason about:

```
    fn set_blerg(mut self, platform: BuildpackToml<BM>) -> DetectContextBuilder<P, BM> {
        self.blerg = Some(BuildpackTomlArg::BuildpackToml(platform));
        self
    }

    fn set_blerg_string(mut self, string: impl AsRef<str>) -> DetectContextBuilder<P, BM> {
        self.blerg = Some(BuildpackTomlArg::Str(string.as_ref().to_string()));
        self
    }

    fn set_blerg_path(mut self, path: impl AsRef<Path>) -> DetectContextBuilder<P, BM> {
        self.blerg = Some(BuildpackTomlArg::Path(path.as_ref().to_path_buf()));
        self
    }
```

```
        if let Some(arg) = self.blerg {
            let foo = match arg {
                BuildpackTomlArg::BuildpackToml(buidpack_toml) => buildpack_toml,
                BuildpackTomlArg::Str(contents) => crate::toml_file::from_string(contents)
                    .map_err(Error::CannotReadBuildpackDescriptor)?,
                BuildpackTomlArg::Path(path) => crate::toml_file::read_toml_file(path)
                    .map_err(Error::CannotReadBuildpackDescriptor)?,
            };
        };
```

> Note: Using blerg here as a temp replacement.

schneems: Or

```rs
let buildpack_toml = self
    .buildpack_toml
    .ok_or_else(|| {
        crate::error::Error::Other(
            "Platform is missing, call platform() or platform_from_path()".to_string(),
        )
    })
    .and_then(|arg| match arg {
        BuildpackTomlArg::BuildpackToml(a) => Ok(a),
        BuildpackTomlArg::Str(contents) => crate::toml_file::from_string(contents)
            .map_err(Error::CannotReadBuildpackDescriptor),
        BuildpackTomlArg::Path(path) => crate::toml_file::read_toml_file(path)
            .map_err(Error::CannotReadBuildpackDescriptor),
    })?;
```

schneems: I really like the result. I pushed up my code after porting everything to an enum.

src/detect.rs
hone: The downside is now we're requiring this dev dependency in production builds. I feel like forcing the tempdir doesn't buy us very much. I expect people will read the docs and use it as intended.

hone:
We use buildpack_descriptor because that is what it's called in the spec. I know why you changed it but I wonder if it makes sense to change it back.

I'd like to remove all of the different buildpack_toml methods. I have an idea that may work which leverages the from trait on the enum.

schneems:
> I have an idea that may work which leverages the from trait on the enum.

Great. I was wondering if that was possible/easy/aesthetic. 

> We use buildpack_descriptor because that is what it's called in the spec.

I can switch it back pretty easily. 

I was also wondering how much work it would be to make `P` and `BuildpackToml<BM>` clone-able (which would allow the builder to not consume itself, I think). Then I think we could have the same builder class be able to generate both a `BuildContext` and `DetectContext` from the same entity, as a lot of the arguments overlap.

src/detect.rs
schneems: >  I expect people will read the docs

Bold claim 😹

> The downside is now we're requiring this dev dependency in production builds. 

That's a good point. It also forces a specific tempdir library, which isn't actually needed. I'll switch to using a Path instead.

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
+
+enum BuildpackTomlArg<BM> {
+    Path(PathBuf),
+    Str(String),
+    BuildpackToml(BuildpackToml<BM>),
+}
+
+enum PlatformArg<P> {
+    Plat(P),
+    Dir(PathBuf),
```
Malax, line 35: Naming & Consistency: In `BuildpackTomlArg`, we use `Path(PathBuf)` and here we use `Dir(PathBuf)`. One points to a file and one to a directory, but I think there is an opportunity to clarify naming here.

What about `File(PathBuf)` and `Directory(PathBuf)` respectively? Another alternative could be using `Path` for both.

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
+}
+
+pub struct DetectContextBuilder<P: Platform, BM: DeserializeOwned> {
+    app_dir: Option<PathBuf>,
```
Malax, line 39: We could model this without `Option` by creating an enum with two options: `Temporary` and `Static(PathBuf)`. In terms of code this is very similar to using an `Option`, but with additional semantics.

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
+    buildpack_dir: Option<PathBuf>,
```
Malax, line 40: We could model this without `Option` by creating an enum with two options: `Temporary` and `Static(PathBuf)`. In terms of code this is very similar to using an `Option`, but with additional semantics.

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
+    /// let temp_dir = tempfile::tempdir().unwrap();
+    /// let context = DetectContextBuilder::new("heroku-20")
+    ///     .buildpack_toml_from_string(include_str!("../examples/example-02-ruby-sample/buildpack.toml"))
+    ///     .from_path(&temp_dir)
+    ///     .build::<std::io::Error>()
+    ///     .unwrap();
+    ///
+    /// detect(context);
+    ///```
+    pub fn from_path(mut self, dir: impl AsRef<Path>) -> DetectContextBuilder<P, BM> {
```
Malax, line 96: In Rust, `from_` is[ a widely used prefix](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) for converting `A` into `B`. This function does something very different and might be confusing since it goes against that convention.

Malax: I feel like we don't need this function at all. It could be default behaviour to create temporary directories for `app_dir` and `platform_dir`. It would also simplify the amount of state we need to keep track of in the builder itself.

In addition, the way this is implemented currently, a builder cannot be safely used concurrently since, when using temporary directories created by this function, every invocation of build will create contexts with the same paths.

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
+        }
+        self
+    }
+
+    pub fn platform(mut self, platform: P) -> DetectContextBuilder<P, BM> {
+        self.platform = Some(PlatformArg::Plat(platform));
+        self
+    }
+
+    pub fn platform_from_path(
```
Malax, line 125: Naming: Avoiding the `from` prefix, this could be called `platform_path`. For consistency, the naming for `app_dir` is slightly different too.

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
+        self.app_dir = Some(app_dir.as_ref().to_path_buf());
+        self
+    }
+
+    pub fn buildpack_dir(mut self, buildpack_dir: impl AsRef<Path>) -> DetectContextBuilder<P, BM> {
+        self.buildpack_dir = Some(buildpack_dir.as_ref().to_path_buf());
+        self
+    }
+
+    pub fn build<E: Display + Debug>(self) -> crate::Result<DetectContext<P, BM>, E> {
```
Malax, line 143: I don't think this should return a `libcnb::Result` but something specific for this builder. We use `Other` liberally for builder specific errors here. Bottom line, our builder errors are just `String`s instead of proper values.

src/error.rs
```diff
@@ -25,6 +25,9 @@ pub enum Error<E: Debug + Display> {
     #[error("Could not determine app directory: {0}")]
     CannotDetermineAppDirectory(std::io::Error),
 
+    #[error("{0}")]
+    Other(String),
```
Malax, line 29: This feels like an anti-pattern to me, especially when we introduce this for a _specific_ feature. I'm not saying we should do this, but even naming this `DetectContextBuilderError(String)` would be miles better. 

Ideally, we don't have this at all.

src/toml_file.rs
```diff
@@ -27,3 +27,7 @@ pub fn read_toml_file<A: DeserializeOwned>(path: impl AsRef<Path>) -> Result<A,
     let contents = fs::read_to_string(path)?;
     Ok(toml::from_str(&contents)?)
 }
+
+pub fn from_string<A: DeserializeOwned>(contents: String) -> Result<A, TomlFileError> {
```
Malax, line 31: `toml_file::from_string` makes no sense to me semantically. My best guess was that it creates a new file from a `String`, but where does TOML come into the equation? 

If we need this somewhere, we should just use `toml::from_str(&contents)` directly as it already does what we want.

Malax (changes requested):

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
...
 }
+
+enum BuildpackTomlArg<BM> {
+    Path(PathBuf),
+    Str(String),
+    BuildpackToml(BuildpackToml<BM>),
+}
+
+enum PlatformArg<P> {
+    Plat(P),
```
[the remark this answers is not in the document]
schneems: Seems good. I'll update.

src/detect.rs
schneems: I like being more specific. I try to differentiate in ruby via naming conventions https://github.com/heroku/buildpacks-ruby#naming-conventions.

src/detect.rs
schneems: I think we need three states. "user set path", "user says to use a temporary directory" and "user didn't say". 

Right now we're depending on injection of a temporary directory. If we want the builder to generate the temporary directories then it would also need to retain a reference to the original tempdir to avoid it being cleared right away. I brought this up in https://github.com/Malax/libcnb.rs/pull/93#issuecomment-925815869

With that API it would execute in a oneliner like this, but it's not clear that the reference to the builder is actually gone and the tempdirs would be cleaned up:

```
let detect_context = DetectContextBuilder::new("heroku-20")
    .buildpack_toml_from_file("../buildpack.toml")
    .with_temp_dirs()
    .build()
    .unwrap()
    .context;
```

I wouldn't want to default to `Temporary` either since in the case where we're using this in `runtime.rs` for example, I want it to error early if no app dir is set by accident.

src/detect.rs
schneems: >  It could be default behavior to create temporary directories for app_dir and platform_dir. 

This requirement of the builder to create the temporary does two things, it forces us to have `tempfile` as a prod dependency instead of runtime. (I'm fine with it, but its a consequence). It also would force us into the pattern I mentioned before where a builder must be stored as a variable before we can generate a context from it and it won't be able to be used as a one-liner.

That's not so different from my suggested TempContext API:

```
let temp = TempContext::new(
    "heroku-20", 
    include_str!("../../examples/example-02-ruby-sample/buildpack.toml")
); // <== Must assign to a variable to prevent tempdir from being cleaned
let build_context = temp.build;
```

Ultimately I'm fine either way. Both methods (dependency injection of temp dir versus generating a temp dir) require at least one assignment. So from the end-user experience, it's about the same.

Maybe there's some way to annotate that I expect the lifetime of the temporary directory to outlast the returned Context object? I'm unsure.

> a builder cannot be safely used concurrently since, when using temporary directories created by this function, every invocation of build will create contexts with the same paths.

Right now there's a 1-1 mapping of builders and contexts. The builder cannot be re-used after it generates a context as it is a consuming builder. I am in favor of making P and BM cloneable which would enable us to to generate multiple contexts from the same builder.  But I've not tried it.

> a builder cannot be safely used concurrently since, when using temporary directories created by this function, every invocation of build will create contexts with the same paths.

As I was writing this I was imagining that if you don't want anything shared with another context you would need to generate another builder. Right now that's a hard requirement due to the consuming nature of the builder. But if that changes it could change the way people want to use it.

I was thinking one builder per test. I wasn't thinking of trying to set some sort of global static value that could be called again and again. 

With that in mind using an enum like `Temporary` and generating the directory on demand in the `build()` call would allow for builder re-use which it seems like is a goal of yours. I can try out that tactic, but it won't pay off unless we can also get clonable `P` and `BM`.

> In Rust, from_ is a widely used prefix for converting A into B. This function does something very different and might be confusing since it goes against that convention.

I agree in terms of the naming being misleading. If I switch back to generating temporary directories internally and using an enum then It could be something like `with_tempdir()` and take no arguments.

src/detect.rs
schneems: Errors are an extreme weak spot i'll defer to @hone here.

src/error.rs
schneems: Agreed, this is very much *bad code* ™️ that I needed to get my code to compile.

src/toml_file.rs
schneems: I did this for murky reasons before. I think I had a hard time getting the code to generate a `libcnb::Error` and that was one way.

I'm able to get this to compile and I think it does what I want:

```rs
                BuildpackTomlArg::Str(contents) => toml::from_str(&contents)
                    .map_err(TomlFileError::TomlDeserializationError)
                    .map_err(Error::CannotReadBuildpackDescriptor),
```

It's not perfectly true since the message in deserialization error says that it couldn't deserialize the file. It will need to be updated. I'm guessing if we decide to keep the feature we'll have to add an error enum to `error.rs`

src/detect.rs
```diff
@@ -18,3 +23,196 @@ pub enum DetectOutcome {
     Pass(BuildPlan),
     Fail,
 }
+
+enum BuildpackTomlArg<BM> {
+    Path(PathBuf),
+    Str(String),
```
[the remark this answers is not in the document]
schneems: > Naming: Can we call this smth. like BuildpackTomlString?

Yes, easy change.

> In general, do we need the ability to use a string as the buildpack TOML in testing? It feels like this is promoting stringly typed programming.

I agree with the hesitation to accept ANY string to ANY method. Ultimately since the string passed in is put through a toml parser it must at least be valid toml which does give us some guarantees (though not until runtime).

Accepting a string has the benefit of being able to stub out contents for tests without having to write anything to disk.

The main reason I added this feature was to make pointing at relative files easy through `include_str!` which doesn't require any dependencies. 

I use it a lot in the tests:

```rs
        let context = DetectContextBuilder::new("heroku-20")
            .buildpack_toml_from_string(include_str!(
                "../examples/example-02-ruby-sample/buildpack.toml"
            ))
            .platform(GenericPlatform::from_path(".").unwrap())
            .app_dir(".")
            .buildpack_dir(".")
            .build::<std::io::Error>()
            .unwrap();

        fake_detect(context);
```

I tried to see if maybe there are other ways to point at files relative to the current file I found https://stackoverflow.com/questions/30511331/getting-the-absolute-path-from-a-pathbuf

Playing around I also found this works:

```
        let foo = PathBuf::from_str(file!()).unwrap().parent().unwrap().join("../examples/example-02-ruby-sample/buildpack.toml");
        println!("{:?}", foo);
        println!("{:?}", foo.exists());
```

But it's...gnarly. Maybe there's a better way to point at a relative path?

I know we can also generate paths from the cargo manifest dir:

```
    PathBuf::from(env!("CARGO_MANIFEST_DIR"))
        .join("test/fixtures")
        .join(fixture_name)
        .join("Procfile")
```

I'm personally in favor of keeping this option, I'm also happy to hear more concerns or objections.

src/detect.rs
schneems: I updated the error handling, please take another look.

src/detect.rs
schneems: I refactored all this logic, it no longer takes a path. Instead it generate and retains it's own tempdir.

src/detect.rs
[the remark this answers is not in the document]
schneems: I've re-written the logic but retained the string option. I'm now using enums and `From` to enable only needing one entry-point.

src/detect.rs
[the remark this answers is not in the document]
schneems: Done.

src/detect.rs
schneems: I went with `File` and `Directory`

src/detect.rs
schneems: I LOVE this approach. I refactored all the options out to custom enums. It makes the match logic really clean and simplifies the "setter" methods.

src/detect.rs
schneems: I've updated so the builder can generate contexts via oneliners and can be re-used. It now also generates and retains temp dirs.

src/detect.rs
schneems: Resolved by moving all this logic to a single `platform` function.

src/error.rs
schneems: I updated the error handling, please take a look.

src/toml_file.rs
schneems: I updated to use `toml` directly as noted before.

src/detect.rs
schneems: Done. Again, great suggestion 🙏

hone:
I get why you want the tempdir support, but I'm still not sold on `tempdir` now being a dependency for anyone pulling in `libcnb`.

src/detect.rs
hone: @Malax I refactored some stuff in #97, there's still many error cases covered by `libcnb::Result`, we could just make our own error type if there are other cases we want to cover.

schneems:
Aside from the tempdir concern and the LibError refactor do you have any other concerns?

> I'm still not sold on tempdir now being a dependency for anyone pulling in libcnb.

Manuel added a strong requirement that temp directories generated from the builder not be shared, so we need a way to generate a new directory for every `build()` or `detect()` call. We could do that by injecting a base tempdir and then creating directories in it:

```
use libcnb::ContextBuilder;
use tempfile::TempDir;

pub fn detect(context: libcnb::GenericDetectContext) {}

let temp_dir = TempDir::new();

let context = ContextBuilder::new("heroku-20")
    .buildpack_descriptor(include_str!("../examples/example-02-ruby-sample/buildpack.toml"))
    .create_temp_dirs_in(temp_dir.path().to_path_buf())
    .detect::<std::io::Error>()
    .unwrap();

detect(context);
```

Internally we create directories inside of that directory. Something like:

```
impl ContextBuilder
  	//...
	fn tempdir() -> PathBuf {
		let mut dir = None;
		while let None = dir { // Real code would have a timeout or a max iteration
			let tmp = self.temp_dir.join(PseudoRand::gen_string());
			if !tmp.exists() {
				dir = Some(tmp);
			}
		}
		dir.unwrap()
	}
```

Consequences: We can now drop tempfile requirement, we need a way to generate a random string (probably pull in a crate like `rand` https://docs.rs/rand/0.8.4/rand/).

Alternatives: 

- We could write our own tempfile implementation from scratch and vendor that. Here's the code that is used to generate temp directories in the `tmpfile` crate

```
impl TempDir {
    pub fn new() -> io::Result<TempDir> {
        Builder::new().tempdir()
    }
}

impl Drop for TempDir {
    fn drop(&mut self) {
        // Path is `None` if `close()` or `into_path()` has been called.
        if let Some(ref p) = self.path {
            let _ = remove_dir_all(p);
        }
    }
}

impl Builder {
    pub fn tempdir(&self) -> io::Result<TempDir> {
        self.tempdir_in(&env::temp_dir())
    }

    pub fn tempdir_in<P: AsRef<Path>>(&self, dir: P) -> io::Result<TempDir> {
        let storage;
        let mut dir = dir.as_ref();
        if !dir.is_absolute() {
            let cur_dir = env::current_dir()?;
            storage = cur_dir.join(dir);
            dir = &storage;
        }

        util::create_helper(dir, self.prefix, self.suffix, self.random_len, dir::create)
    }
}

pub fn create_helper<F, R>(
    base: &Path,
    prefix: &OsStr,
    suffix: &OsStr,
    random_len: usize,
    f: F,
) -> io::Result<R>
where
    F: Fn(PathBuf) -> io::Result<R>,
{
    let num_retries = if random_len != 0 {
        crate::NUM_RETRIES
    } else {
        1
    };

    for _ in 0..num_retries {
        let path = base.join(tmpname(prefix, suffix, random_len));
        return match f(path) {
            Err(ref e) if e.kind() == io::ErrorKind::AlreadyExists => continue,
            res => res,
        };
    }

    Err(io::Error::new(
        io::ErrorKind::AlreadyExists,
        "too many temporary files exist",
    ))
    .with_err_path(|| base)
}

mod env {
	pub fn temp_dir() -> PathBuf {
	    os_imp::temp_dir()
	}
}

mod os_imp {
	pub fn temp_dir() -> PathBuf {
	    crate::env::var_os("TMPDIR").map(PathBuf::from).unwrap_or_else(|| {
	        if cfg!(target_os = "android") {
	            PathBuf::from("/data/local/tmp")
	        } else {
	            PathBuf::from("/tmp")
	        }
	    })
	}
}

mod env {
	pub fn var_os<K: AsRef<OsStr>>(key: K) -> Option<OsString> {
	    _var_os(key.as_ref())
	}	

	fn _var_os(key: &OsStr) -> Option<OsString> {
	    os_imp::getenv(key)
	        .unwrap_or_else(|e| panic!("failed to get environment variable `{:?}`: {}", key, e))
	}
}


mod os_imp {
	pub fn getenv(k: &OsStr) -> io::Result<Option<OsString>> {
	    // environment variables with a nul byte can't be set, so their value is
	    // always None as well
	    let k = CString::new(k.as_bytes())?;
	    unsafe {
	        let _guard = env_read_lock();
	        let s = libc::getenv(k.as_ptr()) as *const libc::c_char;
	        let ret = if s.is_null() {
	            None
	        } else {
	            Some(OsStringExt::from_vec(CStr::from_ptr(s).to_bytes().to_vec()))
	        };
	        Ok(ret)
	    }
		
	}
}
```

It's not purely that simple, but that's the gist of it. I would say that I would much rather add a dependency than re-write this code.

Malax:
After thinking about this a bit more, I've come to the conclusion that this isn't something we should have in the library at this point. I understand the desire to test `detect` and `build` in a pure Rust test, but I don't think it's viable in most of the real-world use-cases outside of very simple buildpacks.

## Thoughts about pure Rust tests of `detect`/`build`

`detect` and `build` are such high level functions that testing them will always be an integration test: running `build` is equivalent of running the whole buildpack (minus detect).

This means that any test executing `build` has to replicate the environment provided by the CNB lifecycle, including environment variables, operating system, filesystem state, installed tools and changes made to the environment by previous buildpacks. Some of these can be implemented, others are **really hard** (like the OS).

Any assertions of such a test would require observing file-system changes and other side-effects. Looking at the signature of `build` reveals this fact easily:

```rust
fn build(context: BuildContext<P, BM>) -> Result<(), E>
```

The only thing that is returned is `Result<(), E>`, which can have the following values:

 - `Ok(())`
 - `Err(BuildpackSpecificErrorType)`

With these types, we cannot determine if a binary has been downloaded or a build tool ran. We _can check_ for a buildpack specific error before it's handled by the [ErrorHandler](https://docs.rs/libcnb/0.3.0/libcnb/trait.ErrorHandler.html) but that's about it. Everything else needs to be asserted by watching for side-effects.

When we look at the input type for `build` (`BuildContext`), we can see it contains the following information:
- Path to the application
- Path to the buildpack
- Path to `buildpack.toml` (which can be inferred from the buildpack directory)
- Path to the layers directory
- The buildpack plan generated from the set of all buildpacks involved in this build
- A `Platform` which is an abstraction over the platform directory from the CNB spec

All of the above are either command line arguments to `pack` or provided by the CNB lifecycle implementation at runtime. From this, we can deduct that we effectively have to re-implement `pack` and the CNB lifecycle if we want to run `build` from inside a test written in pure Rust. I'd argue that it is not worth the effort to do so when we can just use `pack` and the CNB lifecycle in an integration test with a tool similar to Hatchet or Cutlass.

Instead, I suggest we direct our efforts to implement Cutlass in Rust so it is usable from [Rust integration tests](https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html) and ensure it can run our tests as fast as possible: enable parallel tests, enable the usage of a remote docker deamon to run many containers at once regardless of the machine the tests run on, etc.

It's imporant to note that all of the arguments above are language and library indepentent. They're an inherent part of the problem domain (buildpacks).

## Thoughts about the ability to construct instances of `BuildContext` and `DetectContext`

First of all, it's already possible to create instances of both types since they and all their fields are public:

```rust
pub struct BuildContext<P: Platform, BM> {
    pub layers_dir: PathBuf,
    pub app_dir: PathBuf,
    pub buildpack_dir: PathBuf,
    pub stack_id: String,
    pub platform: P,
    pub buildpack_plan: BuildpackPlan,
    pub buildpack_descriptor: BuildpackToml<BM>,
}
```

```rust
pub struct DetectContext<P: Platform, BM> {
    pub app_dir: PathBuf,
    pub buildpack_dir: PathBuf,
    pub stack_id: String,
    pub platform: P,
    pub buildpack_descriptor: BuildpackToml<BM>,
}
```

And it's important that they can be constructed so we can use them in tests for libcnb itself. With regards to implementing a builder for them: Since all fields are required, I don't see the value. 

Either the builder would have to take all required values in its `new` function, allowing only to override individual ones with methods of the builder or the builder can be in an invalid state where it must error out if not all fields are set. Both options look pointless to me.

schneems:
I think strategically the cost/benefit of moving forward here is low-benefit, high-cost. Pulling functionality out of `build` and `detect` is trivially easy, also writing my own implementation of a mocked context wasn't hard either for people who really want to do it.

I actually think our gap right now is in unit testing. That's what motivated my work here and on https://github.com/heroku/libherokubuildpack/pull/32. 

I want to get to a place where I CAN exercise all my interfaces in the Rust context in the same way that I can exercise all the interfaces of the Ruby buildpack in Ruby (yes, all of them). https://github.com/heroku/buildpacks-ruby/tree/main/spec/unit. (Also while I call them "unit" tests, they are very much testing for side effects. Ed suggested maybe calling them "rust tests" or "language tests" to perhaps be more on the same page.)

We can already get by with cutlass as-is. Ideally, we will eventually re-write it in Rust, but that's not a blocker to buildpack work. 

## My immediate blockers are:

- 0.6 support in libcnb is needed for the procfile buildpack before I can hit the definition-of-done https://salesforce.quip.com/wq0iA34PyWIO
- Add cutlass support to the procfile buildpack (also needed for DoD). This isn't strictly blocked by anything but eventually we need to figure out the parallel compilation story eventually (build.sh) https://salesforce.quip.com/veWYASMAibmN#XMEAAA2k8iv.
- Get a testable interface for logging upstreamed somewhere. I would like to get this nailed down before starting on the ruby buildpack in Rust as I don't want to have to rip out logging and re-add it if it changes later, also I want to start writing assertions on the log output as I view our logs as a customer interface https://salesforce.quip.com/qeFRADYvnHo5#CIVAAALZvN5

## Medium term blockers

- Still around unit-testing. I think we should aim to unit test the invoker buildpack https://github.com/heroku/buildpacks-jvm/tree/main/buildpacks/jvm-function-invoker. 

I believe this will give us insight into either patterns we should be encouraging, or possibly opportunities to add or change libcnb interfaces. I certainly understand that types give us a lot of safety, I also want to be able to exercise code where reasonable to give faster feedback around development. 

For some extra context where I'm coming from, there was a bug in our LayerContentMetadata implementation that didn't show up without being exercised. It's not that I need everything to be tested, its that I want the ability to test everything should the need arise.

- Unit test other buildpacks. 
 
David started on the PHP buildpack for instance, Lillian was working on re-writing the CNB "Ruby" example in Rust. I think going through that exercise will further inform our decisions. Eventually writing the Python, Ruby, and PHP buildpacks will be another feedback loop. My personal worry is that I don't want to invest a bunch of time on interfaces in my own buildpack if I just have to re-write them later so they can be tested.

- Internal unit tests to libcnb. 

Specifically the runtime and layer lifecycle. Ideally every file would have one unit test to get started so that if someone needs to add more for new functionality, they're not starting from nothing. This is actually where I think we'll need the work covered here (or something like it) as exercising runtime and layer lifecycle will require a context objects. Again, it doesn't have to specifically be __this__ implementation: it could be a different crate, it could be an internal only struct helper class similar to procfile buildpack, etc. But I do think it will be needed.

> And it's important that they can be constructed so we can use them in tests for libcnb itself. With regards to implementing a builder for them: Since all fields are required, I don't see the value.

Exactly.

Also to iterate, I'm not saying we need codecov == 100%. I am saying that writing tests is hard, and writing tests for someone else's implementation is even harder. When It came to me doing the work in this PR I didn't just have to write the logic, I also had to come up with the test strategy from scratch as BuildContext doesn't have tests https://github.com/Malax/libcnb.rs/blob/main/libcnb/src/build.rs.

For extra context, I am used to building out stub/mock data structures along side of their original implementations, which also gives me feedback into the original design. For instance when I implemented a "metadata" store https://github.com/heroku/buildpacks-ruby/blob/main/lib/heroku_buildpack_ruby/metadata/cnb.rb I also implemented a test-only version called `Null` https://github.com/heroku/buildpacks-ruby/blob/cac3caf0d31b4a69c6ce1743ad63b7e948a4764c/lib/heroku_buildpack_ruby/metadata.rb#L47-L51. Thinking about how I'll get data into and out of the test structure means I end up with a production structure that is more ergonomic to test.

## Long/medium term

- Reduced toil deploy strategy https://salesforce.quip.com/J3ZpAihiOAaZ

This doc outlines some ways we could reduce the current deploy strategy, however with some sort of a re-architecture maybe we can reduce it even further. One comment I made to Ed, was that if all the buildpacks are in Rust then we don't have to rely on ONLY code sharing between buildpacks using the build plan interface. Would that actually reduce deploy toil? I don't know, but I think that we want to start thinking about it now before we get too far down one approach to turn back.

- Re-write cutlass in Rust

I think the value here is to remove a ruby dependency from the ecosystem to have N-1 languages buildpack authors need to be acquainted with to dev. In the short term, we already know it and already have it. There might be more benefits I'm not seeing, but I don't think this is where I want to invest my short term time (but also I wouldn't dissuade anyone else).

My strong preference is to have such a robust local/language/rust test story that we only need minimal integration cases. For instance in the Ruby buildpack there are only 5 cutlass tests https://github.com/heroku/buildpacks-ruby/blob/main/spec/cnb/buildpack_spec.rb while in the legacy buildpack there are 50 

```
⛄️ 2.7.4 🚀 /Users/rschneeman/Documents/projects/work/buildpacks/heroku-buildpack-ruby (main)
$ be rspec spec/hatchet --dry-run --format json | jq '.examples | length'
51
```

## Build the tracks as we go

The "medium term" section is a bit more aspirational. I'm assuming as we do the work, it will guide us for where the next work should be done. I understand that my approach to a dynamic typeless language doesn't 100% translate, but I still see value in this approach...if there is none or if it's meaningfully different, then I'll experience it first hand and be able to update my approach as I/we go. 

## TLDR;

Let's close this PR out so we're all on the same page. Welcome back, I'm happy to be having some of these discussions!

hone:
> And it's important that they can be constructed so we can use them in tests for libcnb itself. With regards to implementing a builder for them: Since all fields are required, I don't see the value.
> 
> Either the builder would have to take all required values in its `new` function, allowing only to override individual ones with methods of the builder or the builder can be in an invalid state where it must error out if not all fields are set. Both options look pointless to me.

@Malax so I'd actually push back on this point specifically. There are a decent amount of "Generic" values here. Builders also have a use IMO when you have defaults and having to do the `GenericContext`/`GenericPlatform` dance is kind of annoying. Builders can provide value if there are default fields.

PR #97, opened 2021-09-25
refactor Error enums

hone:
this commit splits the large Error enum into some smaller ones
* LibError which will happen usually due to bugs from Platform errors
* DataError this will happen when parsing bad data types from the spec

This should help #96 remove the unnecessary Error generics on `fn buildl()`, since it will never have a `BuildpackError(E)`.

```
pub fn build(self) -> Result<DetectContext<P, BM>, crate::error::LibError>
```

The biggest downside of doing this is nested enum handling.

### DIFF
--- a/libcnb/src/error.rs
+++ b/libcnb/src/error.rs
@@ -16,11 +16,38 @@ pub type Result<T, E> = std::result::Result<T, Error<E>>;
 /// An error that occurred during buildpack execution.
 #[derive(thiserror::Error, Debug)]
 pub enum Error<E: Debug + Display> {
+    #[error("libcnb error: {0}")]
+    LibError(LibError),
+    #[error("data format error: {0}")]
+    DataError(DataError),
+    #[error("Buildpack error: {0}")]
+    BuildpackError(E),
+}
+
+#[cfg(feature = "anyhow")]
+impl From<anyhow::Error> for Error<anyhow::Error> {
+    fn from(error: anyhow::Error) -> Self {
+        Error::BuildpackError(error)
+    }
+}
+
+impl<E: Debug + Display> From<LibError> for Error<E> {
+    fn from(error: LibError) -> Self {
+        Self::LibError(error)
+    }
+}
+
+impl<E: Debug + Display> From<LayerLifecycleError> for Error<E> {
+    fn from(error: LayerLifecycleError) -> Self {
+        Self::LibError(LibError::LayerLifecycleError(error))
+    }
+}
+
+/// An error that occurred from libcnb
+#[derive(thiserror::Error, Debug)]
+pub enum LibError {
     #[error("Layer lifecycle error: {0}")]
     LayerLifecycleError(#[from] LayerLifecycleError),
-
-    #[error("Process type error: {0}")]
-    ProcessTypeError(#[from] ProcessTypeError),
 
     #[error("Could not determine app directory: {0}")]
     CannotDetermineAppDirectory(std::io::Error),
@@ -42,14 +69,10 @@ pub enum Error<E: Debug + Display> {
 
     #[error("Cannot write build plan: {0}")]
     CannotWriteBuildPlan(TomlFileError),
-
-    #[error("Buildpack error: {0}")]
-    BuildpackError(E),
 }
 
-#[cfg(feature = "anyhow")]
-impl From<anyhow::Error> for Error<anyhow::Error> {
-    fn from(error: anyhow::Error) -> Self {
-        Error::BuildpackError(error)
-    }
+#[derive(thiserror::Error, Debug)]
+pub enum DataError {
+    #[error("Process type error: {0}")]
+    ProcessTypeError(#[from] ProcessTypeError),
 }
--- a/libcnb/src/layer_lifecycle.rs
+++ b/libcnb/src/layer_lifecycle.rs
@@ -318,12 +318,10 @@ pub fn execute_layer_lifecycle<
     layer_lifecycle.on_lifecycle_end();
 
     match context.read_layer_content_metadata(&layer_name) {
-        Err(toml_file_error) => Err(Error::LayerLifecycleError(
-            LayerLifecycleError::CannotReadLayerContentMetadata(toml_file_error),
-        )),
-        Ok(None) => Err(Error::LayerLifecycleError(
-            LayerLifecycleError::CannotFindLayerMetadataAfterLifecycle(),
-        )),
+        Err(toml_file_error) => {
+            Err(LayerLifecycleError::CannotReadLayerContentMetadata(toml_file_error).into())
+        }
+        Ok(None) => Err(LayerLifecycleError::CannotFindLayerMetadataAfterLifecycle().into()),
         Ok(Some(metadata)) => layer_lifecycle
             .layer_lifecycle_data(&layer_path, metadata)
             .map_err(Error::BuildpackError),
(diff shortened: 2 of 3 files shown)


src/runtime.rs
```diff
@@ -215,13 +216,13 @@ fn parse_build_args_or_exit() -> BuildArgs {
...
-        .map_err(Error::CannotDetermineBuildpackDirectory)
+        .map_err(|e| LibError::CannotDetermineBuildpackDirectory(e).into())
         .map(PathBuf::from)
 }
 
 fn read_buildpack_toml<BM: DeserializeOwned, E: Display + Debug>() -> Result<BuildpackToml<BM>, E> {
     read_buildpack_dir().and_then(|buildpack_dir| {
         read_toml_file(buildpack_dir.join("buildpack.toml"))
-            .map_err(Error::CannotReadBuildpackDescriptor)
+            .map_err(|e| LibError::CannotReadBuildpackDescriptor(e).into())
```
schneems, line 226: I want to try to explain this back. 


Before we were calling

```
.map_err(Error::CannotReadBuildpackDescriptor)
```

Before this was defined here:

```rs
pub enum Error<E: Debug + Display> {
	// ...
    #[error("Cannot read buildpack descriptor (buildpack.toml): {0}")]
    CannotReadBuildpackDescriptor(TomlFileError),
	// ...
}
```

Now it is defined as:

```rs
pub enum Error<E: Debug + Display> {
    #[error("libcnb error: {0}")]
    LibError(LibError),

    // ...
} 


/// An error that occurred from libcnb
pub enum LibError {
	//...
    #[error("Cannot read buildpack descriptor (buildpack.toml): {0}")]
    CannotReadBuildpackDescriptor(TomlFileError),
    // ...
}
```

And we are calling:

```rs
             .map_err(|e| LibError::CannotReadBuildpackDescriptor(e).into())
```

This is needed because before the error was passed to `Error::CannotReadBuildpackDescriptor` and it was formatted with:

```rs
    #[error("Cannot read buildpack descriptor (buildpack.toml): {0}")]
```

And it's bound by generic Display + Debug and also by `libcnb::Result`

```rs
pub type Result<T, E> = std::result::Result<T, Error<E>>;
```

Because we're using a concrete error enum, we cannot just pass the error to the enum. We've also got to tell rust how to convert it to generic `Error<E>`. And the way to do that is calling `into()`. Adding this syntax means we cannot implicitly pass the error and have to switch to using `|e|` syntax as well.

I'm not sure what code allows the LibError to be coerced into a:

```rs
pub enum Error<E: Debug + Display> {
	//...
}
```

On second glance it looks like it's here:

```rs
impl<E: Debug + Display> From<LibError> for Error<E> {
    fn from(error: LibError) -> Self {
        Self::LibError(error)
    }
}
```

Which tells Error<E> how to convert LibError into Error<E> and this is what gets invoked when `into()` is called. https://doc.rust-lang.org/std/convert/trait.Into.html

Let me know if that's correct or if anything in my understanding needs to be adjusted.

src/runtime.rs
```diff
@@ -215,13 +216,13 @@ fn parse_build_args_or_exit() -> BuildArgs {
 
 fn read_buildpack_dir<E: Display + Debug>() -> Result<PathBuf, E> {
     env::var("CNB_BUILDPACK_DIR")
-        .map_err(Error::CannotDetermineBuildpackDirectory)
+        .map_err(|e| LibError::CannotDetermineBuildpackDirectory(e).into())
         .map(PathBuf::from)
 }
 
 fn read_buildpack_toml<BM: DeserializeOwned, E: Display + Debug>() -> Result<BuildpackToml<BM>, E> {
```
schneems, line 223: Is it possible to mix specific errors with generic errors? Could we return a `Result<BuildpackToml<BM>, crate::error::LibError>` here? Would the `into` conversion have to go somewhere else or would it just not be possible?

I'm thinking since we know all the errors that this can produce and none of them come from outside sources this could be more specific, but my understanding about the details of why that might not be possible as well as the tradeoffs are fuzzy.

schneems (commented):
> The biggest downside of doing this is nested enum handling.

This is what you're referring to here?

```rs
             .map_err(|e| LibError::CannotReadBuildpackDescriptor(e).into())
```

It's still pretty readable and it's not happening in too many places. Being able to be specific about places that handle only internal errors versus needing to be mixed with generic user-executed functions seems like a big win. Being able to eliminate the generic in my PR and `pub fn build(self) -> Result<DetectContext<P, BM>, crate::error::LibError>` would make the end API better i.e.

Instead of:

```
    /// let temp_dir = tempfile::tempdir().unwrap();
    /// let context = DetectContextBuilder::new("heroku-20")
    ///     .buildpack_toml_from_string(include_str!("../examples/example-02-ruby-sample/buildpack.toml"))
    ///     .from_path(&temp_dir)
    ///     .build::<std::io::Error>() // <=== Here
    ///     .unwrap();
```

We could

```
    /// let temp_dir = tempfile::tempdir().unwrap();
    /// let context = DetectContextBuilder::new("heroku-20")
    ///     .buildpack_toml_from_string(include_str!("../examples/example-02-ruby-sample/buildpack.toml"))
    ///     .from_path(&temp_dir)
    ///     .build() // <=== Here
    ///     .unwrap();
```

I'm in favor of any change that gets me closer to a more ergonomic API. The modifications to the existing codebase seem somewhat minimal. I'm generally : +1: but I'm curious to hear if my understanding of the modifications were correct.

src/runtime.rs
hone: > Is it possible to mix specific errors with generic errors?

Yes and no. In any individual function, you need a single Error type. There a things to help with that though like `map_err` and enums. This is why a lot of errors are enums.

To your point it's probably worth revisiting all these and return `LibError` instead of `crate::error::Error` since it doesn't touch user input.

hone:
I'd love to sync with @Malax now that he's back on this and #96.

Malax:
I rebased this after #124

Malax:
I like the idea of using the most specific error type wherever possible, it improves composability a bunch. Given that this PR will no longer apply cleanly (or at all) after the changes we made in the last weeks, I suggest we close this PR and file issues for all functions that return an error type that is too broad.

Do @schneems or @hone remember what specific API caused you to to open this PR? I understand this is the outcome of a pairing session of you both.

I have the gut feeling that the issue is not as bad (just a feeling + the changeset of this PR). I agree that adding an additional layer of indirection for errors is a bit annoying and if we can avoid it, we should. Happy to be proven wrong on this one though. If we need the indirection for get more specific errors for some functions/API, we should add it.

hone:
@Malax it's specific to the fact that we're intermixing Buildpack Author errors and the library's errors in `crate::Result`. It's a large enum with ONE value being the Buildpack Author's custom errors. If you look at the new `Buildpack` trait, both `detect` and `build` should not REQUIRE a `crate::Result` as the return value? Why is it not just:

```
fn detect(&self, context: DetectContext<Self>) -> Result<DetectResult, Self::Error>;
```

What's the benefit of the indirection on the Buildpack Author's code?

edmorley:
> If you look at the new `Buildpack` trait, both `detect` and `build` should not REQUIRE a `crate::Result` as the return value? Why is it not just:

So I might be misunderstanding either what you are saying or the code (or both!). 

But it seems that libcnb-specific errors can happen even within the buildpack's `build()`. For example in `build()` it is expected that the buildpack author call `context.handle_layer()` - which can return both custom errors from the `Layer`, but also libcnb-specific errors such as failure writing the layer toml to disk, which is why `handle_layer()` has to have a `crate::Result` return value:
https://github.com/Malax/libcnb.rs/blob/ca78c911d2351c68a3386b7c029cc44c0aae24ca/libcnb/src/build.rs#L99

...and for `build()` to be able to propagate that back, it needs to too.

As such, it doesn't seem like we would ever be able to change the `build()` return type to just `std::result::Result`?

And if we're not able to do that, then is this PR still something we want?

hone:
Yeah, let's close this out for now.

PR #107, opened 2021-10-19
Allow process type to contain `.`

### LINKED ISSUE #105: `ProcessType` currently forbids names containing `.` whilst the spec permits them
The `ProcessType` validation here forbids the `.` character:
https://github.com/Malax/libcnb.rs/blob/f46bdda85838c9cc21cd1de3243d4160b85afc12/src/data/launch.rs#L111-L117

However the spec allows them:
https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml

> MUST only contain numbers, letters, and the characters ., _, and -.

 GUS-W-10055113

schneems:
From the spec:

> MUST only contain numbers, letters, and the characters ., _, and -.

https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml

A dot (`.`) is allowed.

Fixes #105. 
GUS-W-10055113

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,4 +2,6 @@
 
 ## [Unreleased]
 
+- Allow ProcessType to contain a dot (`.`) character
+
 ## [0.3.0] 2021/09/17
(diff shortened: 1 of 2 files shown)


schneems:
> @schneems Could you move the issue number from the PR title to the body? IMO it clutters the title otherwise :-)

Sure. I generally forget to close issues afterward unless GitHub does it for me.

edmorley:
I've added the #105 issue reference to the body.

PR #109, opened 2021-10-19
Add support for "any" stack `*` in StackId

### LINKED ISSUE #104: `StackId` should support the "any" stack (`"*"`)
As of Buildpack API 0.5, a buildpack can declare support for any stack using a stack ID of `"*"`:
https://github.com/buildpacks/rfcs/blob/main/text/0056-any-stack-buildpacks.md
https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpack-implementations
https://github.com/buildpacks/spec/blob/main/platform.md#stack-id

However this isn't currently supported:

```rust
use std::str::FromStr;
use libcnb::data::buildpack::BuildpackId;

let valid = BuildpackId::from_str("*");
assert_eq!(valid.unwrap().as_str(), "*");
```

Gives:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidBuildpackId("*")', src/data/buildpack.rs:8:18
```

Due to:
https://github.com/Malax/libcnb.rs/blob/f46bdda85838c9cc21cd1de3243d4160b85afc12/src/data/buildpack.rs#L209-L224

In addition, there is no validation that `mixins` is not set when the any stack is used (which is forbidden by the spec).

GUS-W-10055742

schneems:
When a `Stack` struct has a `StackId` of `*` then it should have no mixins.

This is validated by a "shadow" struct deserializing working from https://dev.to/equalma/validate-fields-and-types-in-serde-with-tryfrom-c2n.

Based on the comment from https://github.com/serde-rs/serde/issues/939#issuecomment-939514114

In my own words it works like this: We create a duplicate struct and use serde's normal `derive(Deserialize)` on it. Then we specify that the `Stack` struct should be converted using the duplicate struct with `#[serde(try_from = "StackUnchecked")]. Finally we implement `try_from` and include our special logic that checks to see we don't have a `*` stack id AND any mixins.

I also hard wrapped a few comments and put backpacks around special characters in the error output to make them clearer

Close #104 

GUS-W-10055742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Stack id in `buildpack.toml` can now be `*` indicating "any" stack
 - LayerContentMetadata values (build, cache, launch) are now under a "types" key
 - Allow ProcessType to contain a dot (`.`) character
 
--- a/src/data/buildpack.rs
+++ b/src/data/buildpack.rs
@@ -3,6 +3,7 @@ use lazy_static::lazy_static;
 use regex::Regex;
 use semver::Version;
 use serde::{de, Deserialize};
+use std::convert::TryFrom;
 use std::fmt::{Display, Formatter};
 use std::{fmt, str::FromStr};
 use thiserror;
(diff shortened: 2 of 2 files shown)


schneems:
## Error verification

I would also like to verify that the error contents are the ones we expect rather than just `is_err()`. Previously I made a mistake in my toml and was using `mixins = [yolo]` instead of `mixins = ["yolo"]` and it was failing because it was invalid TOML, (not because of my logic). Ideally, I would like a test that accounts for that kind of false positive.

I'm not totally sure the best way to do that. Right now the error returned is a serde error:

```
thread 'data::buildpack::tests::cannot_use_star_stack_id_with_mixins' panicked at 'called `Result::unwrap()` on an `Err` value: Error { inner: ErrorInner { kind: Custom, line: Some(12), col: 0, at: Some(122), message: "Stack with id `*` MUST not contain mixins. mixins: [yolo]", key: ["stacks"] } }', src/data/buildpack.rs:427:16
```

## Other deserialization options

@hone  previously implemented other deserizliation logic on some other structs in https://github.com/Malax/libcnb.rs/commit/b7372671380c0c4d1168450bf5546deabd5d2233. I do not know if there are advantages or disadvantages to implementing Deserialize from scratch. It seems like this does the same thing, but is shorter. If they're effectively identical, I would be interested in re-writing the existing deserialization logic using this shadow/duplicate/temp method.

hone:
> @hone previously implemented other deserizliation logic on some other structs in [b737267](https://github.com/Malax/libcnb.rs/commit/b7372671380c0c4d1168450bf5546deabd5d2233). I do not know if there are advantages or disadvantages to implementing Deserialize from scratch. It seems like this does the same thing, but is shorter. If they're effectively identical, I would be interested in re-writing the existing deserialization logic using this shadow/duplicate/temp method.

I actually didn't know `#[serde(try_from)]` existed at all. It looks like it's "relatively" new when @Malax and I started on this path. It wasn't merged in until [mid 2019](https://github.com/serde-rs/serde/pull/1526). I'm :+1: in removing my hand rolled logic. I copied it from me implementing this in 2018 :P.

Issue #55, opened 2021-09-08
Fix clippy::single-match-else warnings


edmorley:
I ran Clippy's pedantic lints (which are disabled by default since they are typically subjective or have high false positive rates) against this repository out of curiosity, and am filing a few issues for some of the lints that warned that seemed relevant (in part to give some easy issues for people to work through to get involved with this library).

Feel free to close this out if not wanted :-)

See:
https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else

```
$ cargo clippy --all-targets --all-features -- -W clippy::single-match-else
warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
   --> src/runtime.rs:166:5
    |
166 | /     match args.as_slice() {
167 | |         [_, platform_dir_path, build_plan_path] => DetectArgs {
168 | |             platform_dir_path: PathBuf::from(platform_dir_path),
169 | |             build_plan_path: PathBuf::from(build_plan_path),
...   |
175 | |         }
176 | |     }
    | |_____^
    |
    = note: requested on the command line with `-W clippy::single-match-else`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else
help: try this
    |
166 |     if let [_, platform_dir_path, build_plan_path] = args.as_slice() { DetectArgs {
167 |         platform_dir_path: PathBuf::from(platform_dir_path),
168 |         build_plan_path: PathBuf::from(build_plan_path),
169 |     } } else {
170 |         eprintln!("Usage: detect <platform_dir> <buildplan>");
171 |         eprintln!("https://github.com/buildpacks/spec/blob/main/buildpack.md#detection");
  ...

warning: you seem to be trying to use `match` for destructuring a single pattern. Consider using `if let`
   --> src/runtime.rs:181:5
    |
181 | /     match args.as_slice() {
182 | |         [_, layers_dir_path, platform_dir_path, buildpack_plan_path] => BuildArgs {
183 | |             layers_dir_path: PathBuf::from(layers_dir_path),
184 | |             platform_dir_path: PathBuf::from(platform_dir_path),
...   |
191 | |         }
192 | |     }
    | |_____^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else
help: try this
    |
181 |     if let [_, layers_dir_path, platform_dir_path, buildpack_plan_path] = args.as_slice() { BuildArgs {
[...]

PR #119, opened 2021-11-01
change match to if let else

lillianzhang331:
https://rust-lang.github.io/rust-clippy/master/index.html#single_match_else
Addressing issue #55

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,8 +30,6 @@
 #![allow(clippy::redundant_closure_for_method_calls)]
 // https://github.com/Malax/libcnb.rs/issues/58
 #![allow(clippy::semicolon_if_nothing_returned)]
-// https://github.com/Malax/libcnb.rs/issues/55
-#![allow(clippy::single_match_else)]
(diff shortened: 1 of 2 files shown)


edmorley (changes requested):
CI is failing in the rustfmt check:
https://github.com/Malax/libcnb.rs/runs/4073032849?check_suite_focus=true

Just checking you have format on save enabled in your IDE? :-) (https://salesforce.quip.com/geXbAMwRn5Pb)

Also, could you remove the `allow` directive from `lib.rs` that is currently turning off this rule, so that it then runs when clippy is run? :-)
https://github.com/Malax/libcnb.rs/blob/4121fbfc0f87e88a043db0d4d6f06ac9495aa610/src/lib.rs#L33-L34

edmorley (approved):
Thank you for the changes :-)

Could you squash these locally and force push to your branch? Unfortunately Manuel has disabled squash and merge for this repository, which means I can't squash at time of merge.

hone (approved):

Issue #20, opened 2021-03-26
move to Cargo workspaces so examples compile with rest of code


hone:

edmorley:
This was fixed by #128.

PR #124, opened 2021-11-04
Introduce workspace

Malax:
This makes the repository a [Cargo workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html) and moves everything into the `libcnb` crate in this repo.

This is required to tackle #20, #97 and the general wish to be able to use `data` without the whole framework. Later down the line we will introduce additional crates that will be part of the workspace. This includes `libcnb-testing`, `libcnb-cargo` (Cargo plugin to package buildpacks and generate scaffolds for new ones), `libcnb-macros` (procedual macros which must be in their own crate) and more.

For end-users, nothing changes since `libcnb` is still the same crate as it was before. It just lives in a different directory in the repository.


edmorley (approved):
Should the `examples/` directory remain in the root? In the future, presumably the examples might utilise a few of the crates? (eg `libcnb` + `libcnb-testing`)

Malax:
I kept this PR as simple as possible since it changes the layout so drastically. Examples in the root do work, but there are some caveats I haven't fully worked out yet. I think the examples need some love anyway so that would be a good opportunity to review how they work. See: 

- https://github.com/rust-lang/cargo/issues/7467
- #20

PR #125, opened 2021-11-04
Move data module into separate crate

### LINKED ISSUE #81: libcnb-data
I'd like to use the data structures of libcnb independent of the rest of the library. Thoughts on splitting this apart?

Malax:
This allows the low-level bindings to the CNB spec to be used without the higher-level framework and its dependencies. This does not break users of `libcnb` since `libcnb-data` is re-exported as `data` just like before.

Closes #81


libcnb-data/src/layer_content_metadata.rs
```diff
@@ -62,11 +61,11 @@ pub struct LayerContentMetadata<M> {
     pub metadata: M,
 }
 
-impl Default for LayerContentMetadata<GenericMetadata> {
+impl Default for LayerContentMetadata<Option<toml::Value>> {
     fn default() -> Self {
         Self {
             types: LayerContentTypeTable::default(),
-            metadata: GenericMetadata::default(),
+            metadata: Default::default(),
```
edmorley, line 68: Clippy pedantic says:

```
$ cargo clippy --all-targets --all-features -- --deny warnings
    Checking libcnb-data v0.3.0 (/Users/emorley/src/libcnb.rs/libcnb-data)
error: calling `std::option::Option::default()` is more clear than this expression
  --> libcnb-data/src/layer_content_metadata.rs:68:23
   |
68 |             metadata: Default::default(),
   |                       ^^^^^^^^^^^^^^^^^^ help: try: `std::option::Option::default()`
   |
   = note: `-D clippy::default-trait-access` implied by `-D warnings`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#default_trait_access
```

libcnb/Cargo.toml
```diff
@@ -12,13 +12,11 @@ readme = "../README.md"
...
 
 [dependencies]
-lazy_static = "^1.4.0"
-regex = "^1.5.4"
-semver = { version = "^1.0.4", features = ["serde"] }
 serde = { version = "^1.0.126", features = ["derive"] }
 thiserror = "^1.0.26"
 toml = "^0.5.8"
 anyhow = { version = "^1.0.42", optional = true }
+libcnb-data = { path = "../libcnb-data" }
```
edmorley, line 19: Does there need to be a version number here in addition to the `path`, so this works when published to crates.io? 
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-path-dependencies

> However, crates that use dependencies specified with only a path are not permitted on crates.io. 

The solution appears to be to set both:
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#multiple-locations

libcnb-data/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "libcnb-data"
+version = "0.3.0"
```
edmorley, line 3: Should this start at `0.1.0` given it's a new crate and we presumably won't keep them in lock-step forever? I can't decide / happy either way.

edmorley (changes requested):
Could you enable the additional lints that are enabled in `libcnb/src/lib.rs` in the new `libcnb-data/src/lib.rs` too?

Checking locally the following works (only a subset of the `allow`s are required compared to `libcnb`; I also confirmed that there were no redundant `allow`s that could be removed from the original `lib.rs`):

```rust
// Enable rustc and Clippy lints that are disabled by default.
// https://doc.rust-lang.org/rustc/lints/listing/allowed-by-default.html#unused-crate-dependencies
#![warn(unused_crate_dependencies)]
// https://rust-lang.github.io/rust-clippy/stable/index.html
#![warn(clippy::pedantic)]
// Re-disable pedantic lints that are currently failing, until they are triaged and fixed/wontfixed.
// https://github.com/Malax/libcnb.rs/issues/53
#![allow(clippy::missing_errors_doc)]
// https://github.com/Malax/libcnb.rs/issues/83
#![allow(clippy::module_name_repetitions)]
// https://github.com/Malax/libcnb.rs/issues/57
#![allow(clippy::must_use_candidate)]
// https://github.com/Malax/libcnb.rs/issues/61
#![allow(clippy::redundant_closure_for_method_calls)]
```

libcnb/Cargo.toml
Malax: Good catch! I wasn't aware of that restriction when publishing to crates.io. There seems to be an effort going on to simplify use-cases like ours, but it hasn't landed yet: https://github.com/rust-lang/cargo/issues/3931

libcnb-data/Cargo.toml
Malax: +1 on being unsure. It's certainly easier for now to keep the versions in lockstep to reduce possible confusion.

libcnb-data/src/layer_content_metadata.rs
Malax: I changed this for now to unblock this PR and we can have this discussion later.

edmorley: I'm totally fine with disabling pedantic lints that we disagree with - they are in the pedantic category after all :-) 

The reason I enabled the pedantic lints was so that we could get the advantage of the 80% of them that are useful, with the understanding that it's ok to disable subsets as we see fit.

edmorley (approved):

PR #128, opened 2021-11-08
Fix examples

Malax:
#125 broke the examples by omitting them from the workspace. This PR fixes this.

This results in the examples being build on CI, which is great and wasn't done before. But that revealed that they no longer compile and/or lint correctly. This was fixed as well.


edmorley (approved):
Yey linting/... now working for these!

Issue #86, opened 2021-09-13
API design of `cnb_runtime` causes clippy::unnecessary_wraps lint warnings


edmorley:
If a buildpack that's enabled the pedantic `clippy::unnecessary_wraps` lint has a `detect` handler that never returns an error (for example since it uses `std::fs::Path::exists()` due to `try_exists()` being nightly-only), then a lint warning is reported that cannot be resolved other than `allow`ing it:

```rust
pub fn detect(context: GenericDetectContext) -> Result<DetectOutcome, BuildpackError> {
    let procfile_path = context.app_dir.join("Procfile");

    if procfile_path.exists() {
        let buildplan = BuildPlan::new();
        Ok(DetectOutcome::Pass(buildplan))
    } else {
        Ok(DetectOutcome::Fail)
    }
}
```

```
$ cargo clippy
warning: this function's return value is unnecessarily wrapped by `Result`
  --> src/detect.rs:5:1
   |
5  | / pub fn detect(context: GenericDetectContext) -> Result<DetectOutcome, BuildpackError> {
6  | |     let procfile_path = context.app_dir.join("Procfile");
7  | |
8  | |     if procfile_path.exists() {
...  |
13 | |     }
14 | | }
   | |_^
   |
note: the lint level is defined here
  --> src/main.rs:5:9
   |
5  | #![warn(clippy::pedantic)]
   |         ^^^^^^^^^^^^^^^^
   = note: `#[warn(clippy::unnecessary_wraps)]` implied by `#[warn(clippy::pedantic)]`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_wraps
help: remove `Result` from the return type...
   |
5  | pub fn detect(context: GenericDetectContext) -> libcnb::DetectOutcome {
   |                                                 ^^^^^^^^^^^^^^^^^^^^^
help: ...and then change returning expressions
   |
10 |         DetectOutcome::Pass(buildplan)
11 |     } else {
12 |         DetectOutcome::Fail
   |
```

It seems like perhaps we should be using Traits here to make the API with `cnb_runtime` more explicit, and so Clippy can tell we're implementing a known interface?

Malax:
This has been addressed with #129.

Issue #68, opened 2021-09-08
Reduce cognitive overhead of `LayerLifecycle`'s large number of generic arguments


edmorley:
This is a review comment split out of https://github.com/Malax/libcnb.rs/pull/28#pullrequestreview-705465664.

If this suggestion ends up being wontfix we can always close this issue out, however I didn't want it to get lost given that the PR was already merged :-)

> * The number of generic arguments for `LayerLifecycle`. There's 5 and I think it's a lot of cognitive overhead there. I wonder if we can use things like associated types to make this easier.
> -- @hone

Malax:
This has been addressed with #129.

PR #129, opened 2021-11-08
Reduce amount of explicit type annotations

Malax:
By introducing a central type for the buildpack itself, we can leverage associated types to define other types that are valid throughout the buildpack. For example, the type for the buildpack metadata never changes. The same is true about the error type and the target platform. The code has been refactored to reference these associated types instead of requiring explicit values everywhere.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,6 @@
 - The `data` module can now be used without the rest of the framework by depending on the `libcnb-data` crate.
 - Introduced `Buildpack` trait that needs to be implemented for each buildpack
 - `cnb_runtime()` now requires a `Buildpack` instead of `detect` and `build` functions.
-- `ErrorHandler` has been removed. Functionality is not part of the new `Buildpack` type.
+- `ErrorHandler` has been removed. Functionality is now part of the new `Buildpack` type.
 
 ## [0.3.0] 2021/09/17
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,6 @@
 - The `data` module can now be used without the rest of the framework by depending on the `libcnb-data` crate.
 - Introduced `Buildpack` trait that needs to be implemented for each buildpack
 - `cnb_runtime()` now requires a `Buildpack` instead of `detect` and `build` functions.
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -10,5 +10,8 @@
 - Allow ProcessType to contain a dot (`.`) character
 - libcnb now targets [Buildpack API 0.6](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.6) <https://github.com/Malax/libcnb.rs/milestone/2>
 - The `data` module can now be used without the rest of the framework by depending on the `libcnb-data` crate.
+- Introduced `Buildpack` trait that needs to be implemented for each buildpack
+- `cnb_runtime()` now requires a `Buildpack` instead of `detect` and `build` functions.
+- `ErrorHandler` has been removed. Functionality is not part of the new `Buildpack` type.
```
edmorley, line 15:
```suggestion
- `ErrorHandler` has been removed. Functionality is now part of the new `Buildpack` type.
```

edmorley (approved):
This is great, thank you!

I guess this fixes #68, #70, and may possibly also fix #86?

CHANGELOG.md
```diff
@@ -10,5 +10,8 @@
 - Allow ProcessType to contain a dot (`.`) character
 - libcnb now targets [Buildpack API 0.6](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.6) <https://github.com/Malax/libcnb.rs/milestone/2>
 - The `data` module can now be used without the rest of the framework by depending on the `libcnb-data` crate.
+- Introduced `Buildpack` trait that needs to be implemented for each buildpack
+- `cnb_runtime()` now requires a `Buildpack` instead of `detect` and `build` functions.
+- `ErrorHandler` has been removed. Functionality is now part of the new `Buildpack` type.
```
edmorley, line 15:
```suggestion
- `ErrorHandler` has been removed. Functionality is now part of the new `Buildpack` trait.
```

Malax:
> I guess this fixes #68, #70, and may possibly also fix #86?

I think so too! :)

PR #134, opened 2021-11-09
Switch from reqwest to ureq in Ruby example

edmorley:
Since we don't need `reqwest`'s additional features (such as async support), and `ureq`'s much smaller dependency tree reduces compile times and binary size substantially.

This helps with CI times for this repository, which cannot use caching since it is a library so doesn't have a lockfile to use for cache-busting.

This is equivalent to heroku/libherokubuildpack#22.


edmorley:
This reduces the CI lint job runtime by half, and the build/test job runtime by a third.

PR #148, opened 2021-11-15
Fix documentation warnings

### LINKED ISSUE #141: `cargo doc` returns warnings
Seems like we should test doc generation in CI and have warnings be fatal there:

```
$ cargo doc
...
warning: unresolved link to `libcnb::Result`
  --> libcnb/src/build.rs:30:23
   |
30 | /// return a failed [`libcnb::Result`] from the build function.
   |                       ^^^^^^^^^^^^^^ no item named `libcnb` in scope
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: `libcnb` (lib doc) generated 1 warning
```

Malax:
Fixes #141

### DIFF
--- a/libcnb/src/build.rs
+++ b/libcnb/src/build.rs
@@ -27,7 +27,7 @@ pub struct BuildContext<B: Buildpack + ?Sized> {
 /// Describes the result of the build phase.
 ///
 /// In contrast to `DetectResult`, it always signals a successful build. To fail the build phase,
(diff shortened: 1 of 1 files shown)


edmorley (approved):
Should we run cargo doc in CI?

edmorley:
Although I can't see a way to make warnings be errors, so 🤷

PR #149, opened 2021-11-15
Add docs for `Buildpack` and related types

Malax:
Follow-up for #129 - adds Rustdoc for newly introduced types, macros and modules.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -25,6 +25,9 @@ use std::fmt::{Debug, Display};
 /// hard links to this single binary.
 ///
 /// Currently symlinks are recommended over hard hard links due to [buildpacks/pack#1286](https://github.com/buildpacks/pack/issues/1286).
+///
+/// Don't implement this directly and use the [`buildpack_main`] macro instead!
+#[doc(hidden)]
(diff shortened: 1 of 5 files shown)


libcnb/src/runtime.rs
```diff
@@ -25,6 +25,9 @@ use std::fmt::{Debug, Display};
 /// hard links to this single binary.
 ///
 /// Currently symlinks are recommended over hard hard links due to [buildpacks/pack#1286](https://github.com/buildpacks/pack/issues/1286).
+///
+/// Don't implement this directly and use the [`buildpack_main`] macro instead!
+#[doc(hidden)]
```
edmorley, line 30: I guess at the moment this is the only documentation that the resultant binary should be symlinked from `bin/{detect,build}`. Though really there needs to be a lot more, eg about cross-compilation too, and it should go in more of an intro/some guide rather than here.

edmorley (approved):
Thank you!

libcnb/src/runtime.rs
Malax: Yeah, 💯  agree. This is something better addressed in book-type documentation, explaining how libcnb works. If we eventually come around unified packaging tooling the issues is not as bad.

PR #157, opened 2021-11-15
Remove PlatformEnv

Malax:
#30 introduced the `Env` type. It was required to have a base environment type where layer environment deltas could be applied to. `PlatformEnv` overlaps with that type almost perfectly. Thus, this PR is removing `PlatformEnv` in favour of `Env`. Besides removing API surface area, this also improves interop between `Platform` and `LayerEnv` since they now use the same type.

### DIFF
--- a/libcnb/src/generic.rs
+++ b/libcnb/src/generic.rs
@@ -2,7 +2,8 @@
 
 use std::path::Path;
 
-use crate::platform::{Platform, PlatformEnv};
+use crate::platform::Platform;
+use crate::{read_platform_env, Env};
 use std::fmt::{Debug, Display, Formatter};
 
 /// Generic TOML metadata.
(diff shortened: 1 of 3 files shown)


libcnb/src/generic.rs
```diff
@@ -22,17 +23,21 @@ impl Display for GenericError {
 
 /// A generic platform that only provides access to environment variables.
 pub struct GenericPlatform {
-    env: PlatformEnv,
+    env: Env,
+}
+
+impl GenericPlatform {
```
edmorley, line 29: This appears unused - presume here just for consistency?

edmorley (approved):

libcnb/src/generic.rs
edmorley: (this was meant to be on the `new()` definition)

Malax: For eventual testing when we want to create GenericPlatform without reading files from disk. :)

PR #162, opened 2021-11-16
Introduce libcnb_newtype macro

Malax:
The macro generates consistent newtypes and their trait implementations across `libcnb-data`. Existing newtypes have been changed to use this macro but are otherwise unmodified and keep their tests and docs.

The macro improves interoperability or our newtypes with `String` and `str` by implementing `Deref`, `Borrow` and `AsRef`. You can now use newtypes from `libcnb-data` interchangeably with `String` and `&str`. Examples:

```rust
let stack_id: StackId = "heroku-20".parse().unwrap();

assert_eq!("HEROKU-20", stack_id.to_uppercase());

fn length(s: &str) -> usize {
    s.len()
}

assert_eq!(9, length(&stack_id));
```

`libcnb-data` newtypes now consistently implement:
- `Debug` 
- `Display`
- `Eq`
- `PartialEq`
- `serde::Deserialize`
- `serde::Serialize`
- `FromStr`
- `Borrow<String>`
- `Deref<Target=String>`
- `AsRef<String>`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,5 +22,6 @@
 - `Generic*` implementations moved to the `generic` module.
 - `LayerContentTypeTable` has been renamed to `LayerTypes`.
 - Remove `PlatformEnv` and replaced it with the already existing `Env`.
+- `StackId`, `ProcessType` and `BuildpackId` now implement `Deref<Target = String>`, `Borrow<String>`, `AsRef<String>` and `Display`.
 
 ## [0.3.0] 2021/09/17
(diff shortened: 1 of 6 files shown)


edmorley (approved):
I like it!

I see the error messages no longer specify the valid criteria, however that seems like an acceptable trade-off to avoid extra complexity, and it's likely best to just have people look at the rustdocs anyway.

Malax:
> I see the error messages no longer specify the validity criteria
I thought about this a bit, it would be possible to pass something to the macro that would end up in the `Display` implementation of the error. I didn't find a great way to encode it that wasn't special syntax.

edmorley:
Yeah particularly for the extra predicate case, and it would be strange to only add the accepted regex to the error message (in the regex failure cases) and not do something similar to the predicate failed cases.

PR #164, opened 2021-11-17
Add util::default_on_not_found

Malax:
Pulled out from #144

### DIFF
--- a/libcnb/src/util.rs
+++ b/libcnb/src/util.rs
@@ -1,0 +1,36 @@
+/// Removes [`std::io::Error`] values from a [`Result`] that have the
+/// [`std::io::ErrorKind::NotFound`] error kind by replacing them with the default value for `T`.
+pub(crate) fn default_on_not_found<T: Default>(
+    result: Result<T, std::io::Error>,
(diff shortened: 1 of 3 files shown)


libcnb/src/util.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::util::default_on_not_found;
+    use std::io::ErrorKind;
+
+    #[test]
+    fn test_with_not_found_kind() {
```
edmorley, line 21: Do the names of the tests need updating now that the function is called `default_on_not_found`?

Also, should there be a test for the `Ok` case? (up to you)

edmorley (approved):

libcnb/src/util.rs
Malax: Oh, the names are new. It's a test with a `NotFound` kind and the other is with "other". I can add something for `Ok`.

edmorley:
@Malax Meant to say I much prefer the new name (over `without_not_found_error` from #144) :-)

PR #165, opened 2021-11-17
Add LayerEnv::chainable_insert

Malax:
Adds `LayerEnv::chainable_insert` that allows for easy inline creation of `LayerEnv` values without a builder. #144 introduced a `LayerEnvBuilder` which will be no longer necessary with this change, keeping documentation about layer environment variable in a central type.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,5 +26,6 @@
 - Add a more general `Default` implementation for `LayerContentMetadata`.
 - Add `PartialEq` implementation for `LayerContentMetadata`.
 - Add `PartialEq` and `Eq` implementations for `LayerTypes`.
+- Add `LayerEnv::chainable_insert`
 
 ## [0.3.0] 2021/09/17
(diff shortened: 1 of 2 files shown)


edmorley (commented):
I notice we already use chainable methods in other place where `self` is mutated, but in those places it's the only method available, rather than having an additional `chainable_` version of the method.

For example in these:
https://github.com/Malax/libcnb.rs/blob/ccdf762ae310ad3d99de7efe421a9e1da3f5e6dd/libcnb/src/detect.rs#L70-L73
https://github.com/Malax/libcnb.rs/blob/ccdf762ae310ad3d99de7efe421a9e1da3f5e6dd/libcnb/src/build.rs#L85-L88
https://github.com/Malax/libcnb.rs/blob/ccdf762ae310ad3d99de7efe421a9e1da3f5e6dd/libcnb/src/build.rs#L90-L93

...they can either be chained, or used without chaining.

Should we be consistent here?

edmorley:
Oh I suppose they are all builders, in which case that is expected.

Malax:
Yet, your concern is valid. We have some builders that are not really advertising that they are builders. Nothing for this PR, but consistency is something we want.

edmorley (approved):

Issue #101, opened 2021-10-12
Prevent the use of special layer names such as "build", "launch" or "store"


edmorley:
In [Buildpack API 0.6](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.6), attempts to create layers with a name of "build", "launch" or "store" are forbidden by the spec, since those layer names conflict with the layer metadata files (eg `build.toml`).

Whilst using those names will result in failures during the build, it would be helpful for libcnb to validate the layer names itself and fail early if those reserved names are used.

This issue is not in the Buildpack API 0.6 milestone, since this is more of a nice-to-have that's not required for 0.6 support.

schneems:
I looked into it and we don't have one abstraction for layer names. Rather every place we're looking for a layer name we're accepting an `AsRef<str>`. 

Thinking aloud a bit, It looks like the options would be:

A) Check `layer_name` everywhere it's present, there's a bunch of usage in `layer_lifecycle.rs` and `build.rs`. For example:
  - https://github.com/Malax/libcnb.rs/blob/05836128f73de9ae67e78a1c68f28165d1d2945c/src/build.rs#L28
  - https://github.com/Malax/libcnb.rs/blob/6a94d587814bce9f2a2f411cd655c2c09063a033/src/layer_lifecycle.rs#L163

Not very DRY, might be prone to logic leakage if new logic is implemented without the checks. A lot of duplicate checking. No build time errors.

B) Be strategic and only check/error on `layer_name` in some locations (but not all)

Less duplicate checking. Not very DRY, might be prone to logic leakage if new logic is implemented without the checks. No build-time errors.

C) Introduce a `LayerName` type with a constructor that forces these constraints.  

The most work, but probably the most verifiably correct thing to do. We could perhaps make this better by implementing automatic conversions from string and &str to that type. Something like this conversion I used in another PR:

```rs
#[derive(Clone)]
pub enum BuildpackDescriptorArg<BM: Clone> {
    File(PathBuf),
    String(String),
    BuildpackToml(BuildpackToml<BM>),
    None,
}

impl<BM: Clone> From<PathBuf> for BuildpackDescriptorArg<BM> {
    fn from(value: PathBuf) -> Self {
        BuildpackDescriptorArg::File(value)
    }
}
```

[...]

PR #172, opened 2021-11-19
Add macros to construct `libcnb-data` newtypes from literal strings

Malax:
## Problem Statement

With the use of newtypes to enforce correct values for CNB defined types also comes a drawback: we no longer cannot use string literals in places where the a literal value is needed. This forces the user to put `.unwrap` calls in these cases:

```rust
let process = Process::new(
    "web".parse().unwrap(),
    "bundle",
    vec!["exec", "ruby", "app.rb"],
    false,
    true,
);
```

Currently, this is not too bad as we only have this case with `ProcessType` really, usually once per buildpack if at all. But as soon as we implement #101, we will have this for every layer. I still think having newtypes for these scenarios is great but we need a solution for literal values and this is where this PR comes in.

## Solution

This PR extends the `libcnb_newtype!` macro introduced in #162 so that it generates another macro for every newtype that allows us to create literal newtype values that are checked at _compile time_! The previous example becomes:

```rust
let process = Process::new(
    process_type!("web"),
    "bundle",
    vec!["exec", "ruby", "app.rb"],
    false,
    true,
);
```

Should the string not be a valid `ProcessType`, the buildpack will not compile:

```
Compiling example-02-ruby-sample v0.1.0 (/Users/manuel.fuchs/projects/libcnb.rs/libcnb/examples/example-02-ruby-sample)
error: "wörker" is not a valid ProcessType value!
  --> libcnb/examples/example-02-ruby-sample/src/main.rs:56:25
   |
56 |                         process_type!("wörker"),
   |                         ^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: this error originates in the macro `process_type` (in Nightly builds, run with -Z macro-backtrace for more info)
```

Since these are regular compile errors, the error nicely propagates into IDEs out of the box:

image

### Regex
The implementation is very generic since we can validate all our newtype values via regular expressions. This allows us to compile the regular expression at compile time and use it against the literal string passed to the macro. As soon as we need to use arbitrary Rust code to validate these, the current approach with automatically generated macros is no longer feasible, unless we write a Rust interpreter or use a scripting language for the validation logic. 

To extend the cases where regular expressions can be used, I replaced the `regex` crate with [fancy_regex](https://github.com/fancy-regex/fancy-regex) to allow for look-around expression which cannot be implemented in a NFA regex implementation. This means our regular expressions are no longer guaranteed to run in linear time. I think this is a fine trade-off for us, since we're only using it with very simple expressions that are entirely controlled by us and only used in a very limited use-case.

See: [Catastrophic Backtracking](https://www.regular-expressions.info/catastrophic.html)

### Removal of extra predicates
See above, we cannot use arbitrary Rust code to do the validation at compile time since we would have no way to execute it. We can implement allow- and blocklists via regex just fine and don't need extra predicates at this point.

### New crate: `libcnb-proc-macros`
This feature can only be implemented via [procedural macros](https://doc.rust-lang.org/reference/procedural-macros.html). By design, procedural macros cannot be used in the crate they're defined it and need a special crate type. Therefore, we have a new crate for the proc macro required by this PR and potential future ones.

### Testing Strategy
Testing code that, by design, should not compile _is hard_. Rust currently provides no way to test non-compiling code in unit tests. This limits the tests to only test the positive cases. I would like to test the negative code path as well. There is [compiletest-rs](https://github.com/Manishearth/compiletest-rs) which seems overkill in our-case. 

To keep things simpler, I'd fine to not test the negative path for now. Note that the string will still be validated at runtime as well, the worst-case scenario would be that code with an illegal newtype value compiles, but fails at runtime as if these macros would not exist. I'm happy to be convinced otherwise. :)


libcnb-proc-macros/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "libcnb-proc-macros"
```
edmorley, line 2: We talked about "libcnb" not being a great name for the higher level concepts recently. If we wanted to switch to another naming strategy, it seems that doing so before we have too many more crates published might be best. Though if we were going to keep libcnb for the lower-level/data structure components, then I presume `libcnb-proc-macros` might still be a our preferred name even after a naming shuffle, so therefore doesn't affect this PR?

libcnb-data/src/buildpack.rs
```diff
@@ -178,6 +178,19 @@ impl Display for BuildpackApi {
...
+    /// Construct a [`BuildpackId`] value at compile time.
+    ///
+    /// Passing a string that is not a valid `BuildpackId` value will yield a compilation error.
+    ///
+    /// # Examples:
+    /// ```
+    /// use libcnb_data::buildpack_id;
+    /// use libcnb_data::buildpack::BuildpackId;
+    ///
+    /// let process_type: BuildpackId = buildpack_id!("heroku/java");
```
edmorley, line 191:
```suggestion
    /// let buildpack_id: BuildpackId = buildpack_id!("heroku/java");
```

libcnb-data/src/buildpack.rs
```diff
@@ -196,11 +209,23 @@ libcnb_newtype!(
...
+    /// Construct a [`StackId`] value at compile time.
+    ///
+    /// Passing a string that is not a valid `StackId` value will yield a compilation error.
+    ///
+    /// # Examples:
+    /// ```
+    /// use libcnb_data::stack_id;
+    /// use libcnb_data::buildpack::StackId;
+    ///
+    /// let process_type: StackId = stack_id!("heroku-20");
```
edmorley, line 226:
```suggestion
    /// let stack_id: StackId = stack_id!("heroku-20");
```

libcnb-data/src/newtypes.rs
```diff
@@ -1,4 +1,8 @@
-/// Macro to generate newtypes backed by `String`
+// The compiler has a hard time figuring out that we actually use libcnb_sys in here. To avoid
+// false-positive warnings about unused crates, we import it as _ here:
+use libcnb_proc_macros as _;
```
edmorley, line 3: This doesn't appear to be needed any more. Perhaps the import in `libcnb-data/src/internals.rs` meant it's no longer needed?

libcnb-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,67 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn::parse::{Parse, ParseStream};
+use syn::parse_macro_input;
+use syn::Token;
+
```
edmorley, line 6: Could you enable the additional lints enabled in the other crates?
https://github.com/Malax/libcnb.rs/blob/a66f1f90f458867ccc1b59ff3b3cceeea80c1918/libcnb-data/src/lib.rs#L3-L7

(I'm looking forward to rust-lang/rust-clippy/issues/6625 et al)

libcnb-data/src/newtypes.rs
```diff
@@ -12,43 +16,45 @@
...
+///     // from a user's buildpack crate.
+///     tests::doctest
+///     /// RustDoc for the macro (optional)
+///     buildpack_id,
 ///     /// RustDoc for the newtype itself (optional)
 ///     BuildpackId,
 ///     /// RustDoc for the newtype error (optional)
 ///     BuildpackIdError,
-///     // The regular expression that must match for the String to be valid
+///     // The regular expression that must match for the String to be valid. Uses the fancy_regex
```
edmorley, line 37:
```suggestion
///     // The regular expression that must match for the String to be valid. Uses the `fancy_regex`
```

edmorley (approved):

edmorley:
Thank you for doing this! I much prefer the idea of compile time validation :-)

libcnb-proc-macros/Cargo.toml
Malax: Yeah, I think it's fine for this PR.

PR #174, opened 2021-11-19
Add LayerName newtype

### LINKED ISSUE #101: Prevent the use of special layer names such as "build", "launch" or "store"

Malax:
Fixes #101 

Note: A follow-up PR will change the RustDoc for other newtypes to match the docs for `layer_name` and `LayerName` to achieve consistency.

### DIFF
--- a/libcnb-data/src/newtypes.rs
+++ b/libcnb-data/src/newtypes.rs
@@ -52,7 +52,7 @@ macro_rules! libcnb_newtype {
         $error_name:ident,
         $regex:expr
     ) => {
-        #[derive(Debug, Eq, PartialEq, ::serde::Deserialize, ::serde::Serialize)]
+        #[derive(Debug, Eq, PartialEq, ::serde::Deserialize, ::serde::Serialize, Clone)]
         $(#[$type_attributes])*
         pub struct $name(String);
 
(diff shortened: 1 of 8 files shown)


libcnb-data/src/newtypes.rs
```diff
@@ -52,7 +52,7 @@ macro_rules! libcnb_newtype {
         $error_name:ident,
         $regex:expr
     ) => {
-        #[derive(Debug, Eq, PartialEq, ::serde::Deserialize, ::serde::Serialize)]
+        #[derive(Debug, Eq, PartialEq, ::serde::Deserialize, ::serde::Serialize, Clone)]
```
edmorley, line 55: The list of traits in this macro's rustdoc needs updating to add `Clone`

edmorley (approved):

libcnb-data/src/newtypes.rs
Malax: Your attention to detail is top notch Ed. Thanks! :)

PR #175, opened 2021-11-22
Unify newtype docs

Malax:
The documentation pieces for our newtypes are a bit fragmented and this PR unifies them.

### DIFF
--- a/libcnb-data/src/buildpack.rs
+++ b/libcnb-data/src/buildpack.rs
@@ -191,20 +191,28 @@ libcnb_newtype!(
     /// let buildpack_id: BuildpackId = buildpack_id!("heroku/java");
     /// ```
     buildpack_id,
-    /// buildpack.toml Buildpack Id. This is a newtype wrapper around a String.
-    /// It MUST only contain numbers, letters, and the characters ., /, and -.
-    /// It also cannot be `config` or `app`.
-    /// Use [`std::str::FromStr`] to create a new instance of this struct.
+    /// The ID of a buildpack.
+    ///
+    /// It MUST only contain numbers, letters, and the characters `.`, `/`, and `-`.
+    /// It also MUST NOT be `config` or `app`.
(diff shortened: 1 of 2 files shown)


libcnb-data/src/buildpack.rs
```diff
@@ -226,25 +234,28 @@ libcnb_newtype!(
     /// let stack_id: StackId = stack_id!("heroku-20");
     /// ```
     stack_id,
-    /// buildpack.toml Stack Id. This is a newtype wrapper around a String.
-    /// It MUST only contain numbers, letters, and the characters ., /, and -.
-    /// or be `*`.
+    /// The ID of a stack.
+    ///
+    /// It MUST only contain numbers, letters, and the characters `.`, `/`, and `-`.
```
edmorley, line 239: This should technically have `*`, but might as well omit for now given it will be true shortly :-)

libcnb-data/src/buildpack.rs
```diff
@@ -226,25 +234,28 @@ libcnb_newtype!(
...
-    /// Use [`std::str::FromStr`] to create a new instance of this struct.
+    /// Use the [`stack_id`](crate::buildpack_id) macro to construct a `StackId` from a
+    /// literal string. To parse a dynamic string into a `StackId`, use
+    /// [`str::parse`](str::parse).
     ///
     /// # Examples
     /// ```
-    /// use std::str::FromStr;
-    /// use libcnb_data::buildpack::StackId;
+    /// use libcnb_data::buildpack::BuildpackId;
```
edmorley, line 247: This and the below references buildpack ID rather than stack ID

edmorley (approved):

libcnb-data/src/buildpack.rs
Malax: I'm sorry - I don't know how this slipped by. 🙄

PR #178, opened 2021-11-22
Remove unused defaults module

Malax:


edmorley (approved):
I take it this is unused as of #159.

I'd interpreted the `#[serde(default = "defaults::r#false")]` as either some custom Serde thing, or from the stdlib. I hadn't realised this `defaults` module existed to provide a bool?! 😆

Malax:
I think this was done to be more explicit about the actual default value. I don't think it was a bad idea, but also not really necesary.

PR #184, opened 2021-11-23
Remove `Display` trait bound for `Buildpack::Error`

Malax:
This bound was only necessary for the default implementation of `Buildpack::handle_error`. However, we can provide a useful implementation without it by just using `Debug` for buildpack specific errors. This makes it easier to start working on a buildpack as `Debug` can be derived (`Display` cannot).

In practice, buildpack authors should always implement `Buildpack::handle_error` where they can decide to rely on `Display`, independently from the framework.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -14,7 +14,7 @@ use crate::error::Error;
 use crate::platform::Platform;
 use crate::toml_file::{read_toml_file, write_toml_file};
 use crate::{Result, LIBCNB_SUPPORTED_BUILDPACK_API};
-use std::fmt::{Debug, Display};
+use std::fmt::Debug;
 
 /// Main entry point for this framework.
 ///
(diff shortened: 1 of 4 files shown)


libcnb/src/runtime.rs
```diff
@@ -200,13 +200,13 @@ fn parse_build_args_or_exit() -> BuildArgs {
     }
 }
 
-fn read_buildpack_dir<E: Display + Debug>() -> Result<PathBuf, E> {
+fn read_buildpack_dir<E: Debug>() -> crate::Result<PathBuf, E> {
```
edmorley, line 203: I presume the `s/Result/crate::Result/` is to prevent confusion with the stdlib `Result`. Should the `use crate::Result;` in this file be removed to force all usages to specify `crate::Result`?

edmorley (approved):

libcnb/src/runtime.rs
Malax: Yeah, that's the intention. I like your idea and think we should apply this to the whole codebase! I filed #185 for it and will merge this as-is for now to keep this PR focussed on the actual change.

PR #186, opened 2021-11-23
Adjust `Stack` and `StackId` to better model the "any" stack

### LINKED ISSUE #147: Incorrect data model for StackId
https://github.com/Malax/libcnb.rs/commit/034b2a2f4be62f08ced2ed34a4c3a3ab06b521c4 changed `StackId` so that `*` is a valid `StackId`, which is incorrect.

There can't be a stack named `*`, but you can specify `*` in `buildpack.toml` to indicate that the buildpack is compatible with any stack. Those are very different things.

We should revert `StackId` to its previous implementation and introduce a new type that is used in the buildpack descriptor to model the "compatible with any stack" case. An enum like this comes to mind:

```rust
enum StackCompatibility {
    Any,
    Specific<Vec<StackId>>
}
```

GUS-W-10204027.

edmorley:
Previously the "any" stack was modelled as a `Stack` with a `StackId` of `"*"`, and an empty `Vec` of mixins, which does not accurately model the spec:
https://github.com/buildpacks/spec/blob/buildpack/0.6/platform.md#stack-id
https://github.com/buildpacks/spec/blob/buildpack/0.6/buildpack.md#buildpack-implementations

Now:
- `Stack` is an enum with `Any` and `Specific` variants.
-  `StackId` no longer permits IDs of `*`, since `Stack::Any` should be used instead.
- `BuildpackTomlError::InvalidStarStack` has been replaced by `BuildpackTomlError::InvalidAnyStack`.

Fixes #147.
GUS-W-10204027.

### DIFF
--- a/libcnb-data/src/buildpack.rs
+++ b/libcnb-data/src/buildpack.rs
@@ -74,22 +74,21 @@ pub struct License {
     pub uri: Option<String>,
 }
 
-#[derive(Deserialize, Debug)]
-#[serde(try_from = "StackUnchecked")]
-pub struct Stack {
-    pub id: StackId,
-    pub mixins: Vec<String>,
-}
-
 // Used as a "shadow" struct to store
 // potentially invalid `Stack` data when deserializing
 // https://dev.to/equalma/validate-fields-and-types-in-serde-with-tryfrom-c2n
 #[derive(Deserialize)]
 struct StackUnchecked {
-    pub id: StackId,
-
+    pub id: String,
     #[serde(default)]
     pub mixins: Vec<String>,
+}
+
+#[derive(Deserialize, Debug, Eq, PartialEq)]
+#[serde(try_from = "StackUnchecked")]
+pub enum Stack {
+    Any,
+    Specific { id: StackId, mixins: Vec<String> },
 }
 
 impl TryFrom<StackUnchecked> for Stack {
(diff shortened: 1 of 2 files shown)


libcnb-data/src/buildpack.rs
```diff
@@ -282,6 +288,8 @@ pub enum BuildpackTomlError {
 mod tests {
     use super::*;
 
+    type GenericBuildpackToml = BuildpackToml<Option<toml::value::Table>>;
```
edmorley, line 291: I'll update the existing tests in this file to use this new type (and other test cleanup) after this merges.

libcnb-data/src/buildpack.rs
```diff
@@ -260,16 +266,16 @@ libcnb_newtype!(
...
 
 #[derive(thiserror::Error, Debug)]
 pub enum BuildpackTomlError {
     #[error("Found `{0}` but value MUST be in the form `<major>.<minor>` or `<major>` and only contain numbers.")]
     InvalidBuildpackApi(String),
 
-    #[error("Stack with id `*` MUST not contain mixins. mixins: [{0}]")]
-    InvalidStarStack(String),
+    #[error("Stack with id `*` MUST NOT contain mixins, however the following mixins were specified: {0}")]
+    InvalidAnyStack(String),
```
Malax, line 278: Ideally, this would not be a `String` but a `Vec<String>` (or `Vec<Mixin>` when we come to it). I don't know from the top of my head if `thiserror` can deal with this out of the box or if would have to provide a `Display` implementation for it.

This isn't critical, but worth looking into if it's easily fixable.

Malax (approved):
Well done! 👍🏻  I especially like the tests, they're neatly done!

libcnb-data/src/buildpack.rs
edmorley: It turns out thiserror does allow doing this :-)
ba185605446ffb82d923b6e136e3540e2bb6335d

Malax: Awesome! Thanks for looking into it!

PR #187, opened 2021-11-23
Fix `LayerName` validation

### LINKED ISSUE #180: `layer_name!` macro permits empty string layer names
It seems the regex here needs to use `.+` not `.*`:
https://github.com/Malax/libcnb.rs/blob/2f57cd18beff95a52fd85229823e0c8dc05892ce/libcnb-data/src/layer.rs#L42

...otherwise it permits the empty string.

edmorley:
Previously the validation:
- permitted empty string layer names, which should be rejected.
- rejected layer names containing {build,launch,store} as a substring, when those are allowed so long as it's
  only a partial substring.

Tests have now been added.

I didn't add a changelog entry, since this is a bug fix to a change that's already documented in the changelog and not yet released.

Fixes #180.
GUS-W-10219476.

### DIFF
--- a/libcnb-data/src/layer.rs
+++ b/libcnb-data/src/layer.rs
@@ -39,5 +39,34 @@ libcnb_newtype!(
     /// ```
     LayerName,
     LayerNameError,
-    r"^(?!build|launch|store).*$"
+    r"^(?!(build|launch|store)$).+$"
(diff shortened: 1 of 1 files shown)


Malax (approved):
Thank you!

PR #190, opened 2021-11-23
Refactor ` libcnb-data::buildpack`

edmorley:
Since at 500 lines (even before we finish adding tests), it was getting a bit unwieldy.

The public API has not changed, aside from `BuildpackTomlError` having been split into `BuildpackApiError` and `StackError`.

Future PRs will add additional unit tests to the split out types, and clean up the TOML integration tests in the root of the `buildpack` module.

GUS-W-10222615.


libcnb-data/src/buildpack/mod.rs
```diff
@@ -0,0 +1,288 @@
+mod api;
+mod id;
```
edmorley, line 2: I couldn't decide whether to use `api` / `id` or `buildpack_api` , / `buildpack_id` as the internal module name, given the parent module is `buildpack` so the latter seems duplicative.

libcnb-data/src/buildpack.rs
```diff
@@ -1,520 +0,0 @@
...
-    InvalidBuildpackApi(String),
-
-    #[error("Stack with id `*` MUST NOT contain mixins, however the following mixins were specified: `{}`", .0.join("`, `"))]
-    InvalidAnyStack(Vec<String>),
-
-    #[error("Invalid Stack ID: {0}")]
-    InvalidStackId(#[from] StackIdError),
-
-    #[error("Invalid Buildpack ID: {0}")]
-    InvalidBuildpackId(#[from] BuildpackIdError),
```
edmorley, line 283: This `InvalidBuildpackId` variant was unused, so didn't get migrated anywhere, instead removed.

libcnb-data/src/buildpack/api.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[error("Found `{0}` but value MUST be in the form `<major>.<minor>` or `<major>` and only contain numbers.")]
+    InvalidBuildpackApi(String),
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn buildpack_api_from_str_major_minor() {
```
edmorley, line 79: The name of this and the other test in this file were changed in addition to moving the tests here, since they were previously incorrectly named `buildpack_id_*`.

Malax (approved):

PR #195, opened 2021-11-24
Fix newtype `Deserialize` implementation

### LINKED ISSUE #189: newtype validation not working during Serde deserialization
If a type is created using one of the libcnb newtype macros, validation works as expected when using `.parse::<Type>()` or `Type::from_str()`.

However this validation is not triggered when Serde deserializes that type.

For example the second test here should pass (as an error is expected), however it does not.

```rust
use serde::Deserialize;

use crate::newtypes::libcnb_newtype;

libcnb_newtype!(testcase, foo_id, FooId, FooIdError, r"^a+$");

#[derive(Deserialize, Debug)]
struct MyToml {
    foo: FooId,
}

#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn works_as_expected() {
        assert!("aaa".parse::<FooId>().is_ok());
        assert_eq!(
            "bbb".parse::<FooId>(),
            Err(FooIdError::InvalidValue(String::from("bbb")))
        );
    }

    #[test]
    fn should_pass_but_does_not() {
        let raw = r#"
foo = "bbb"
"#;

        let err = toml::from_str::<MyToml>(raw).unwrap_err();
        assert!(err
            .to_string()
            .contains(&FooIdError::InvalidValue(String::from("bbb")).to_string()));
    }
}
```

Instead the output is:

```
running 2 tests
test testcase::tests::should_pass_but_does_not ... FAILED
test testcase::tests::works_as_expected ... ok

failures:

---- testcase::tests::should_pass_but_does_not stdout ----
thread 'testcase::tests::should_pass_but_does_not' panicked at 'called `Result::unwrap_err()` on an `Ok` value: MyToml { foo: FooId("bbb") }', libcnb-data/src/testcase.rs:31:49


failures:
    testcase::tests::should_pass_but_does_not

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 23 filtered out; finished in 0.00s
```

GUS-W-10222486.

Malax:
The derived `Deserialize` implementation does not use the validation logic from `FromStr`. By implementing a custom `Deserialize`, we can ensure it does.

Skips changelog as this is a fix for unreleased code.

Fixes #189

### DIFF
--- a/libcnb-data/src/newtypes.rs
+++ b/libcnb-data/src/newtypes.rs
@@ -57,14 +57,6 @@ macro_rules! libcnb_newtype {
         $(#[$type_attributes])*
         pub struct $name(String);
 
-        impl<'de> ::serde::Deserialize<'de> for $name {
-            fn deserialize<D: ::serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
-                String::deserialize(d)?
-                    .parse::<$name>()
-                    .map_err(::serde::de::Error::custom)
-            }
-        }
-
         #[derive(::thiserror::Error, Debug, Eq, PartialEq)]
         $(#[$error_type_attributes])*
         pub enum $error_name {
@@ -94,6 +86,14 @@ macro_rules! libcnb_newtype {
                 } else {
                     Err($error_name::InvalidValue(String::from(value)))
                 }
+            }
+        }
+
+        impl<'de> ::serde::Deserialize<'de> for $name {
+            fn deserialize<D: ::serde::Deserializer<'de>>(d: D) -> Result<Self, D::Error> {
+                String::deserialize(d)?
+                    .parse::<$name>()
+                    .map_err(::serde::de::Error::custom)
             }
         }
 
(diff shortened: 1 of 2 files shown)


libcnb-data/src/newtypes.rs
```diff
@@ -53,10 +53,18 @@ macro_rules! libcnb_newtype {
         $error_name:ident,
         $regex:expr
     ) => {
-        #[derive(Debug, Eq, PartialEq, ::serde::Deserialize, ::serde::Serialize, Clone)]
+        #[derive(Debug, Eq, PartialEq, ::serde::Serialize, Clone)]
         $(#[$type_attributes])*
         pub struct $name(String);
 
+        impl<'de> ::serde::Deserialize<'de> for $name {
```
edmorley, line 60: The implementations for `$name` are interleaved between those for `$error_name`. Is this intentional?

edmorley (approved):
Thank you for fixing this! :-)

libcnb-data/src/newtypes.rs
Malax: Nope. :) I'll move the `Deserialize` implementation.

PR #196, opened 2021-11-24
Add more tests for the libcnb_newtype generated types

edmorley:
Adds tests for the `libcnb_newtype` generated types - specifically that the validation regex is working as expected.

Prompted by #187.

GUS-W-10222615.

### DIFF
--- a/libcnb-data/src/buildpack/id.rs
+++ b/libcnb-data/src/buildpack/id.rs
@@ -40,23 +40,46 @@ libcnb_newtype!(
     /// ```
     BuildpackId,
     BuildpackIdError,
-    r"^(?!app$|config$)[[:alnum:]./-]+$"
+    r"^(?!(app|config)$)[[:alnum:]./-]+$"
 );
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::str::FromStr;
 
     #[test]
-    fn buildpack_id_does_not_allow_app() {
-        let result = BuildpackId::from_str("app");
-        assert!(result.is_err());
+    fn buildpack_id_validation_valid() {
+        assert!("heroku/jvm".parse::<BuildpackId>().is_ok());
+        assert!("Abc123./-".parse::<BuildpackId>().is_ok());
+        assert!("app-foo".parse::<BuildpackId>().is_ok());
+        assert!("foo-app".parse::<BuildpackId>().is_ok());
     }
 
     #[test]
-    fn buildpack_id_does_not_allow_config() {
-        let result = BuildpackId::from_str("config");
-        assert!(result.is_err());
+    fn buildpack_id_validation_invalid() {
+        assert_eq!(
+            "heroku_jvm".parse::<BuildpackId>(),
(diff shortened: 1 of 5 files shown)


libcnb-data/src/buildpack/id.rs
```diff
@@ -40,23 +40,46 @@ libcnb_newtype!(
     /// ```
     BuildpackId,
     BuildpackIdError,
-    r"^(?!app$|config$)[[:alnum:]./-]+$"
+    r"^(?!(app|config)$)[[:alnum:]./-]+$"
```
edmorley, line 43: This change is a no-op, and was to make the regex consistent with that for `LayerName`, and reduce chance of bugs, if a new entry is added here but the closing `$` omitted.

libcnb-data/src/launch.rs
```diff
@@ -128,32 +128,36 @@ libcnb_newtype!(
     /// ```
     ProcessType,
     ProcessTypeError,
-    r"^[[:alnum:]\._-]+$"
+    r"^[[:alnum:]._-]+$"
```
edmorley, line 131: This change is a no-op - the escaping is unnecessary, as confirmed by the new tests. The new regex style is now consistent with that used by the other types.

libcnb-data/src/launch.rs
```diff
@@ -128,32 +128,36 @@ libcnb_newtype!(
...
 );
 
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::str::FromStr;
 
     #[test]
-    fn test_process_type_eq() {
```
edmorley, line 140: This type equality test has been moved to `libcnb-data/src/newtypes.rs`.

libcnb-data/src/launch.rs
Malax: I was doubtful when I read this comment and checked. You're right, it does not change the validation because of the context where the `.` is used in. However, what do we gain by removing it? I argue that having the escape there makes is more obvious that the regex is correct and does not match any character at this location.

Definitely not a blocker for me, we can move forward as-is, but maybe the above is worth considering.

libcnb-data/src/newtypes.rs
```diff
@@ -167,6 +167,18 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_type_eq() {
```
Malax, line 171: I ❤️  how you kept the test values consistent with the other tests. 😬

Malax (approved):
This is great, thank you! 👍🏻

libcnb-data/src/launch.rs
edmorley: @Malax So I first wondered if the `\` might be taken literally inside the `[]` and actually had to test to confirm it didn't. The other usages don't escape the `.`.

My main preference is to (a) be consistent (so if we want to use this, we should change the others), (b) have tests so we don't have to think about this (which we do, for the `.` case).

Malax: Again, not a blocker for me :) This is really just about the case if someone reads the regex and wonders. Super minor. Happy to move forward! :)

PR #198, opened 2021-11-24
fix: set proper file ending for delimiter: .delim

lead4good:


edmorley:
Yeah this should be`.delim`:
https://github.com/buildpacks/spec/blob/main/buildpack.md#environment-variable-modification-rules

And `delim` is actually used elsewhere already, eg:
https://github.com/Malax/libcnb.rs/blob/344eabf8ccc36018f3fa65f4000612c0983eab9e/libcnb/src/layer_env.rs#L528

...which is why these tests pass:
https://github.com/Malax/libcnb.rs/blob/344eabf8ccc36018f3fa65f4000612c0983eab9e/libcnb/src/layer_env.rs#L617-L618

@Malax I wonder if we should try and have a single mapping of file extensions <-> `ModificationBehavior` variants, to avoid them from getting out of sync?

edmorley:
And thank you for the PR!

Malax:
@lead4good Thanks for the PR!

@edmorley:
> I wonder if we should try and have a single mapping of file extensions

We could do that. But looking at the code, we only have one mapping for writing and one for reading - it's not like we require that mapping all over the place. 

While certainly an improvement, I'm not sure if it's worth time investment (since we can't use the literal matches anymore for example). We should 💯  add a test for this though!

Malax:
I opened an issue for the missing test here: #206

Malax (approved):

PR #199, opened 2021-11-24
Add Cargo command for cross-compiling and packaging buildpacks

Malax:
This PR adds a Cargo command binary crate to build and package `libcnb.rs` based buildpacks.

Previously, every buildpack author had to come up with a way to package their buildpack that conformed both to the CNB spec as well as the assumptions `libcnb.rs` makes about how it's packaged. Some used [cargo-make](https://github.com/sagiegurari/cargo-make), others bash scripts. Writing a proper build script and setting up the project is hard since it's common that a buildpack must be cross-compiled to another platform. Our own Ruby example shipped with a `Makefile.toml` for `cargo-make`, but did not document how to use it. All of the above made getting started with `libcnb.rs` very hard, especially since there were no docs around this either.

With this new Cargo command, `libcnb.rs` users can use a centrally managed packaging tool for their buildpacks and can fully concentrate on writing their buildpack instead of build logic. It tries to make cross-compiling easier by providing useful hints.

The README [has been updated](https://github.com/Malax/libcnb.rs/blob/malax/cargo/README.md) to transform the existing "Hello World Buildpack" section into an end-to-end quickstart guide that includes packaging and running the buildpack.

## Installation

Users will install it via `cargo install libcnb-cargo`, to try it out before the release on crates.io run:

```bash
$ cargo install --path /path/to/libcnb.rs/libcnb-cargo
```

## Usage

```
USAGE:
    cargo libcnb package [FLAGS] [OPTIONS]

FLAGS:
    -h, --help       Prints help information
        --release    Build in release mode, with optimizations
    -V, --version    Prints version information

OPTIONS:
    -o, --output <output-path>    Write buildpack tarball to this path instead of Cargo's target directory
        --target <target>         Build for the target triple [default: x86_64-unknown-linux-musl]
```

## Example Output

```shell
$ cargo libcnb package
INFO - Reading buildpack metadata...
INFO - Found valid buildpack with id "libcnb-examples/my-buildpack" @ 0.1.0!
INFO - Building buildpack binary (x86_64-unknown-linux-musl)...
   Compiling syn v1.0.82
   Compiling bit-vec v0.6.3
   Compiling regex-syntax v0.6.25
   # Omitting further compilation output...
    Finished dev [unoptimized + debuginfo] target(s) in 19.71s
INFO - Writing buildpack tarball...
INFO - Successfully wrote buildpack tarball target/libcnb-examples_my-buildpack_0.1.0_dev.tar.gz (6.5M)
INFO - Packaging successfully finished!
INFO - Hint: To test your buildpack locally with pack, run: pack build my-image -b target/libcnb-examples_my-buildpack_0.1.0_dev.tar.gz --path /path/to/application
```

## What's next
I consider this an MVP, there are many improvements that could be made and it could have more features. We'll implement these features and fixes along the way.

This new `libcnb-cargo` crate is, besides a binary crate, also a library crate that exposes the building blocks for the packaging process. The intent is to use these building blocks in an integration test solution, ensuring parity between command line packaging and integration testing.

GUS-W-10229113.


README.md
```diff
@@ -13,79 +13,228 @@
...
+[[stacks]]
+id = "heroku-20"
+```
 
-A basic hello world buildpack looks like this:
+That's all we need! We can now move on to finally write some buildpack code!
+
+
+### Writing the Buildpack
+As aforementioned, the buildpack we're writing will be very simple. We will just log a "Hello World" image during build
```
edmorley, line 92:
```suggestion
As aforementioned, the buildpack we're writing will be very simple. We will just log a "Hello World" message during the build
```

README.md
```diff
@@ -13,79 +13,228 @@
...
     type Metadata = GenericMetadata;
 
-    // The error type for this buildpack. Buildpack authors usually implement an enum with 
-    // specific errors that can happen during buildpack execution. This error type should 
-    // only contain error specific to this buildpack, such as CouldNotExecuteMaven or 
-    // InvalidGemfileLock. This example buildpack uses GenericError which means this buildpack 
+    // The error type for this buildpack. Buildpack authors usually implement an enum with
+    // specific errors that can happen during buildpack execution. This error type should
+    // only contain error specific to this buildpack, such as CouldNotExecuteMaven or
+    // InvalidGemfileLock. This example buildpack uses GenericError which means this buildpack
```
edmorley, line 123:
```suggestion
    // only contain error specific to this buildpack, such as `CouldNotExecuteMaven` or
    // `InvalidGemfileLock`. This example buildpack uses `GenericError` which means this buildpack
```

README.md
```diff
@@ -13,79 +13,228 @@
...
+    // The CNB platform this buildpack targets, usually GenericPlatform. See the CNB spec for
+    // more information about platforms:
     // https://github.com/buildpacks/spec/blob/main/buildpack.md
     type Platform = GenericPlatform;
 
-    // The type for the metadata of the buildpack itself. This is the data found in the 
-    // `[metadata]` section of your buildpack's buildpack.toml. The framework will automatically 
-    // try to parse it into the specified type. This example buildpack uses GenericMetadata which 
+    // The type for the metadata of the buildpack itself. This is the data found in the
+    // `[metadata]` section of your buildpack's buildpack.toml. The framework will automatically
```
edmorley, line 115:
```suggestion
    // `[metadata]` section of your buildpack's `buildpack.toml`. The framework will automatically
```

README.md
```diff
@@ -13,79 +13,228 @@
...
 use libcnb::detect::{DetectContext, DetectResult, DetectResultBuilder};
 use libcnb::generic::{GenericError, GenericMetadata, GenericPlatform};
 use libcnb::{buildpack_main, Buildpack};
 
 struct HelloWorldBuildpack;
 
 impl Buildpack for HelloWorldBuildpack {
-    // The CNB platform this buildpack targets, usually GenericPlatform. See the CNB spec for 
-    // more information about platforms: 
+    // The CNB platform this buildpack targets, usually GenericPlatform. See the CNB spec for
```
edmorley, line 109:
```suggestion
    // The CNB platform this buildpack targets, usually `GenericPlatform`. See the CNB spec for
```

README.md
```diff
@@ -13,79 +13,228 @@
...
     //
-    // More generic errors that happen during buildpack execution such as I/O errors while 
+    // More generic errors that happen during buildpack execution such as I/O errors while
     // writing CNB TOML files are handled by libcnb.rs itself.
     type Error = GenericError;
 
-    // This method will be called when the CNB lifecycle calls detect. Use the DetectContext to 
-    // access CNB data such as the stack this buildpack is currently executed on, the app 
-    // directory and similar things. When using libcnb.rs, you never have to read environment 
+    // This method will be called when the CNB lifecycle calls detect. Use the DetectContext to
```
edmorley, line 130:
```suggestion
    // This method will be called when the CNB lifecycle calls detect. Use the `DetectContext` to
```

README.md
```diff
@@ -13,79 +13,228 @@
...
+    // access CNB data such as the stack this buildpack is currently executed on, the app
+    // directory and similar things. When using libcnb.rs, you never have to read environment
     // variables or read/write files to disk to interact with the CNB lifecycle.
     //
-    // One example of this is the return type of this method. DetectResult encapsulates the 
-    // required exit code as well as the data written to the build plan. libcnb.rs will, 
-    // according to the returned value, handle both writing the build plan and exiting with 
+    // One example of this is the return type of this method. DetectResult encapsulates the
```
edmorley, line 135:
```suggestion
    // One example of this is the return type of this method. `DetectResult` encapsulates the
```

README.md
```diff
@@ -13,79 +13,228 @@
...
+                false,
+                true,
+            )))
+            .build())
     }
 }
 
 // Implements the main function and wires up the framework for the given buildpack.
 buildpack_main!(HelloWorldBuildpack);
+
```
edmorley, line 163: (Means when using the "copy" button in the rendered markdown, the resultant text doesn't have two trailing empty lines)

```suggestion
```

README.md
```diff
@@ -13,79 +13,228 @@
...
 It currently targets version `0.6` of the CNB spec.
 
+## Quick Start Guide
+This quick start guide walks you through writing a simple Cloud Native Buildpack in Rust with libcnb.rs. It will also 
+demonstrate how to package your buildpack and how to build an application image with it.
+
+### Development Environment Setup
+In addition to the libcnb crate, we need some tools to compile, package and run the buildpack. These steps only need to
+be carried out once and don't need to be repeated for each buildpack you will write. 
+#### libcnb Cargo Command
```
edmorley, line 23:
```suggestion

#### libcnb Cargo Command
```

README.md
```diff
@@ -13,79 +13,228 @@
...
+   # Omitting further compilation output...
+    Finished dev [unoptimized + debuginfo] target(s) in 19.71s
+INFO - Writing buildpack tarball...
+INFO - Successfully wrote buildpack tarball target/libcnb-examples_my-buildpack_0.1.0_dev.tar.gz (6.5M)
+INFO - Packaging successfully finished!
+INFO - Hint: To test your buildpack locally with pack, run: pack build my-image -b target/libcnb-examples_my-buildpack_0.1.0_dev.tar.gz --path /path/to/application
 ```
+
+If you get errors with hints about how to install required tools to cross-compile from your host platform to the
+target platform, follow them and try again. These differ from platform to platform, that's why they're not part of this
```
edmorley, line 189:
```suggestion
target platform, follow them and try again. These differ from platform to platform, which is why they're not part of this
```

README.md
```diff
@@ -13,79 +13,228 @@
...
+quick start guide.
+
+### Running the Buildpack
+You might have seen in the hint in the output of the packaging step that we're now ready to run our buildpack locally 
+with `pack`. Before we can do this, we need an application to build. Since our buildpack does not interact with the
+application code at all, we just create an empty directory and use that as our application:
+
+```shell
+$ mkdir bogus-app
+$ pack build my-image -b target/libcnb-examples_my-buildpack_0.1.0_dev.tar.gz --path bogus-app
```
edmorley, line 199: Will this fail if the user's pack config still has the initial default builder? (Which I think might be the samples builder, I can't remember)

If so, we either need to pass `--builder heroku/buildpacks:20` here, or have making it the default (https://buildpacks.io/docs/tools/pack/cli/pack_config_default-builder/) be part of the initial setup section earlier.

README.md
```diff
@@ -13,79 +13,228 @@
...
+Setting default process type 'web'
+Saving my-image...
+*** Images (d4f67a828236):
+      my-image
+Successfully built image my-image
+```
+
+### Running the image
+Just run the crated Docker image as you normally would. If all went well, you should see our "Hello World!" message in
+your terminal:
```
edmorley, line 230: "Just" can sometimes read a bit negative ("oh you just have to do X, it's not that hard"). The usage here isn't too bad, but perhaps something like this would be gentler?

```suggestion
The newly created Docker image can be run in the same way as you would a Docker image created via `docker build`.
If all went well, you should see our "Hello World!" message in your terminal:
```

README.md
```diff
@@ -13,79 +13,228 @@
...
+
+```shell
+$ docker run my-image
+Hello World!
+```
+
+### Next Steps
+While the buildpack we've written in this quick start guide is not very useful, it can serve as a starting point for a 
+more useful buildpack. To discover more of the libcnb API, browse the [examples directory](example) and the 
+[documentation on docs.rs](docs.rs).
```
edmorley, line 240: The link is currently broken, I think this will fix it, but haven't checked.

```suggestion
[documentation on docs.rs][docs.rs].
```

libcnb-cargo/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "libcnb-cargo"
+version = "0.1.0"
+edition = "2021"
```
edmorley, line 4: We need to add a number of other fields here - not all of the ones in the other crates, but a fair few - eg `license`, `include` (so the license gets bundled), `repository` etc.

Though these fields are also missing from `libcnb-proc-macros`, so happy to defer to a future PR where we standardise on the fields to use across all crates. Also we should decide what to do for README for the ancillary crates. Perhaps they should not use the repo root README (and only `libcnb` use that), and instead have a shorter readme that links to `libcnb`.

edmorley (commented):
Thank you so much for working on this! Also love the new guide in README.

Here's an interim review (mainly on docs) whilst I work on the rest :-)

libcnb-cargo/Cargo.toml
Malax: I'll add them in this PR for the new crate and create an issue for `libcnb-proc-macros`: #208

README.md
Malax: This should be a direct link to the libcnb.rs docs. I'll fix it! :)

README.md
Malax: Good catch, I didn't try. I'll figure this out and update the docs accordingly.

README.md
edmorley: There is already an named link of `[libcnb.rs]` used at the top of the doc

Malax: Oooooh, I see now. That's clever.

libcnb-cargo/Cargo.toml
Malax: This has been fixed! :)

README.md
Malax: I added an explicit builder flag in the guide. Again, great catch!

libcnb-cargo/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+rust-version = "1.56"
+license = "BSD-3-Clause"
+description = "Cargo command for managing buildpacks written with libcnb.rs"
+repository = "https://github.com/Malax/libcnb.rs"
+readme = "../README.md"
```
edmorley, line 9: Should `libcnb-cargo` have a different README than to `libcnb`? eg: One at `libcnb-cargo/README.md` that's only a few lines long and instead redirects to the `libcnb` crate? Or else none at all for the moment?

Having the full libcnb project readme for a sub-crate seems like it might be confusing?

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,61 @@
...
+        Ok(vec![])
+    }
+}
+
+#[derive(Debug)]
+pub enum CrossCompileError {
+    CouldNotFindRequiredBinary(String),
+}
+
+/// Returns a human-readable help text about cross compiling from the user's host platform to
```
edmorley, line 45:
```suggestion
/// Returns a human-readable help text about cross-compiling from the user's host platform to
```

edmorley (changes requested):
Thank you so much for this! I've reviewed most things so far, apart from `assemble_buildpack_tarball()` (and also how that might interact with meta-buildpacks). I'll finish the rest by Monday.

Some general comments:
- there are no tests, what do you think should be tested now vs in the future?
- the cargo-make and bash implementations run `strip` on the release binary on Linux (can't run on OS X since only gnu strip can read the cross-compiled binary), whereas `cargo libcnb package` does not. This is likely fine to skip for now, given that Cargo is about to stabilise support for stripping via `Cargo.toml` (rust-lang/cargo/pull/10088), so likely not worth us implementing in the meantime, but mentioning here for completeness.
- I presume we may want to publish compiled binaries for at least Linux to speed up CI - not something to worry about for this PR, but could you file a GitHub issue for this?

For the comments left below, feel free to defer (via filing a GitHub issue) or wontfix as you feel appropriate. I just wanted to mention everything I spotted up-front rather than try and guess what should be a GitHub issue vs not.

Marking requested changes since (a) I still have `assemble_buildpack_tarball` to review, (b) I'd like a glance at the rest post changes (though I don't imagine any issues).

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+                .map(|metadata| {
+                    metadata
+                        .target_directory
+                        .join(default_buildpack_tarball_filename(
+                            &buildpack_data.buildpack_toml,
+                            cargo_profile,
+                        ))
+                        .into_std_path_buf()
+                })
+                .ok()
```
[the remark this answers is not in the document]
edmorley: I also get the above error message, if I run `cargo libcnb package` from a meta-buildpack's directory, since it doesn't have a `Cargo.toml` at all.

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    IoError(std::io::Error),
+    DeserializationError(toml::de::Error),
+}
+
+pub struct BuildpackData<BM> {
+    pub buildpack_toml_path: PathBuf,
+    pub buildpack_toml: BuildpackToml<BM>,
+}
+
+/// Assembles a buildpack tarball and writes it to the given destination path.
```
edmorley, line 131: I'd like to discuss the approach of using a tarfile for output a bit more, however first some additional context may be useful (both for current and future readers), from the current status quo and also my experiments with packaging a couple of months ago.

At the moment there are quite a few issues in the way Cutlass works, which makes CI very slow (15 mins for just the maven specs in the JVM repo for example):
- In order to provide a streamlined local test UX, Cutlass builds and packages buildpacks as part of running tests, rather than it being a setup step prior to cutlass being invoked.
- Running Ruby tests in parallel is typically achieved via `parallel_split_test` or similar, which don't support a "before spawning processes" setup hook / "after all processes have exited" teardown hook.
- As such each concurrent test process has to perform the buildpack build/packaging, which at best duplicates work (even with caches) and at worst results in race conditions around setup/teardown.
- Due to those race conditions, concurrency is disabled in the JVM repo when using Cutlass.

Therefore, in whatever solution we have for packaging (this PR) and integration testing (eg future PR to add a Rust alternative to Cutlass), we need the ability to both perform setup/teardown once and only once, and still run tests concurrently. 

In addition, when testing buildpacks ideally we would just pass a directory path to `--buildpacks` (rather than an OCI image, as Cutlass does) since it reduces the setup/teardown complexity/time (and would likely avoid the race conditions in Cutlass), and actually makes `pack build` runs slightly faster (counter-intuitively, providing a packaged buildpack does not save pack any time).

However at present `pack build --buildpack` doesn't perform any of the steps that `pack buildpack package` performs, so effectively ignores `package.toml` and doesn't pull in dependent buildpacks of a meta-buildpack:
https://github.com/buildpacks/pack/issues/1320

As such if wanting to use `--buildpacks` with a meta-buildpack, one has to either:
1. Pass in all dependent buildpacks via additional `--buildpacks` arguments (though this breaks the defined `order`, so not viable for meta-buildpacks with multiple `order`s).
2. Package the meta-buildpack to a `.cnb` or OCI image first, using `pack buildpack package`, and then use that packaged buildpack in the `--buildpacks` argument instead.

ie: When testing we'll need to:
1. For buildpacks that aren't meta-buildpacks (and are written in Rust):
  - Cross-compile
  - Assemble the binary/symlinks/assets somewhere (eg a directory)
2. For meta-buildpacks:
  - Have a custom test `package.toml` that points to the development versions in the repo from (1)
  - `pack buildpack package` using that `package.toml`

At the moment this PR does (1) by creating a tarfile, rather than a directory.

I like the fact that assembling the tarball is portable (since it's using the `tar` crate, not shelling out to subprocesses), and reading the files straight off disk instead of performing intermediate copies.

However:
- For buildpacks that aren't meta-buildpacks, it seems creating a tarfile is more complex/slower than creating a directory (and unnecessary, since not affected by buildpacks/pack/issues/1320).
- For meta-buildpacks, we would have to use a different implementation anyway (that uses `pack buildpack package`), meaning we don't have any code reuse benefits.

Are there other benefits of using an output tarfile over a directory? I can imagine one might be for platforms where it's problematic to create a symlink (eg Windows), however there are other ways we could work around that, and I suspect it's not an issue in practice for people working on buildpacks written in Rust (since modern Windows does support symlinks in developer mode and also via WSL2 etc).

As a datapoint, using directory format via my bash script takes 450ms to no-op rebuild and package 5x buildpacks in a workspace, compared to 10s for `cargo libcnb package`. Of that 10s, 80% is spent in the tar file creation.

hone: Thanks for the writeup @edmorley, my experience is fairly similar. When doing this by hand, I almost never make a tarball since it's slow.

To that point, should we just not use `pack buildpack package -f` for everything vs making our own tarball?

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+/// problematic characters such as `/`.
+pub fn default_buildpack_tarball_filename<BM>(
+    buildpack_toml: &BuildpackToml<BM>,
+    cargo_profile: CargoProfile,
+) -> String {
+    format!(
+        "{}_{}_{}.tar.gz",
+        buildpack_toml.buildpack.id.replace("/", "_"),
+        buildpack_toml.buildpack.version,
+        match cargo_profile {
```
hone, line 200: should we just implement the `Display` trait on `CargoProfile` or have a `target_dir` fn?

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,218 @@
...
+            |_| String::from("unknown size"),
+            |metadata| SizeFormatterSI::new(metadata.len()).to_string()
+        )
+    );
+
+    info!("Packaging successfully finished!");
+    info!("Hint: To test your buildpack locally with pack, run: pack build my-image -b {} --path /path/to/application", relative_output_path.to_string_lossy());
+}
+
+fn setup_cli_parsing<'a, 'b>() -> clap::App<'a, 'b> {
```
hone, line 182: I'd like to see this moved into it's own file. This is a pretty big `main.rs`. From various CLIs that I've been exposed to in rust will have a [`args.rs`](https://github.com/BurntSushi/ripgrep/blob/master/crates/core/main.rs#L56) or `cli.rs` that encapsulates this.

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+                            Arg::with_name("target")
+                                .long("target")
+                                .default_value("x86_64-unknown-linux-musl")
+                                .help("Build for the target triple"),
+                        )
+                        .arg(
+                            Arg::with_name("output-path")
+                                .long("output")
+                                .short("o")
+                                .help("Write buildpack tarball to this path instead of Cargo's target directory")
```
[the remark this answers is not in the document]
hone: We could also change "tarball" to "output"

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+                                .takes_value(true),
+                        ),
+                ),
+        )
+}
+
+fn setup_logging() {
+    if let Err(error) = stderrlog::new().quiet(false).verbosity(2).init() {
```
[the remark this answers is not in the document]
hone: we could also set this as a constant

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
+use cargo_metadata::MetadataCommand;
```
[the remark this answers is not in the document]
edmorley: Added in f1b1b87433ebb0f60fdcead54316bfe2862a438f

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+                    error!("Hint: Verify that your \"buildpack.toml\" is valid.")
+                }
+            }
+
+            std::process::exit(1);
+        }
+    };
+
+    info!(
+        "Found valid buildpack with id \"{}\" @ {}!",
```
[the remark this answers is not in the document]
Malax: Added in 1f5de4f09ea7ee11f89e2f33f9b28c71835566f1

libcnb-cargo/src/lib.rs
Malax: I don't think we should implement `Display` for `CargoProfile`. Looking at this code, you can see that we map the profile to different strings throughout. For example, Cargo puts the binaries created with the `dev` profile in a directory called `debug`.

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        "Successfully wrote buildpack tarball {} ({})",
+        relative_output_path.to_string_lossy(),
+        output_path
+            .metadata()
+            .map(|metadata| SizeFormatterSI::new(metadata.len()).to_string())
+            .unwrap_or_else(|_| String::from("unknown size"))
+    );
+
+    info!("Packaging successfully finished!");
+    info!("Hint: To test your buildpack locally with pack, run: pack build my-image -b {} --path /path/to/application", relative_output_path.to_string_lossy());
```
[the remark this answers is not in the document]
Malax: Added in 0ee8d90ecc2ba7a27460d0258333542b9e4fb16a

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+}
+
+fn setup_cli_parsing<'a, 'b>() -> clap::App<'a, 'b> {
+    App::new(env!("CARGO_PKG_NAME"))
+        .bin_name("cargo")
+        .setting(AppSettings::SubcommandRequiredElseHelp)
+        .subcommand(
+            App::new("libcnb")
+                .setting(AppSettings::SubcommandRequiredElseHelp)
```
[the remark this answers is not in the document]
Malax: Added in e863d50

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+            .unwrap_or_else(|_| String::from("unknown size"))
+    );
+
+    info!("Packaging successfully finished!");
+    info!("Hint: To test your buildpack locally with pack, run: pack build my-image -b {} --path /path/to/application", relative_output_path.to_string_lossy());
+}
+
+fn setup_cli_parsing<'a, 'b>() -> clap::App<'a, 'b> {
+    App::new(env!("CARGO_PKG_NAME"))
+        .bin_name("cargo")
```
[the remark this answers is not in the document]
Malax: Added in e863d50

libcnb-cargo/Cargo.toml
Malax: Added in 8226d4f

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+    if exit_status.success() {
+        Ok(cargo_metadata
+            .target_directory
+            .join(target_triple.as_ref())
+            .join(match cargo_profile {
+                CargoProfile::Dev => "debug",
+                CargoProfile::Release => "release",
+            })
+            .join(&target.name)
+            .into_std_path_buf())
```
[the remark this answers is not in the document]
Malax: I think it's pretty clear, but hard to gauge when you're the one that wrote it. I introduced a `binary_path` variable as suggested in fd9084b.

libcnb-cargo/src/lib.rs
Malax: > To that point, should we just not use pack buildpack package -f for everything vs making our own tarball?

I'm not sure if I got this correctly, but I don't think adding another dependency on a command line tool is desirable, even if it's a tool that buildpack authors most likely have installed.

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+
+    let target_triple = match matches.value_of("target") {
+        None => {
+            error!("Could not determine target triple!");
+            std::process::exit(1);
+        }
+        Some(target_triple) => target_triple,
+    };
+
+    let current_dir = match std::env::current_dir() {
```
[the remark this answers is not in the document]
Malax: > and this is faster than building each package separately

I wonder why that is, wouldn't sequential `cargo build` commands for each subproject use the same workspace cache regardless?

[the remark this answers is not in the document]
Malax: Filed this as https://github.com/Malax/libcnb.rs/issues/215

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        buildpack_data.buildpack_toml.buildpack.id, buildpack_data.buildpack_toml.buildpack.version
+    );
+
+    let output_path = matches
+        .value_of("output-path")
+        .map(PathBuf::from)
+        .or_else(|| {
+            MetadataCommand::new()
+                .manifest_path(&current_dir.join("Cargo.toml"))
+                .exec()
```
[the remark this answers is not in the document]
Malax: I'll mark this as resolved as this is not a task and having so many open conversations in the PR makes it hard to track which comments still need fixes or discussion. Feel free to mark it as unresolved if you feel it should stay as unresolved.

libcnb-cargo/src/main.rs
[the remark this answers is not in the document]
Malax: Fixed in 0e4e8e9

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,191 @@
...
+pub fn default_buildpack_tarball_filename<BM>(
+    buildpack_toml: &BuildpackToml<BM>,
+    cargo_profile: CargoProfile,
+) -> String {
+    format!(
+        "{}_{}_{}.tar.gz",
+        buildpack_toml.buildpack.id.replace("/", "_"),
+        buildpack_toml.buildpack.version,
+        match cargo_profile {
+            CargoProfile::Dev => "dev",
```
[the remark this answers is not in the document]
Malax: And what would make you think that "debug" isn't part of the version in that case?

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+            error!("Could not determine current directory: {}", io_error);
+            std::process::exit(1);
+        }
+    };
+
+    info!("Reading buildpack metadata...");
+    let buildpack_data = match read_buildpack_data(&current_dir) {
+        Ok(buildpack_data) => buildpack_data,
+        Err(error) => {
+            error!("Packaging buildpack failed due to an error while reading buildpack metadata!");
```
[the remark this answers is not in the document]
Malax: Changed in 53ef3c6

libcnb-cargo/src/main.rs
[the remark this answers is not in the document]
Malax: Fixed in 1e8e24a

libcnb-cargo/src/main.rs
Malax: I'd argue that 200 lines is hardly "big" and honestly I'm totally fine leaving that function in `main.rs`. But to not enter a bikeshed loop, I've moved this out in e1b8090.

libcnb-cargo/src/lib.rs
edmorley: So something to bear in mind for "create tarfile in Rust" vs `pack buildpack package -f` is that pack actually accepts two types of archives:
1. A single buildpack archive where the `bin/` directory is in the root of the archive.
2. A `.cnb` style archive where one or more buildpacks are at the `/cnb/buildpacks/<buildpack ID>/<buildpack version>/` locations described in the [distribution spec](https://github.com/buildpacks/spec/blob/main/distribution.md#buildpackage).

This PR implements style (1), which cannot be used for meta-buildpacks.

In order to support meta-buildpacks, style (2) will need to be used, which means that implementation starts to get more complex, and means we're potentially reimplementing parts of pack, if we stick with creating the tarfile in Rust. 

That may be something we still want to do, however it feels like a decision for later, when meta-buildpacks are actually supported by `cargo libcnb package`?

Malax: Yeah, we have nothing for meta-buildpacks in libcnb as of now. As you said this is something for later.

libcnb-cargo/src/lib.rs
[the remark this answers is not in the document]
edmorley: "dev" is a common semver suffix for say JS packages in-development versions (ie: built from the local repo).

I also more strongly associate "debug" as a word with the build created via not specifying `--release`, rather than "dev". (I know the Cargo profile is "dev", but still). Cargo also uses "debug" for the name of the output directory in `target`, so it seemed to fit better with that.

However I'm happy to just skip this for now if you'd prefer "dev".

libcnb-cargo/src/lib.rs
Malax: Buildpack assembly was replaced with a directory based approach in 7f52227. It currently does not any deletion of existing directories and will fail if the directory already exists. Happy to hear what people think is a good strategy for handling deletion when packaging in run multiple times.

Deleting the output directory is dangerous, if the user accidentally points the output path at `$HOME` we might delete their home directory. But forcing users to delete older versions by hand is also annoying.

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+                BuildpackDataError::IoError(io_error) => {
+                    error!("Unable to read buildpack metadata: {}", io_error);
+                    error!("Hint: Verify that a readable file named \"buildpack.toml\" exists at the root of your project.");
+                }
+                BuildpackDataError::DeserializationError(deserialization_error) => {
+                    error!(
+                        "Unable to deserialize buildpack metadata: {}",
+                        deserialization_error
+                    );
+                    error!("Hint: Verify that your \"buildpack.toml\" is valid.");
```
schneems, line 71:
```suggestion
                    error!("Hint: Verify that your \"buildpack.toml\" is valid TOML.");
```

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Assembly of the directory follows the constraints set by the libcnb framework. For example,
+/// the buildpack binary is only copied once and symlinks are used to refer to it when the CNB
+/// spec requires different file(name)s.
+///
+/// This function will not validate if the buildpack descriptor at the given path is valid and will
+/// use it as-is.
+///
+/// # Errors
+/// Will return `Err` if the buildpack directory could not be assembled.
+pub fn assemble_buildpack_directory(
```
schneems, line 151: In the voice of Captain America: Buildpack directory ASSEMBLE!

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+/// Creates a buildpack directory and copies all buildpack assets to it.
+///
+/// Assembly of the directory follows the constraints set by the libcnb framework. For example,
+/// the buildpack binary is only copied once and symlinks are used to refer to it when the CNB
+/// spec requires different file(name)s.
+///
+/// This function will not validate if the buildpack descriptor at the given path is valid and will
+/// use it as-is.
+///
+/// # Errors
```
schneems, line 149:
```suggestion
/// # Errors
///
```

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+#[derive(Copy, Clone)]
+pub enum CargoProfile {
+    Dev,
+    Release,
+}
+
+/// Reads buildpack data from the given project path.
+///
+/// # Errors
```
schneems, line 111:
```suggestion
/// # Errors
///
```

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+/// installed on the user's machine. When a tool is missing, a `BuildError::CrossCompileError` is
+/// returned which provides additional information. Use the `cross_compile::cross_compile_help`
+/// function to obtain human-readable instructions on how to setup the required tools.
+///
+/// This function currently only supports projects with a single binary target. If the project
+/// does not contain exactly one target, the appropriate `BuildError` is returned.
+///
+/// This function will write Cargo's output to stdout and stderr.
+///
+/// # Errors
```
schneems, line 36:
```suggestion
/// # Errors
///
```

libcnb-cargo/README.md
```diff
@@ -0,0 +1,29 @@
+# libcnb-cargo
+
+Cargo command for managing buildpacks written with [libcnb.rs](https://github.com/malax/libcnb.rs).
+
+## Installation
```
schneems, line 5:
```suggestion
## Installation

```

schneems (commented):
Looks like a good first start. I left comments and some suggestions.

I would like to see the command for running the CLI in the docs. I was going to try to suggest this change, but can't (because the last line has backticks which I'm not sure how to escape).


    ## Execute locally

    From the top-level repo directory, run:

    ```
    $ cargo build && target/debug/cargo-libcnb libcnb --help
    ```

libcnb-cargo/src/lib.rs
schneems: > [cutlass] eg creating OCI images, rather than leaving as directories 

- Fixed in https://github.com/heroku/cutlass/pull/37 (needs review please 🙏)
- Possible proposal to address the concurrency issue https://github.com/heroku/cutlass/pull/38

hone:
I much prefer the `structopt` pattern with using derive Struct pattern, but `structopt` won't be integrated into `clap` until 3 and that is still in flight. :(

https://github.com/clap-rs/clap/issues/2869

libcnb-cargo/src/lib.rs
hone: > I'm not sure if I got this correctly, but I don't think adding another dependency on a command line tool is desirable, even if it's a tool that buildpack authors most likely have installed.

I think we should just get out of the `tar` business which you already did. If a user wants to create a tarball, they can use `pack` like everyone else vs us re-inventing the wheel.

hone: > Deleting the output directory is dangerous, if the user accidentally points the output path at $HOME we might delete their home directory. But forcing users to delete older versions by hand is also annoying.

We don't default to `$HOME`, I'm not sure I'm too worried about this edge case. I personally would prefer to have it clear / rebuild. We can add configure options if you really want.

README.md
```diff
@@ -13,79 +13,226 @@
...
+quick start guide.
+
+### Running the Buildpack
+You might have seen in the hint in the output of the packaging step that we're now ready to run our buildpack locally 
+with `pack`. Before we can do this, we need an application to build. Since our buildpack does not interact with the
+application code at all, we just create an empty directory and use that as our application:
+
+```shell
+$ mkdir bogus-app
+$ pack build my-image -b target/libcnb-examples_my-buildpack_0.1.0_dev.tar.gz --path bogus-app --builder heroku/buildpacks:20
```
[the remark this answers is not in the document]
Malax: Yes, we don't know which default builder the user has set in `pack`, but this buildpack requires `heroku-20`.

README.md
```diff
@@ -13,79 +13,226 @@
...
+    // according to the returned value, handle both writing the build plan and exiting with
     // the correct status code for you.
-    fn detect(&self, context: DetectContext<Self>) -> libcnb::Result<DetectResult, Self::Error> {
+    fn detect(&self, _context: DetectContext<Self>) -> libcnb::Result<DetectResult, Self::Error> {
         DetectResultBuilder::pass().build()
     }
 
-    // Similar to detect, this method will be called when the CNB lifecycle executes the 
+    // Similar to detect, this method will be called when the CNB lifecycle executes the
     // build phase.
```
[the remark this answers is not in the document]
Malax: "build" is the proper name of the phase: https://github.com/buildpacks/spec/blob/74f91666fc22e7568acdd0ca9e2e314ff150d740/buildpack.md#phase-3-build

README.md
```diff
@@ -13,79 +13,226 @@
...
         println!("Build runs on stack {}!", context.stack_id);
-        BuildResultBuilder::new().build()
+
+        BuildResultBuilder::new()
+            .launch(Launch::new().process(Process::new(
+                process_type!("web"),
+                "echo",
+                vec!["Hello World!"],
+                false,
+                true,
```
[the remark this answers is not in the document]
Malax: This is something for `Process` itself, and IMO out of scope for this PR. Do you want to file an issue for improving `Process::new`?

README.md
```diff
@@ -13,79 +13,226 @@
...
+INFO - Building buildpack binary (x86_64-unknown-linux-musl)...
+   Compiling syn v1.0.82
+   Compiling bit-vec v0.6.3
+   Compiling regex-syntax v0.6.25
+   # Omitting further compilation output...
+    Finished dev [unoptimized + debuginfo] target(s) in 19.71s
+INFO - Writing buildpack directory...
+INFO - Successfully wrote buildpack directory: ../../target/libcnb-examples_ruby_0.1.0_dev (103.6M)
+INFO - Packaging successfully finished!
+INFO - Hint: To test your buildpack locally with pack, run: pack build my-image --buildpack ../../target/libcnb-examples_ruby_0.1.0_dev --path /path/to/application
```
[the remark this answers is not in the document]
Malax: This is caused by incorporating suggestions from review comments and not updating the example output. I'll update them after the dust settles on this PR.

libcnb-cargo/src/cli.rs
```diff
@@ -0,0 +1,37 @@
...
+                        .arg(
+                            Arg::with_name("release")
+                                .long("release")
+                                .help("Build in release mode, with optimizations"),
+                        )
+                        .arg(
+                            Arg::with_name("target")
+                                .long("target")
+                                .default_value("x86_64-unknown-linux-musl")
+                                .help("Build for the target triple"),
```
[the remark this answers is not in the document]
Malax: This is intentionally copied from Cargo itself as it is a pass-through: 

```shell
$ cargo build --help
...
        --target <TRIPLE>...         Build for the target triple
...
```

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,187 @@
...
+        CargoProfile::Dev
+    };
+
+    let target_triple = match matches.value_of("target") {
+        None => {
+            error!("Could not determine target triple!");
+            std::process::exit(1);
+        }
+        Some(target_triple) => target_triple,
+    };
```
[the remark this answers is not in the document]
Malax: That's incorrect. We support any target triple as we just do a pass-through to Cargo. Users can configure compile options for the target platform in their `Cargo.toml` if they need to. IMO, this tool should not intentionally cut features from Cargo itself.

The cross-compile code in this PR is merely a helper to make it easier to set it up.

libcnb-cargo/src/main.rs
Malax: This goes beyond valid TOML. It also verifies if the file conforms to the CNB spec. It will fail is required fields from `buildpack.toml` are missing or of a wrong type for example.

libcnb-cargo/src/lib.rs
Malax: image

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,61 @@
...
+pub fn cross_compile_env(
+    target_triple: impl AsRef<str>,
+) -> Result<Vec<(OsString, OsString)>, CrossCompileError> {
+    if target_triple.as_ref() == "x86_64-unknown-linux-musl" && cfg!(target_os = "macos") {
+        let ld_binary_name = "x86_64-linux-musl-ld";
+        let cc_binary_name = "x86_64-linux-musl-gcc";
+
+        Ok(vec![
+            (
+                OsString::from("CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER"),
```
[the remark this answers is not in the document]
Malax: Added in f53b0ed

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,61 @@
...
+                which(ld_binary_name)
+                    .map_err(|_| {
+                        CrossCompileError::CouldNotFindRequiredBinary(String::from(ld_binary_name))
+                    })?
+                    .into_os_string(),
+            ),
+            (
+                OsString::from("CC_x86_64_unknown_linux_musl"),
```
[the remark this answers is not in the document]
Malax: Added in f53b0ed

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,61 @@
...
+    target_triple: impl AsRef<str>,
+) -> Result<Vec<(OsString, OsString)>, CrossCompileError> {
+    if target_triple.as_ref() == "x86_64-unknown-linux-musl" && cfg!(target_os = "macos") {
+        let ld_binary_name = "x86_64-linux-musl-ld";
+        let cc_binary_name = "x86_64-linux-musl-gcc";
+
+        Ok(vec![
+            (
+                OsString::from("CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_LINKER"),
+                which(ld_binary_name)
```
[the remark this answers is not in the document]
Malax: Fixed in f53b0ed

libcnb-cargo/src/lib.rs
Malax: I'm marking this as resolved as the tar generation has been replaced by a directory based approach in 7f52227. Feel free to add review comments to the new code.

libcnb-cargo/src/main.rs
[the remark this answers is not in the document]
Malax: This has been changed together with the switch from tar to directory output.

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,61 @@
...
+            (
+                OsString::from("CC_x86_64_unknown_linux_musl"),
+                which(cc_binary_name)
+                    .map_err(|_| {
+                        CrossCompileError::CouldNotFindRequiredBinary(String::from(cc_binary_name))
+                    })?
+                    .into_os_string(),
+            ),
+        ])
+    } else {
```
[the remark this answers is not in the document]
Malax: Added in 382098e

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,65 @@
...
+        Ok(vec![])
```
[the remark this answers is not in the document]
Malax: It is not meant to provide cross-compile configuration for all possible host/target combinations. It aims to provide defaults for the most common case. If people have exotic buildpack targets, they can configure their toolchain manually (in `Cargo.toml` and/or their local env). I don't think we should error when we cannot provide the optional help for cross-compilation.

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,65 @@
...
+
+/// Constructs a set of environment variables to enable cross-compiling from the user's host
+/// platform to the desired target platform.
+///
+/// This function will not install required toolchains, linkers or compilers automatically. It will
+/// look for the required tools and returns an error if they cannot be found. See
+/// [`cross_compile_help`] on how to support users setting up their machine for cross-compilation.
+///
+/// # Errors
+/// Will return `Err` if the cross-compile environment could not be created for the current host
```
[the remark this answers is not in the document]
Malax: This follows the standard Rust conventions when documenting errors. See: https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+        None => {
+            error!("Could not determine output path for tarball!");
+            std::process::exit(1);
+        }
+    };
+
+    let relative_output_path =
+        pathdiff::diff_paths(&output_path, &current_dir).unwrap_or_else(|| output_path.clone());
+
+    info!("Building buildpack binary ({})...", &target_triple);
```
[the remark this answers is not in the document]
Malax: We surely could. I included the target_triple there because it's not part of the Cargo build output. The profile will be logged by Cargo itself:

```
INFO - Building buildpack binary (x86_64-unknown-linux-musl)...
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
```

If you feel strongly, I'm happy to add it to the heading. But I'd not add it to have less noise in the already quite noisy `handle_libcnb_package` function.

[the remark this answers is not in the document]
edmorley: > The profile will be logged by Cargo itself:

Ah good point. I'm happy to skip this then :-)

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Constructs a set of environment variables to enable cross-compiling from the user's host
+/// platform to the desired target platform.
+///
+/// This function will not install required toolchains, linkers or compilers automatically. It will
+/// look for the required tools and returns an error if they cannot be found. See
+/// [`cross_compile_help`] on how to support users setting up their machine for cross-compilation.
+pub fn cross_compile_env(
+    target_triple: impl AsRef<str>,
+) -> Result<Vec<(OsString, OsString)>, CrossCompileError> {
+    if target_triple.as_ref() == "x86_64-unknown-linux-musl" && cfg!(target_os = "macos") {
```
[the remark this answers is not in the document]
Malax: Added 81d9a61

Malax:
> I'd like to revisit this b/c I think it looks much cleaner/easier to maintain.

We can always revisit this. It's kinda the whole idea behind this PR - adding a good base so that we can iterate. It really applies to all of libcnb.rs - we're early and nothing is set in stone.

README.md
```diff
@@ -13,79 +13,226 @@
...
     //
-    // More generic errors that happen during buildpack execution such as I/O errors while 
+    // More generic errors that happen during buildpack execution such as I/O errors while
     // writing CNB TOML files are handled by libcnb.rs itself.
     type Error = GenericError;
 
-    // This method will be called when the CNB lifecycle calls detect. Use the DetectContext to 
-    // access CNB data such as the stack this buildpack is currently executed on, the app 
-    // directory and similar things. When using libcnb.rs, you never have to read environment 
+    // This method will be called when the CNB lifecycle calls detect. Use the `DetectContext` to
```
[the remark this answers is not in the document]
Malax: Can you open a dedicated PR for this if you feel strongly? This comment wasn't added in this PR and I'd like to keep this as clean as possible.

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,214 @@
...
+    let binary_path = match build_buildpack_binary(&current_dir, cargo_profile, &target_triple) {
+        Ok(binary_path) => binary_path,
+        Err(error) => {
+            error!("Packaging buildpack failed due to a build related error!");
+
+            match error {
+                BuildError::IoError(io_error) => {
+                    error!("IO error while executing Cargo: {}", io_error);
+                }
```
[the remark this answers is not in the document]
edmorley: > We could wrap this IO error in another error, but it seems overkill to me. Did you have something in mind when you wrote this?

Let's skip

Malax:
@schneems 
> I would like to see the command for running the CLI in the docs.

The standard `cargo run` should just work fine:
```
$ cargo run -- libcnb --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `/Users/manuel.fuchs/projects/libcnb.rs/target/debug/cargo-libcnb libcnb --help`
cargo-libcnb 0.1.0
Allows working with buildpacks written with libcnb.rs

USAGE:
    cargo libcnb <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

SUBCOMMANDS:
    help       Prints this message or the help of the given subcommand(s)
    package    Packages a libcnb.rs Cargo project as a Cloud Native Buildpack
```

We can surely add this to the README. If you ask me, it's not worth adding though. We're talking about one of the most basic Cargo commands to run a binary and end-users are supposed to install the binary with `cargo install` anyway.

libcnb-cargo/src/lib.rs
```diff
@@ -0,0 +1,202 @@
...
+///
+/// Will return `Err` if the buildpack directory could not be assembled.
+pub fn assemble_buildpack_directory(
+    destination_path: impl AsRef<Path>,
+    buildpack_toml_path: impl AsRef<Path>,
+    buildpack_binary_path: impl AsRef<Path>,
+) -> std::io::Result<()> {
+    fs::create_dir_all(destination_path.as_ref())?;
```
edmorley, line 159: I think we should remove the directory if it already exists, since making users delete it before re-running `cargo libcnb package` isn't ideal for UX.

My understanding is that the concern is over accidentally deleting a user-provided `output-path` (eg if they specify `~/`)?

Possible options:
1. Remove `output-path` for now and use only the default paths, which are safe to delete.
2. Change `output-path` to instead be a containing directory, rather than the actual output directory. For example `--output-path ~/foo` would result in the deletion/creation of `~/foo/my_buildpack/`, which is safe to delete.
3. Leave `output-path` as is for now, and perform the directory deletion only iff `output-path` is not used.

It's worth noting that when we add workspace support to `cargo libcnb package` in a future PR, we're going to have to do (2) for workspaces at the least. As such, `output-path` is more complex than it seems at first glance, and since this PR is an MVP I think it may be best to just drop that feature for now (and add it back in a later PR).

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,190 @@
...
+        Err(error) => {
+            error!("Could obtain metadata from Cargo: {}", error);
+            std::process::exit(1);
+        }
+    };
+
+    let output_path = matches.value_of("output-path").map_or_else(
+        || {
+            cargo_metadata
+                .target_directory
```
edmorley, line 98: The default directory location generated here is of form:
`target/debug/libcnb-examples_basics_0.1.0/`

Whereas the assets for the cross-compiled builds are inside the appropriate target triple directory:
`target/x86_64-unknown-linux-musl/debug/`

So if we're being consistent, the path should probably be something like:
`target/x86_64-unknown-linux-musl/debug/libcnb-examples_basics_0.1.0/`

Although that path is now very long, and annoying to deal with if people are calling pack by hand on the CLI.

What do you think about removing both the {debug/release} and buildpack version components from the directory path?

IMO including them:
- doesn't really add much value (packaging is super fast, overwriting isn't an issue for switching back and forth between debug and release; we don't really need the version in the path etc)
- makes a bunch of use-cases more annoying (eg manually calling pack or other non-Rust-library usages, and having to adjust paths repeatedly, as versions change etc)
- means over time lots of old version directories will accumulate, affecting things like tab-complete for non-Rust-library usages

Removing those would give something like:
```
target/libcnb-examples_basics/
target/libcnb-examples_ruby/
```

Or if we wanted to not clutter up target, then something like:

```
target/dist/libcnb-examples_basics/
target/dist/libcnb-examples_ruby/
```

edmorley (changes requested):
Thank you for all the changes :-)

libcnb-cargo/src/lib.rs
Malax: I still believe that implicitly deleting is dangerous in any case, even if we don't allow custom output directories. The situation was a bit easier when we had tarballs instead since they were atomic.

I agree that output-path is tricky in the workspace scenario and removal of that feature might be the best option at this point. Since that also makes deletion much less dangerous, I'm happy to implement that as well. I'll add a comment when I'm done with that.

libcnb-cargo/src/main.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+    };
+
+    info!("Writing buildpack directory...");
+    if let Err(io_error) = assemble_buildpack_directory(
+        &output_path,
+        &buildpack_data.buildpack_toml_path,
+        &binary_path,
+    ) {
+        error!("IO error while writing buildpack directory: {}", io_error);
```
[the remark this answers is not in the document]
Malax: Yeah, it'll go away :)

schneems:
> We can surely add this to the README. If you ask me, it's not worth adding though. We're talking about one of the most basic Cargo commands to run a binary and end-users are supposed to install the binary with cargo install anyway.

I was referring to having it for local dev debugging/execution. I trust my tests and all, but sometimes I just want to exercise things manually.

That command you've provided doesn't work in the libcnb dev directory:

```
⛄️ 3.0.3 🚀 /Users/rschneeman/Documents/projects/work/libcnb.rs (malax/cargo)
$ cargo run -- libcnb --help
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: cargo-libcnb, example-01-basics, example-02-ruby-sample
⛄️ 3.0.3 🚀 /Users/rschneeman/Documents/projects/work/libcnb.rs (malax/cargo)
$ cargo build
    Finished dev [unoptimized + debuginfo] target(s) in 0.44s
⛄️ 3.0.3 🚀 /Users/rschneeman/Documents/projects/work/libcnb.rs (malax/cargo)
$ cargo run -- libcnb --help
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: cargo-libcnb, example-01-basics, example-02-ruby-sample
```

edmorley:
@schneems That's because this repository is using the Cargo workspaces feature (there are multiple crates here).

The command to use is still `cargo run` (you should never need to manually build and run the binary), but it needs to be told which crate to run. The error message you pasted mentioned how to do this, and listed the binary names available. 

Using the method from the error message, gives:

`cargo run --bin cargo-libcnb`

In addition, these alternatives work too:
`cargo run --package libcnb-cargo`
`cd cargo-libcnb && cargo run`

These alternatives also work for any other cargo command, for example when you want to run `cargo test` against a single package (to save running all tests):
`cargo test --package  libcnb-cargo` (`-p` can also be used instead of `--package`)

There is more on this here:
https://doc.rust-lang.org/cargo/reference/workspaces.html

README.md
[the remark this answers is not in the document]
schneems: https://github.com/Malax/libcnb.rs/pull/216

libcnb-cargo/src/cross_compile.rs
```diff
@@ -0,0 +1,91 @@
...
+
+The easiest way to install 'x86_64-linux-musl-gcc' is to follow the instructions in the linked
+GitHub repository:
+
+https://github.com/FiloSottile/homebrew-musl-cross"#,
+        ),
+    );
+
+    help_texts.insert(
+        ("linux", X86_64_UNKNOWN_LINUX_MUSL),
```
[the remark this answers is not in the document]
hone: @edmorley I have this in #214, maybe we can punt on improving it there?

libcnb-cargo/src/lib.rs
hone: :+1: on just removing the configuration for now.

Another suggestion is we can just use a tmpdir for each build vs a static location and just output where it's built. This way we know it's safe to prune/delete/recreate.

Malax: Resolved with 7f35903 and 80d2cc2

libcnb-cargo/src/main.rs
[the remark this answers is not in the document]
Malax: I both improved the error message (which should never occur when using the CLI) and added removal of the directory in 80d2cc2.

libcnb-cargo/src/cross_compile.rs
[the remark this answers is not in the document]
Malax: I refactored the way the cross-compilation functions work. With Linux, we don't need to set environment variables, but we want to check for binaries. This made both the naming and design of `cross_compile_env` not a good fit. It was replaced by a `cross_compile_assistance` function which can much easier extended in the future.

Also, splitting the help text from the actual checks was clunky. `cross_compile_assistance` now returns the help text instead of an error if the cross-compile settings could not be determined. This removes the burden from the call-site to first detect the environment could not be set and then display the help text. With this change, the help text can change when the issue changes. Right now we only have single checks, but it will be useful when we have to do more.

The change also made it clearer that it is not an error if there are no cross-compile settings for a target platform as this is just an heuristic at best.

Oh, and the Linux help text is now displayed if `musl-gcc` is missing. ;-)

See: fb16494

libcnb-cargo/src/main.rs
Malax: I think both version and Cargo profile are important. I can see removing the version from the path since it can be read from the `buildpack.toml`, but the Cargo profile used to compile the buildpack should be there IMO.


WDYT about smth. like this?
```
target/buildpack/debug/libcnb-examples_basics
target/buildpack/debug/libcnb-examples_ruby
```

edmorley: > It's still a long path, but it will be stable across subsequent invocations

Yeah this was my primary concern previously. The proposed path works for me :-)

Malax: Added here: 53265524c3bcd4bea4a56fa9ad6efceb11cb7a72

edmorley (approved):
Thank you! :-)

Malax:
@schneems @hone: @edmorley has approved, but since you had input before I'll not merge yet.

hone:
@Malax this is really a net new project/crate. With my other PR and some other suggestions I feel more blocked as it stands now without it being merged. Let's get this in and open PRs against this existing code base IMO.

hone (approved):

Malax:
> Let's get this in and open PRs against this existing code base IMO.

Exactly the approach I was thinking too: 

> I consider this an MVP, there are many improvements that could be made and it could have more features. We'll implement these features and fixes along the way.

Thanks for the approve!

schneems:
> Using the method from the error message, gives:

I tried running as:

```
$ cargo run --bin cargo-libcnb --help
```

It needs:

```
$ cargo run --bin cargo-libcnb -- --help
```

Which makes sense. But it's still nice to not make people guess/hunt. My comment is less about me being able to figure it out, and more about being in the mind space that whoever is dev-ing will be tired/forgetful/new/whatever to one or more of those concepts. 

Where I'm coming from: I appreciate that this might be common expected knowledge from rust devs. In Ruby where every project ever has a `Gemfile` and bundler, every gem library still explains to people how to `bundle install`. For 99% of people, it's duplicate info. For that 1%, it's exactly what they need. 

I'm in the mind space where if I have to look something up once, it's almost always worth asking "why", then retracing my steps to finding somewhere to make the next time easier (if not eliminating the need altogether). I wanted to run the CLI, couldn't immediately figure it out...so I wanted to document it. Which has the added bonus (for me) of a knowledge share showing more canonical/less-hacky ways to get to it.

schneems (approved):
🚀⛴🛳🚢🛸

PR #200, opened 2021-11-24
Add unit tests for `libcnb_data::buildpack::stack`

edmorley:
Also removes overlapping unit tests from `libcnb_data::buildpack`, since we don't need to test every single case in the `buildpack.toml` integration tests (note: those integration tests still need some work, which will occur in later PRs).

GUS-W-10222615.

### DIFF
--- a/libcnb-data/src/buildpack/stack.rs
+++ b/libcnb-data/src/buildpack/stack.rs
@@ -48,3 +48,69 @@ pub enum StackError {
     #[error("Invalid Stack ID: {0}")]
     InvalidStackId(#[from] StackIdError),
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deserialize_specific_stack_without_mixins() {
+        let toml_str = r#"
+id = "heroku-20"
+"#;
+        assert_eq!(
+            toml::from_str::<Stack>(toml_str),
+            Ok(Stack::Specific {
+                // Cannot use the `stack_id!` macro due to: https://github.com/Malax/libcnb.rs/issues/179
+                id: "heroku-20".parse().unwrap(),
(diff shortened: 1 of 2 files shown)


libcnb-data/src/buildpack/stack.rs
```diff
@@ -48,3 +48,69 @@ pub enum StackError {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deserialize_specific_stack_without_mixins() {
+        let toml_str = r#"
+id = "heroku-20"
+"#;
+        assert_eq!(
+            toml::from_str::<Stack>(toml_str),
```
edmorley, line 62: The existing TOML parsing tests use `raw` as the variable name here, however I'm not that keen on it, since it's a bit vague (eg `toml::from_str::<Stack>(raw)`). As part of cleaning up the tests I was going to standardise on something else. Open to ideas :-)

libcnb-data/src/buildpack/stack.rs
```diff
@@ -48,3 +48,69 @@ pub enum StackError {
...
+        assert_eq!(toml::from_str::<Stack>(toml_str), Ok(Stack::Any));
+    }
+
+    #[test]
+    fn reject_specific_stack_with_invalid_name() {
+        let toml_str = r#"
+id = "io.buildpacks.stacks.*"
+"#;
+        let err = toml::from_str::<Stack>(toml_str).unwrap_err();
+        assert!(err
```
edmorley, line 100: For the failure cases, I've intentionally used an intermediate variable (unlike the success cases), since in the event the test fails with a different error message, it means one can more easily insert additional statements here before the assert to debug further (for those not using the debugger).

Malax (approved):

PR #202, opened 2021-11-24
Add more unit tests for `libcnb_data::buildpack::api`

edmorley:
The tests have also been changed to exercise parsing of TOML rather than just using `BuildpackApi::from_str()`, since:
- previously the `TryFrom` and serde `Derserialize` implementations were untested (when they are actually what's used in the library), and this approach now tests both
- IMO we should drop support for parsing strings directly (ie: remove the `FromStr` implementation in favour of just having `TryFrom`), which we can now do in a later PR without needing to adjust the tests.

GUS-W-10222615.

### DIFF
--- a/libcnb-data/src/buildpack/api.rs
+++ b/libcnb-data/src/buildpack/api.rs
@@ -20,6 +20,9 @@ impl TryFrom<BuildpackApiUnchecked> for BuildpackApi {
     }
 }
 
+/// The Buildpack API version.
+///
+/// This MUST be in form `<major>.<minor>` or `<major>`, where `<major>` is equivalent to `<major>.0`.
 #[derive(Deserialize, Debug, Eq, PartialEq)]
 #[serde(try_from = "BuildpackApiUnchecked")]
 pub struct BuildpackApi {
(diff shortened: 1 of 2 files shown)


Malax:
Can we use [serde_test](https://docs.serde.rs/serde_test/) here? This would remove the need to the `TestBuildpack` struct and `toml` as we would test `Deserialize` with serdes data model. We already have an example how to do this here: https://github.com/Malax/libcnb.rs/blob/5e0d592860d6c6a3924ab2565b5f31c05b0c5498/libcnb-data/src/newtypes.rs#L206-L213

edmorley:
> Can we use [serde_test](https://docs.serde.rs/serde_test/) here?

It's a lot cleaner, and I would like to eventually switch to that, however trying now it doesn't play nicely with the shadow struct:

```
---- buildpack::api::tests::deserialize_valid_api_versions stdout ----
thread 'buildpack::api::tests::deserialize_valid_api_versions' panicked at 'tokens failed to deserialize: invalid type: string "1.3", expected tuple struct BuildpackApiUnchecked', /Users/emorley/.cargo/registry/src/github.com-1ecc6299db9ec823/serde_test-1.0.130/src/assert.rs:164:19
```

I was going to remove the shadow struct for the buildpack API case (since it's unnecessary) in a later PR - ok if I switch to `serde_test` then? I'd rather avoid making non-test changes at the same time as adding tests here.

edmorley:
@Malax Could you approve the PR? I've opened #203 (currently based on this branch, will rebase once this merges), which removes `BuildpackApiUnchecked`, which unblocks switching to `serde_test`.

Malax (approved):

PR #203, opened 2021-11-25
Deserialize BuildpackApi directly instead of via a shadow struct

edmorley:
- Removes `BuildpackApiUnchecked`, since it is unnecessary as it only wrapped a `String`.
- Moves the parsing from the `FromStr` to the `TryFrom` implementation, and removes the now unused `FromStr` implementation.
- Switches tests from manual `toml::from_str()` calls to using helpers from `serde_test`: https://serde.rs/unit-testing.html

GUS-W-10229170.


Malax (approved):
🎉

Issue #211, opened 2021-11-29
`libcnb_data::buildpack::Buildpack` should not permit meta-buildpacks that define `stacks`


edmorley:
libcnb-data's `Buildpack` type permits (and actually requires) having both `stacks` and `order` defined (ie meta-buildpacks with `stacks`), however pack forbids this:

```
ERROR: failed to build: downloading buildpack: extracting from file:///Users/emorley/src/testcase: reading buildpack: invalid buildpack.toml: buildpack heroku/example@0.1.0: cannot have both stacks and an order defined
```

As does the spec, however the wording makes it hard to spot:

> A buildpack descriptor that specifies `order` MUST be resolvable into an ordering of buildpacks that implement the [Buildpack Interface](https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpack-interface).

https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpack-implementations

The "Buildpack Interface" means `bin/detect` etc, so excludes meta-buildpacks.

GUS-W-10289129.

edmorley:
We need to decide whether to:
1. Have separate `Buildpack` and `Metabuildpack` types (perhaps as variants of an enum), which are identical other than the former having a mandatory `stacks` field, and the latter a mandatory `order` field. (And handle deserialisation similar to how the Stack any/specific variants are implemented.)
2. Continue to have one `Buildpack` type, remove the existing ` #[serde(default)]` from `order`, and make both `stacks` and `order` an `Option`, perhaps along with some additional custom validation to ensure only one is `None` at a time.
3. Continue to have one `Buildpack` type, and add ` #[serde(default)]` to `stacks` (like we have on `order`), perhaps along with some additional custom validation to ensure only one is empty at a time.

Option 1/2 seem best to me.

@Malax / @hone, do you have a preference?

Malax:
I strongly prefer the first option. Out of the three mentioned, it's the only one that actually makes impossible states impossible whereas the others rely on validation code that might or might not run. Having a strictly correct data model is something I value highly.

There's a downside to this though. If we make `Buildpack` an enum with variants for regular and meta buildpacks, DX within libcnb for buildpack implementations suffers as any code that wants to access buildpack data has to handle the metadata case as well, even if that would in practice never happen since meta buildpacks don't have code.

[...]

PR #212, opened 2021-11-29
Add remaining unit tests for `libcnb_data::buildpack`

edmorley:
This adds the last of the missing unit tests for `libcnb_data::buildpack`.

The `Eq` and `PartialEq` Derive attributes had to be added so that the assertions in the new tests could compare the relevant structs.

Whilst working on this I found #211. Until that is fixed, the meta-buildpack tests will need to continue to have an invalid `[[stacks]]` table.

GUS-W-10222615.


hone (approved):
thanks Ed!

Malax (approved):

PR #216, opened 2021-12-01
Specify interface in readme docs

schneems:
https://github.com/Malax/libcnb.rs/pull/199/files/8344f9a9dbca44e7547b8538dd4bdbbd7448512a#r759646147

Not touching existing whitespace

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,10 +132,10 @@ impl Buildpack for HelloWorldBuildpack {
     // writing CNB TOML files are handled by libcnb.rs itself.
     type Error = GenericError;
 
-    // This method will be called when the CNB lifecycle calls detect. Use the `DetectContext` to
-    // access CNB data such as the stack this buildpack is currently executed on, the app
-    // directory and similar things. When using libcnb.rs, you never have to read environment
-    // variables or read/write files to disk to interact with the CNB lifecycle.
+    // This method will be called when the CNB lifecycle executes the detect phase (`bin/detect`).
+    // Use the `DetectContext` to access CNB data such as the stack this buildpack is currently
+    // executed on, the app directory and similar things. When using libcnb.rs, you never have
+    // to read environment variables or read/write files to disk to interact with the CNB lifecycle.
     //
     // One example of this is the return type of this method. `DetectResult` encapsulates the
     // required exit code as well as the data written to the build plan. libcnb.rs will,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,7 +78,7 @@ impl Buildpack for HelloWorldBuildpack {
     }
 
     // Similar to detect, this method will be called when the CNB lifecycle executes the 
-    // build phase.
+    // `bin/build` phase.
```
Malax, line 81: As commented on #199, "build phase" is a proper CNB term and the docs match what happens during that phase.

README.md
```diff
@@ -64,7 +64,7 @@ impl Buildpack for HelloWorldBuildpack {
     // writing CNB TOML files are handled by libcnb.rs itself.
     type Error = GenericError;
 
-    // This method will be called when the CNB lifecycle calls detect. Use the DetectContext to 
+    // This method will be called when the CNB lifecycle calls `bin/detect`. Use the DetectContext to
```
Malax, line 67: I agree that this is inconsistent with the docs for `build`, IMO this should be `executes the detection phase`.

Malax (commented):
I'm not sure how I feel about adding binary names to the `Buildpack` docs when the whole point of it is to try to abstract from the low-level interface.

"detection phase" and "build phase" are CNB terms from the spec and describe a phase of buildpack execution instead of a concrete low-level interface (calling binaries). Of course, the spec talks about calling these binaries, but I think we're at a higher level with this trait.

https://github.com/buildpacks/spec/blob/main/buildpack.md#phase-1-detection
https://github.com/buildpacks/spec/blob/main/buildpack.md#phase-3-build

I'll not mark this as "request changes" since this is PR is really minor in the grand scheme of things and wait for someone else to chime in. @edmorley maybe? :)

edmorley:
I agree that the docs are inconsistent at the moment, in that they say:
> calls detect
and 
> executes the build phase

We should either say `calls <binary name>` or `executes the <phase> phase` for both.

I'm leaning towards the latter.

schneems:
My desire with putting the name in there is to clear up ambiguity for developers who are coming from writing a buildpack in other languages. "Build phase" is somewhat ambiguous to me. The word "build" tends to be overloaded. It's used in the concept of a builder. We build a buildpack so we can pass it to a lifecycle where the build phase is executed so it can build a docker image.

My thought in specifying `bin/build` was to clear up some of that ambiguity and peel back a tiny bit of magic. When I first came to a libcnb project, my first question was "where is `bin/build`" and I was surprised to not find it. I think having `bin/build` on the readme, even if it's only in the docs will help.

> We should either say calls <binary name> or executes the <phase> phase for both.

I updated it to do both. WDYT?

edmorley:
That reasoning makes sense to me + I like it now having both :-)

Malax (approved):
Still not super keen on the binary names in the docs, but I can get behind your reasoning @schneems! :) Let's add this in!

PR #218, opened 2021-12-02
Extend test_layer_env_delta_fs_read_write for LayerEnv

### LINKED ISSUE #206: Add test for `LayerEnv`: Ensure file extensions are correct
#198 contains a fix for an issue around file extensions when a `LayerEnv` is written to disk. We need to add a test for this case to avoid regressions.

GUS-W-10243213.

Malax:
It now uses all available `ModificationBehavior` variants. This would've caught the issue fixed in PR #198.

Fixes #206, GUS-W-10243213


edmorley (approved):
Ah even better, can adapt the existing test! I hadn't seen there was an existing one :-)

PR #222, opened 2021-12-07
Tests for `Layer` and `BuildContext::handle_layer`

### LINKED ISSUE #169: Add tests for `BuildContext::handle_layer` and `Layer`
PR #144 does not have tests for the `BuildContext::handle_layer` method as it requires some extra work to make it easily testable, especially since it uses `BuildContext` and all it's related values.

The same is true for default implementations in `Layer`.

GUS-W-10222533.

Malax:
We decided to add tests for the refactored `Layer` and related `handle_layer` outside of the original PR #144. This PR adds tests for `Layer` and `BuildContext::handle_layer`, testing all scenarios of create and update combinations. 

Since the tests require a `Layer` and therefore a `Buildpack`, the code is quite long. I've chosen to not split the tests up as I think it makes the tests harder to read. Many tests share some common logic such as testing for specific files written to the layer directory. For the same reason, I choose to not abstract these cases out to a) make changes to the tests easy and b) to keep the test logic where it belongs: the test itself.

Fixes #169, GUS-W-10222533

### DIFF
--- a/libcnb/src/layer/handling.rs
+++ b/libcnb/src/layer/handling.rs
@@ -107,6 +107,10 @@ fn handle_create_layer<B: Buildpack + ?Sized, L: Layer<Buildpack = B>>(
 ) -> Result<LayerData<L::Metadata>, HandleLayerErrorOrBuildpackError<B::Error>> {
     let layer_dir = context.layers_dir.join(layer_name.as_str());
 
+    fs::create_dir_all(&layer_dir)
+        .map_err(HandleLayerError::IoError)
+        .map_err(HandleLayerErrorOrBuildpackError::HandleLayerError)?;
+
     let layer_result = layer
         .create(context, &layer_dir)
         .map_err(HandleLayerErrorOrBuildpackError::BuildpackError)?;
--- a/libcnb/src/layer/test.rs
+++ b/libcnb/src/layer/test.rs
@@ -656,13 +656,13 @@ fn error_handling_no_directory() {
     let context = build_context(&temp_dir);
     let layer_name = random_layer_name();
 
-    let layer_dir = temp_dir
+    let layer_toml_path = temp_dir
         .path()
         .join("layers")
         .join(format!("{}.toml", layer_name.as_str()));
 
     fs::write(
-        &layer_dir,
+        &layer_toml_path,
         r#"
 [types]
 launch = true
--- a/libcnb/src/layer/test.rs
+++ b/libcnb/src/layer/test.rs
@@ -388,7 +388,16 @@ fn create_then_keep() {
     )
     .unwrap();
 
-    let handle_layer_result = handle_layer(&context, layer_name.clone(), test_layer).unwrap();
+    let handle_layer_result = handle_layer(
+        &context,
+        layer_name.clone(),
+        TestLayer {
+            // Since we want to keep an existing layer as-is, this version must not be written
+            write_version: String::from("UNUSED_BECAUSE_OF_EXISTING_LAYER_STRATEGY_KEEP"),
+            ..test_layer
+        },
+    )
+    .unwrap();
 
     // Assert layer content metadata, returned and on disk
     let expected_content_layer_metadata = LayerContentMetadata {
--- a/libcnb/src/layer/test.rs
+++ b/libcnb/src/layer/test.rs
@@ -682,11 +682,6 @@ fn error_handling_no_directory() {
     fs::write(
         &layer_toml_path,
         r#"
-[types]
-launch = true
-build = true
-cache = true
-
 [metadata]
 version = "3.2.1"
     "#,
--- a/libcnb/src/layer/test.rs
+++ b/libcnb/src/layer/test.rs
@@ -826,6 +826,77 @@ fn default_layer_method_implementations() {
     assert_eq!(update_result.metadata, simple_layer_metadata);
 }
 
+#[test]
+fn layer_env_read_write() {
+    #[derive(Clone)]
+    struct LayerDataTestLayer {
+        expected_layer_env: LayerEnv,
+    }
+
+    impl Layer for LayerDataTestLayer {
+        type Buildpack = TestBuildpack;
+        type Metadata = GenericMetadata;
+
+        fn types(&self) -> LayerTypes {
+            LayerTypes {
+                launch: true,
+                build: true,
+                cache: true,
+            }
+        }
+
+        fn create(
+            &self,
+            _context: &BuildContext<Self::Buildpack>,
+            _layer_path: &Path,
+        ) -> Result<LayerResult<Self::Metadata>, <Self::Buildpack as Buildpack>::Error> {
+            LayerResultBuilder::new(GenericMetadata::default())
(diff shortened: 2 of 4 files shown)


libcnb/src/layer/test.rs
```diff
@@ -0,0 +1,792 @@
...
+const TEST_LAYER_UPDATE_FILE_NAME: &str = "update";
+
+#[derive(Clone)]
+struct TestLayer {
+    existing_layer_strategy: ExistingLayerStrategy,
+    write_layer_env: Option<LayerEnv>,
+    write_version: String,
+}
+
+impl Layer for TestLayer {
```
edmorley, line 50: Do we need a few more tests that also test a `Layer` that only has implementations for the mandatory methods, so we can test the default implementations for these?
- `existing_layer_strategy()`
- `update()`
- `migrate_incompatible_metadata()`

edmorley (approved):

libcnb/src/layer/handling.rs
```diff
@@ -107,6 +107,10 @@ fn handle_create_layer<B: Buildpack + ?Sized, L: Layer<Buildpack = B>>(
 ) -> Result<LayerData<L::Metadata>, HandleLayerErrorOrBuildpackError<B::Error>> {
     let layer_dir = context.layers_dir.join(layer_name.as_str());
 
+    fs::create_dir_all(&layer_dir)
```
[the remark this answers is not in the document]
edmorley: Actually, this is `layer_dir`, not `layers_dir`, so I don't know how this worked before for the Ruby example.

libcnb/src/layer/test.rs
```diff
@@ -0,0 +1,792 @@
...
+    assert_eq!(update_file_contents, None);
+}
+
+#[test]
+fn error_handling_no_directory() {
+    let temp_dir = tempdir().unwrap();
+    let context = build_context(&temp_dir);
+    let layer_name = random_layer_name();
+
+    let layer_dir = temp_dir
```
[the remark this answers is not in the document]
Malax: Thank you, that's correct! Fixed in 2f7a139

libcnb/src/layer/test.rs
Malax: Hard to argue against it when we had an actual defect previously :) I'll see how these can be implement without too much boilerplate. Ideally, we just test the specific implementations and not end-to-end.

libcnb/src/layer/test.rs
```diff
@@ -0,0 +1,792 @@
...
+        existing_layer_strategy: ExistingLayerStrategy::Recreate,
+        write_version: metadata_version_string.clone(),
+        write_layer_env: None,
+    };
+
+    let handle_layer_result =
+        handle_layer(&context, layer_name.clone(), test_layer.clone()).unwrap();
+
+    // Add a random file to the layer directory between handle_layer calls to assess if the layer is
+    // actually recreated without any residue left in the directory
```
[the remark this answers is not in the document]
Malax: Good catch, added in e09bcda

libcnb/src/layer/test.rs
```diff
@@ -0,0 +1,792 @@
...
+        existing_layer_strategy: ExistingLayerStrategy::Keep,
+        write_version: metadata_version_string.clone(),
+        write_layer_env: None,
+    };
+
+    let handle_layer_result =
+        handle_layer(&context, layer_name.clone(), test_layer.clone()).unwrap();
+
+    // Add a random file to the layer directory between handle_layer calls to assess if the layer is
+    // kept as-is.
```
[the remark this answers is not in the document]
Malax: Good catch! Fixed in 84cfc6d

libcnb/src/layer/test.rs
Malax: Added in 5599990

libcnb/src/layer/test.rs
```diff
@@ -0,0 +1,898 @@
...
+    assert_eq!(
+        create_file_contents,
+        Some(String::from(TEST_LAYER_CREATE_FILE_CONTENTS))
+    );
+
+    assert_eq!(update_file_contents, None);
+}
+
+#[test]
+fn write_layer_env() {
```
edmorley, line 720: This test tests that the env is written to disk, but do we also need a test confirming that a previously written env is read back in successfully on next `handle_layer()` (ie check the `LayerEnv::read_from_layer_dir()` codepath inside `read_layer()`)? 

As far as I can see none of the other tests cover this case so far.

libcnb/src/layer/test.rs
```diff
@@ -0,0 +1,898 @@
...
+        .join("layers")
+        .join(format!("{}.toml", layer_name.as_str()));
+
+    fs::write(
+        &layer_toml_path,
+        r#"
+[types]
+launch = true
+build = true
+cache = true
```
edmorley, line 688: I suppose technically this `types` table wouldn't exist here, since the restored layer would have had that stripped by the lifecycle?

libcnb/src/layer/handling.rs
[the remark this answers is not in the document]
edmorley: @Malax Ah so it looks like this was indeed broken for buildpacks before (I've confirmed via adding an `.exists()` check to one of the examples), and the examples only worked because:
1. The hello world example doesn't write to the layer dir.
2. For the Ruby example:
   - the Ruby layer's first write to the `<layers>/ruby` layer dir is via the `tar` crate's `unpack()`, which [creates the directory](https://docs.rs/tar/0.4.37/src/tar/archive.rs.html#163-166).
   - the bundler layer's first write to the `<layers>/bundler` layer dir is via `bundle install --path <layer_dir>` which [creates the directory](https://github.com/rubygems/rubygems/blob/v3.2.33/bundler/lib/bundler/installer.rb#L69) if it doesn't exist. (The `gem install` command run prior to `bundle install` installs bundler into `<layers>/ruby` instead, so doesn't touch `<layers>/bundler`.)

So this fix is needed for more than just the tests :-)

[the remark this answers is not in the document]
Malax: Thanks for looking into this before I could :) Yeah, I was puzzled too when I had to add this and thought that I'd investigate later why it works with the example buildpacks. 👍🏻

libcnb/src/layer/test.rs
Malax: It currently doesn't matter for this test, but you're totally right! I removed the types in 224f3d4.

libcnb/src/layer/test.rs
Malax: Very good point!

Malax: Added a test in aaf730d

libcnb/src/layer/test.rs
edmorley: Thank you :-) I agree it doesn't affect the test, but it both cleans them up and means more representative of actual input.

It can also be removed from `update_with_incompatible_metadata_recreate()` and `update_with_incompatible_metadata_replace()`, but not important if you'd rather not :-)

edmorley (approved):

PR #223, opened 2021-12-07
Allow process comparisons

schneems:
As seen in the procfile buildpack work. I want to be able to compare equality of Process variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Add `PartialEq` and `Eq` implementations for `Process`.
 - Add support for `default` key in `launch.toml` in `Process` struct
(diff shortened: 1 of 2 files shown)


Malax (approved):

edmorley:
Should we add `Eq` / `PartialEq` to more of the structs in libcnb-data?

Malax:
@edmorley 
> Should we add `Eq` / `PartialEq` to more of the structs in libcnb-data?

Probably, yeah. It's a little bit more complicated with structs that take generic metadata since we cannot derive due to the fact that the generic type might implement `PartialEq` itself. We would need smth. like this:

```rust
impl<T: PartialEq> PartialEq for SomeType<T> {
    // Implement me!
}
```

PR #229, opened 2021-12-08
Make `libcnb_data::build_plan::Require` fields public

### LINKED ISSUE #197: libcnb::data::build_plan::Require metadata field is private and cannot be set via a method
```
error[E0451]: field `metadata` of struct `Require` is private
  --> src/main.rs:32:54
   |
32 |                   Require { name: String::from("pip"), metadata:
   |  ______________________________________________________^
33 | |                     Map::from_iter([(String::from("build"), toml::Value::Boolean(true))])
   | |_________________________________________________________________________________________^ private field
```

Malax:
This change makes the `Require` fields public so that the current API can be used properly when a buildpack's requirements need metadata. We need to look into an overhaul of the build plan API, but this is out of scope for this PR.

Fixes #197

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,6 @@
 - Fixed file extension for delimiters when writing `LayerEnv` to disk.
 - Fixed the `group` field on `buildpack::Order` to now be public.
 - Add an external Cargo command for packaging libcnb buildpacks. See the README for usage.
-- libcnb_data::build_plan::Require fields are now public.
+- `libcnb_data::build_plan::Require` fields are now public.
 
 ## [0.3.0] 2021/09/17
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -52,5 +52,6 @@
 - Fixed file extension for delimiters when writing `LayerEnv` to disk.
 - Fixed the `group` field on `buildpack::Order` to now be public.
 - Add an external Cargo command for packaging libcnb buildpacks. See the README for usage.
+- libcnb_data::build_plan::Require fields are now public.
```
edmorley, line 55:
```suggestion
- `libcnb_data::build_plan::Require` fields are now public.
```

edmorley (approved):

Issue #146, opened 2021-11-12
Incorrect data model for LayerContentMetadata


Malax:
When deserialising `LayerContentMetadata` values, if the `types` are missing from it (due to the change in the CNB 0.6 spec), its default value is used. This is problematic as it does not reflect the state accurately and downstream code cannot discern if the default value was used.

The spec is not explicit about layer content metadata being valid without types, but given the reference lifecycle implementation writes layer content metadata files without types, we should model our type accordingly. Especially since we have to read those from libcnb.

For reference, this is how LayerContentMetadata looks like right now:

```rust
#[derive(Debug, Deserialize, Serialize)]
pub struct LayerContentMetadata<M> {
    #[serde(default)]
    pub types: LayerTypes,

    /// Metadata that describes the layer contents.
    pub metadata: M,
}
```

Changing the type of `types` to `Option<LayerTypes>` would get us what we want. Unfortunately, in most cases we actually want to work with `LayerContentMetadata` that has `types` populated. We need to either change  all call-sites to be able to work with `LayerContentMetadata` that does not have `types`, or we introduce a second, stricter type, that can be created from the weaker version.

PR #235, opened 2021-12-10
Remove builder-style methods from `LayerContentMetadata`

### LINKED ISSUE #234: Remove builder methods for `LayerContentMetadata`
Currently `LayerContentMetadata` has a number of builder-style methods:
https://github.com/Malax/libcnb.rs/blob/ca78c911d2351c68a3386b7c029cc44c0aae24ca/libcnb-data/src/layer_content_metadata.rs#L67-L94

With the new `Layer` API, buildpacks no longer construct `LayerContentMetadata` themselves, and libcnb itself doesn't use these methods.

I think we should remove them, since:
- On the most part `LayerContentMetadata` is going to be a type that is only consumed rather than constructed by end-users.
- These methods are a bit of a mess/inconsistent
- They'll require a number of changes for #146

edmorley:
Since with the new `Layer` API (#144), buildpacks no longer need to construct values with this type, and so don't need to use these builders-style methods.

In addition these methods:
- Were inconsistent and needed cleanup.
- Would be another thing to update for #146.

Fixes #234.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Add `must_use` attributes to a number of pure public methods.
+- Remove builder-style methods from `LayerContentMetadata`.
 
 ## [0.4.0] 2021/12/08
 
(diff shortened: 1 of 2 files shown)


Malax (approved):
I'm fine with removing these, I couldn't come up with a good use-case after the changes.

PR #236, opened 2021-12-10
Correct the data model for `LayerContentMetadata`

### LINKED ISSUE #146: Incorrect data model for LayerContentMetadata

edmorley:
The `types` field of `LayerContentMetadata` is now of type `Option<LayerTypes>` rather than `LayerType`.

This better models the fact that the `[types]` table can be omitted entirely from the layer TOML file, and in fact the table is intentionally removed by the lifecycle when layers are restored in Buildpack API 0.6+.

In addition, `LayerContentMetadata::Default()` has been removed, since:
- It's no longer needed for `#[serde(default)]`.
- It's not really clear what the default for `types` should be (`None` or `LayerTypes::Default()`), and it's probably better if consumers of `LayerContentMetadata` make that choice explicitly.

Lastly, more tests have been added for `LayerContentMetadata`.

Fixes #146.
GUS-W-10289194.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Add `must_use` attributes to a number of pure public methods.
 - Remove builder-style methods from `LayerContentMetadata`.
+- Make `LayerContentMetadata`'s `types` field an `Option`.
+- Remove `LayerContentMetadata::Default()`.
 
 ## [0.4.0] 2021/12/08
 
--- a/libcnb-data/src/layer_content_metadata.rs
+++ b/libcnb-data/src/layer_content_metadata.rs
@@ -1,5 +1,22 @@
 use serde::{Deserialize, Serialize};
 
+/// Describes Layer Content Metadata
+///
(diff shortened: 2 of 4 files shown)


Malax (approved):
That's a simpler solution than I've expected! 👍🏻

PR #237, opened 2021-12-13
Link to PRs from changelog entries

edmorley:
When consuming other project's changelogs, I find it useful to be able to look at the PR related to a changelog entry, when the changelog entry is not clear, or I want to know more about the motivation/exactly what changed.

This adds PR links for anything newer than v0.4.0 (I don't think it's worth the effort of back-filling previous entries; there was more churn in the project earlier on + fewer users etc).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,10 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


hone:
I agree. I really like when things are linked.

PR #241, opened 2021-12-16
Fix the data model and validation of `Buildpack.version`

### LINKED ISSUE #239: Buildpack `version` validation permits values not allowed by the spec
The version of a buildpack is specified by the `version` field in `buildpack.toml` like so:

```toml
api = "0.6"

[buildpack]
id = "foo/bar"
version = "0.0.1"

[[stacks]]
id = "*"
```

The CNB spec says:

> The buildpack version:
>    MUST be in the form `<X>.<Y>.<Z>` where `X`, `Y`, and `Z` are non-negative integers and must not contain leading zeros.
>     -  Each element MUST increase numerically.
>     -  Buildpack authors will define what changes will increment `X`, `Y`, and `Z`.

However currently the version is implemented using the [semver](https://crates.io/crates/semver) crate:
https://github.com/Malax/libcnb.rs/blob/6cba43aa9fa70fb28b5ef5ff126987df5f3fad95/libcnb-data/src/buildpack/mod.rs#L61

...which permits any allowed semver version - including those that have `-foo` and `+foo` suffixes (which are not allowed in buildpack versions):

```
<valid semver> ::= <version core>
                 | <version core> "-" <pre-release>
                 | <version core> "+" <build>
                 | <version core> "-" <pre-release> "+" <build>

<version core> ::= <major> "." <minor> "." <patch>
```

(from https://semver.org)

We should switch to our own parsing/validation to ensure only versions of form `X.Y.Z` are permitted.

This came up in a recent JVM buildpack PR, where a buildpack was given an invalid version but was only found when run with pack in CI.

edmorley:
Previously it was a `semver::Version`, which:
- Allowed for version values not permitted by the CNB spec, such as those with a prerelease (eg: `-foo`) or build metadata (eg `+abc`) suffix.
- Had `prerelease` and `build` fields on the struct, which are not part of the allowed version scheme.

Version parsing/validation is now performed via a new `BuildpackVersion` type, which follows a similar style to that of `BuildpackApi`. However the implementation of  `BuildpackVersion::try_from` has had to use regex, since buildpack versions cannot contain leading zeros (unlike the buildpack API version), which would be silently accepted by `str::parse()` if we instead used the simpler `.split(".")`approach.

As an added bonus of this fix, we get to drop a dependency :-)

Fixes  #239.
GUS-W-10321651.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Remove builder-style methods from `LayerContentMetadata` ([#235](https://github.com/Malax/libcnb.rs/pull/235)).
 - Make `LayerContentMetadata`'s `types` field an `Option` ([#236](https://github.com/Malax/libcnb.rs/pull/236)).
 - Remove `LayerContentMetadata::Default()` ([#236](https://github.com/Malax/libcnb.rs/pull/236)).
+- Switch `Buildpack.version` from type `semver::Version` to `BuildpackVersion`, in order to validate versions more strictly against the CNB spec ([#241](https://github.com/Malax/libcnb.rs/pull/241)).
 
 ## [0.4.0] 2021/12/08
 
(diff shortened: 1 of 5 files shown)


Malax (approved):
👍🏻

PR #243, opened 2021-12-17
Launch struct, add serde default value for field

lead4good:


lead4good:
Regarding Process the spec states:

> For each process, the buildpack:
> 
>     MUST specify a type, which:
>         MUST NOT be identical to other process types provided by the same buildpack.
>         MUST only contain numbers, letters, and the characters ., _, and -.
>     MUST specify a command that is either:
>         A command sequence that is valid when executed using the shell, if args is not specified.
>         A path to an executable or the file name of an executable in $PATH, if args is a list with zero or more elements.
>     MAY specify an args list to be passed directly to the specified executable.
>     MAY specify a direct boolean that bypasses the shell.
>     MAY specify a default boolean that indicates that the process type should be selected as the buildpack-provided default during the export phase.

For proper deserialization we need to therefor wrap args, direct and default in an Option.

libcnb-data/src/launch.rs
```diff
@@ -62,30 +62,47 @@ pub struct Label {
...
+        if process.args.is_some() {
+            self.args = process.args;
+        }
+        if process.direct.is_some() {
+            self.direct = process.direct;
+        }
+        if process.default.is_some() {
+            self.default = process.default;
+        }
+    }
```
schneems, line 102: What's the motivation behind the update function? I don't see it being used anywhere?

schneems:
This will be a change to any buildpack using the Process struct directly. I think that making those *may* fields an Option is more technically accurate. I think we've talked about it before. I'm fine with that change. It also seems reasonable to expect to be able to parse through arbitrary `launch.toml` files. I asked one question around the `update` function. Also it looks like your listing is off, can you check that?

@Malax I would love to hear your thoughts here.

libcnb-data/src/launch.rs
lead4good: The spec mentions a special case regarding processes from more than one buildpack with the same process type. Say a process with a certain process type already exists, if one or more of the _may_ fields of subsequent definitions are not defined, the _may_fields of the previous process definition will be taken - hence the update function.

lead4good:
> Also it looks like your listing is off, can you check that?

I'm not sure what you mean by that, can you elaborate?

edmorley:
@lead4good Hi! Could you explain more about the use-case? 

Parsing previous buildpack's layers sounds a bit like an anti-pattern, since it presumably means hardcoding layer paths or traversing all previous layers, and doing so manually since iirc the CNB API doesn't provide process types defined in previous buildpacks. If this is a common use-case we'd be better off getting the API changed upstream to surface previously defined processes in another way IMO.

lead4good:
@edmorley 
Sure! I'm developing a buildpack for gramine, which provides a LibOS, basically wraps around your existing application to make it run in a constrained (SGX) environment. So I want to develop this buildpack in such a way that it is agnostic to the previous layers, it just needs to know the processes defined by the previous layers and then it adapts the process command in such way that gramine is wrapped around it. You are right, CNB API doesn't provide process types defined in previous buildpacks, which is why I had to manually parse the layer directories for the launch.toml's and collect the processes as defined by the spec. The update function is part of this code, so for me it is perfectly fine to remove that function from this PR and keep it as part of the buildpack itself.

lead4good:
Rebased

Malax (approved):
As much as this looks much more verbose, it's correct. 👍🏻  Let's tackle the verbosity issues around `Process` separately - it wasn't great before either.

I think we can merge this as soon as we have the missing CHANGELOG entry and get it into the next release (which should be this or next week).

Thank you @lead4good for your contribution!

libcnb-data/src/launch.rs
```diff
@@ -60,29 +60,31 @@ pub struct Label {
...
+#[derive(Deserialize, Serialize, Debug, Clone, PartialEq, Eq)]
 #[serde(deny_unknown_fields)]
 pub struct Process {
     pub r#type: ProcessType,
     pub command: String,
-    pub args: Vec<String>,
-    pub direct: bool,
-    #[serde(default)]
-    pub default: bool,
+    #[serde(skip_serializing_if = "Option::is_none")]
```
edmorley, line 68: What does serde output if the `skip_serializing_if = "Option::is_none"` isn't used?

At a guess, I would have thought `None` would serialize to nothing by default, and thus the `skip_serializing_if` here might be redundant?

Malax: FWIW, the docs mention this case explicitly: https://serde.rs/field-attrs.html#skip_serializing_if. I haven't tested this, but I assume `None` will become `null`.

libcnb-data/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - All libcnb-data struct types now reject unrecognised fields when deserializing ([#252](https://github.com/Malax/libcnb.rs/pull/252)).
 - `BuildpackToml` has been replaced by `BuildpackDescriptor`, which is an enum with `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types. The new types now reject `buildpack.toml` files where both `stacks` and `order` are present ([#248](https://github.com/Malax/libcnb.rs/pull/248)).
 - Implement `Borrow<str>` for types generated using the `libcnb_newtype!` macro (currently `BuildpackId`, `LayerName`, `ProcessType` and `StackId`), which allows them to be used with `.join()` ([#258](https://github.com/Malax/libcnb.rs/pull/258)).
+- `Process::args`, `Process::direct`, `Process::default` are now wrapped in `Option` in accordance to the CNB spec.
```
edmorley, line 13:
```suggestion
- `Process::args`, `Process::direct`, `Process::default` are now wrapped in `Option` in accordance with the CNB spec ([#243](https://github.com/Malax/libcnb.rs/pull/243)).
```

Issue #168, opened 2021-11-18
Naming consistency: `BuildpackToml` and `buildpack_descriptor`


Malax:
We currently have a `BuildpackToml` type that was renamed from `BuildpackDescriptor` some time ago. We also have a `buildpack_descriptor` field in the `DetectContext` and `BuildContext` that returns that type.

We should have consistent naming for it. We should either rename the type back to `BuildpackDescriptor` or change the field name to `buildpack_toml`. 

@schneems made a comment about the fact that it wasnt obvious for him that `buildpack_descriptor` is the `buildpack.toml` file. (https://github.com/Malax/libcnb.rs/pull/144#discussion_r751516238)

edmorley:
There are so many overloaded terms in the CNB spec, plus many things without an explicit name - it's quite frustrating.

For example we can't really rename `BuildpackToml` to `BuildpackMetadata`, since within the file there is also a `metadata` table. And similarly, we can't rename it instead to `Buildpack`, because there's a nested `buildpack` table too.

I see "Buildpack descriptor" is a term referenced in the docs:
https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml

...though I agree it's not immediately intuitive.

One option, would be to rename `BuildpackToml` to `Buildpack`, and have it accessible from the context via `buildpack`, and then to avoid the nested `buildpack` key with the same name, flatten those keys into the top-level during deserialisation, giving something like:

```
context.buildpack.api
context.buildpack.name
context.buildpack.stacks
```

PR #248, opened 2022-01-05
Correct the data model for `buildpack.toml`

### LINKED ISSUE #211: `libcnb_data::buildpack::Buildpack` should not permit meta-buildpacks that define `stacks`

edmorley:
The CNB spec states that a `buildpack.toml` must contain exactly one of the tables `stacks` or `order`, but not both - since they relate to different buildpack types (single buildpacks and meta-buildpacks respectively):
https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpack-implementations

Previously `BuildpackToml` attempted to represent both of these as a single type, which both meant the type wasn't an accurate representation, but also that invalid `buildpack.toml` files would pass `libcnb package` validation only to fail later at runtime with a Pack CLI error message.

Now `BuildpackToml` has been replaced with the enum `BuildpackDescriptor`, which has `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types.

Consumers of libcnb-data can now either parse `buildpack.toml` files using `BuildpackDescriptor` when they wish to support both variants, or else use the specific sub-types when a specific buildpack type is required for improved error messages.

Fixes #211.
GUS-W-10289129.

### DIFF
--- a/libcnb-data/src/buildpack/mod.rs
+++ b/libcnb-data/src/buildpack/mod.rs
@@ -12,11 +12,17 @@ pub use version::*;
 
 use serde::Deserialize;
 
-/// Data structure for the Buildpack descriptor (buildpack.toml).
+/// Data structures for the Buildpack descriptor (buildpack.toml).
 ///
-/// # Examples
+/// For parsing of [buildpack.toml](https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml)
+/// files when support for multiple types of buildpack is required.
+///
+/// When a specific buildpack type is expected, use [`SingleBuildpackDescriptor`] or [`MetaBuildpackDescriptor`] directly instead,
+/// since it allows for more detailed error messages if parsing fails.
+///
+/// # Example:
 /// ```
-/// use libcnb_data::buildpack::BuildpackToml;
+/// use libcnb_data::buildpack::BuildpackDescriptor;
 ///
 /// let toml_str = r#"
 /// api = "0.6"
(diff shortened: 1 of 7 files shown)


edmorley:
CI is failing due to the unrelated #249, which will be fixed by #250 (I'll rebase once that merges).

libcnb-data/src/buildpack/mod.rs
```diff
@@ -34,22 +40,111 @@ use serde::Deserialize;
...
+/// match buildpack_toml {
+///     BuildpackToml::Single(buildpack) => println!("Found buildpack: {}", buildpack.buildpack.id),
+///     BuildpackToml::Meta(buildpack) => println!("Found meta-buildpack: {}", buildpack.buildpack.id),
+/// };
+/// ```
+#[derive(Deserialize, Debug)]
+#[serde(untagged)]
+pub enum BuildpackToml<BM> {
+    Single(SingleBuildpack<BM>),
+    Meta(MetaBuildpack<BM>),
```
Malax, line 56: I'm not yet sold on the naming of these types. I feel like this PR is a good opportunity to resolve #168, or at least finalize a naming for the types so we can change the variable names separately to match the type names.

Since at least `SingleBuildpack` is used standalone in buildpack implementations, I don't think it really conveys what it is. Granted, naming CNB related things is hard with all the overloaded terms in the spec. When I first read this in a context where `SingleBuildpack` is used standalone (later in this PR), I stopped for a moment, thinking "why does this function get a buildpack as an argument?".

If we'd want to keep the `BuildpackToml` naming for the composite type, I could see us using `SingleBuildpackToml` and `MetaBuildpackToml` for the variant structs.

In general, I think I don't like the `BuildpackToml` name so much and would prefer `BuildpackDescriptor` instead. `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` work fine for the variants IMO.

edmorley: Yeah I agree all of the names are suboptimal, and that the CNB spec doesn't make this easy.

I like the idea of `BuildpackDescriptor` / `SingleBuildpackDescriptor` / `MetaBuildpackDescriptor`.

The other type name that bugs me, is `libcnb_data::buildpack::Buildpack`, which is the `buildpack` table inside `buildpacks.toml`. My first thought was that we could name it `BuildpackMetadata`, but then we also have the `metadata` table which sounds too similar to that. Other options for that might be `BuildpackTable`, `BuildpackProperties` (but then what's the difference between "properties" and "metadata"?), `BuildpackDetails` etc.

I guess really `metadata` is "custom metadata" and the `buildpack` table is `BuildpackMetadata`?

Eugh

hone: I think I named it `BuildpackToml` so you can blame me for that. I agree that `BuildpackDescriptor` is the better name.

libcnb-data/src/buildpack/mod.rs
```diff
@@ -34,22 +40,111 @@ use serde::Deserialize;
...
+/// let buildpack_toml = toml::from_str::<SingleBuildpack<Option<toml::value::Table>>>(toml_str).unwrap();
+/// assert_eq!(buildpack_toml.buildpack.id, buildpack_id!("foo/bar"));
+/// assert_eq!(buildpack_toml.stacks, vec![Stack::Any]);
 /// ```
 #[derive(Deserialize, Debug)]
-pub struct BuildpackToml<BM> {
+// `deny_unknown_fields` is required since otherwise a buildpack TOML with both
+// `stacks` and `order` will be deserialized into the first type that matches,
+// rather than being rejected for being invalid.
+#[serde(deny_unknown_fields)]
```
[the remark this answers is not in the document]
hone: I think we should, afaik all the field should be spec'd except for things like `[metadata]` tables.

hone:
Besides the naming bikeshed this implementation looks good.

libcnb-data/src/buildpack/mod.rs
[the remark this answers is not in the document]
edmorley: Yeah I agree

libcnb-data/src/buildpack/mod.rs
Malax: `libcnb_data::buildpack::Buildpack` is something for another PR I guess?

edmorley: Types renamed in ed15a71.

libcnb-data/src/buildpack/mod.rs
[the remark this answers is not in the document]
edmorley: Filed #251

Malax (approved):
Thank you! :shipit:

PR #252, opened 2022-01-06
Reject unknown fields when deserializing TOML files

### LINKED ISSUE #251: Add `#[serde(deny_unknown_fields)]` to more types
Serde provides the ability for unknown fields to be rejected, by annotating types with:
```rust
#[serde(deny_unknown_fields)]
```

We added this to a few types as part of #248, however there are other places that could benefit from it too.

See:
https://serde.rs/container-attrs.html#deny_unknown_fields

GUS-W-10393275.

edmorley:
The Serde attribute `deny_unknown_fields` ensures that any unknown fields cause the parsing of TOML files to fail, making it easier to spot when the file does not match the expected schema.

This will ensure issues like heroku/buildpacks-jvm#227 are spotted sooner.

See:
https://serde.rs/container-attrs.html#deny_unknown_fields

Closes #251.
GUS-W-10393275.


hone (approved):

Malax (approved):
Glad to see that the tests didn't require fixes after adding `deny_unknown_fields`! :)

edmorley:
> Glad to see that the tests didn't require fixes after adding `deny_unknown_fields`! :)

Yes me too! 😆

PR #256, opened 2022-01-10
Switch to separate changelogs per crate

### LINKED ISSUE #255: Switch to separate changelogs per crate
Each crate in this monorepo has a separate crate version, so it's not accurate to lump all changes under the libcnb version currently listed in the repo root `CHANGELOG.md`.

We should split up that `CHANGELOG.md` into separate files, one inside each crate directory.

When it comes to changes to bumping versions, we'll also want to add "Update to libcnb-data X.Y.Z" etc in the main libcnb changelog, and potentially add some further detail for changes that impact end-users of libcnb to save them having to look up the dependency changes themselves.

edmorley:
Since each crate has an independent (and not synced) version, so a single unified changelog is no longer a good fit.

Minimal changelog entries have also been added for historic releases that predated the creation of the original changelog, for completeness.

The check changelog check still works, since it doesn't require the changelogs to be in the repo root.

Closes #255.
GUS-W-10412360.


Malax (approved):

Malax:
What should we do for libcnb-data changes? They will affect users of `libcnb` too and people might not know that there are two crates and changelogs that are important for them.

edmorley:
@Malax For each release, when we bump the versions in `Cargo.toml` we should add a corresponding entry to `libcnb` saying "Updating libcnb-data to X.Y.Z (link to changes)" -- like I did for this PR, for the libcnb 0.4.0 release.

PR #257, opened 2022-01-11
Rename `TargetLifecycle` to `Scope`

### LINKED ISSUE #170: Rename `TargetLifecycle`
Discussed in https://github.com/Malax/libcnb.rs/pull/144/files#r751525550

https://github.com/Malax/libcnb.rs/blob/b2ba884a37dd58ce010eb01a66748cbfe5bd6c20/libcnb/src/layer_env.rs#L424-L431

Maybe we want to consider renaming this. To me the lifecycle is the thing we're currently in (the layer/lifecycle). So I was reading this code and thinking "this env will apply to all layers" when what this really means is "this env applies to all layer types (build, cache, launch)". 

Above we call the struct `LayerTypes` so perhaps `TargetLayerTypes::All`. The "target" term is also a little ambiguous. Maybe `ApplyToLayerType::All` or `OnLayerType::All`. WDYT?

edmorley:
To make its purpose clearer.

Fixes #170.
GUS-W-10415008.

### DIFF
--- a/examples/example-02-ruby-sample/src/layers/ruby.rs
+++ b/examples/example-02-ruby-sample/src/layers/ruby.rs
@@ -9,7 +9,7 @@ use libcnb::build::BuildContext;
 use libcnb::data::layer_content_metadata::LayerTypes;
 use libcnb::generic::GenericMetadata;
 use libcnb::layer::{Layer, LayerResult, LayerResultBuilder};
-use libcnb::layer_env::{LayerEnv, ModificationBehavior, TargetLifecycle};
+use libcnb::layer_env::{LayerEnv, ModificationBehavior, Scope};
 
 pub struct RubyLayer;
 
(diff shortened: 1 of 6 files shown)


Malax (commented):
Welcome to yet another bikeshed discussion. :D 

I like the "scope" wording at lot! @hone was complaining about `TargetLifecycle` to me yesterday too. In our discussion he made the point that it could just be `Target` as the module name provides enough context in addition to the type name itself. I think I agree.

That said, I think `Scope` is an even better name and @hone thinks so too. WDY think about just naming it `Scope`?

edmorley:
Ah yeah I like `Scope` even better here, given that this enum is only used by/with `LayerEnv` so I agree adding a prefix is redundant.

Malax (approved):
I like it even more when actually seeing the change in front of me. :)

edmorley:
Yeah same :-)

PR #258, opened 2022-01-11
Implement `Borrow<str>` for types generated using `libcnb_newtype!`

### LINKED ISSUE #182: Unable to `.join()` newtype macro types
A `Vec<String>` (which is what `mixins` currently is) can be joined like so:
https://github.com/Malax/libcnb.rs/blob/2f57cd18beff95a52fd85229823e0c8dc05892ce/libcnb-data/src/buildpack.rs#L102

However if the `Vec<String>` is converted to `Vec<StackMixin>`, with `StackMixin` having been created using the `libcnb_newtype` macro, a trait bounds not satisfied error is generated when `.join()` is used.

For example:

```rust
    let mixins = vec![stack_mixin!("a"), stack_mixin!("b")];
    println!("mixins: {}", mixins.join(", "));
```

```
error[E0599]: the method `join` exists for struct `std::vec::Vec<buildpack::StackMixin>`, but its trait bounds were not satisfied
   --> libcnb-data/src/buildpack.rs:319:35
    |
319 |     println!("mixins: {}", mixins.join(", "));
    |                                   ^^^^ method cannot be called on `std::vec::Vec<buildpack::StackMixin>` due to unsatisfied trait bounds
    |
    = note: the following trait bounds were not satisfied:
            `<[buildpack::StackMixin] as std::slice::Join<_>>::Output = _`

For more information about this error, try `rustc --explain E0599`.
```

I presume we're missing a trait implementation on `libcnb_newtype` that would allow this?

edmorley:
This allows the types to be used with `.join()`.

Fixes #182.
GUS-W-10415269.

### DIFF
--- a/libcnb-data/CHANGELOG.md
+++ b/libcnb-data/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Switch `Buildpack.version` from type `semver::Version` to `BuildpackVersion`, in order to validate versions more strictly against the CNB spec ([#241](https://github.com/Malax/libcnb.rs/pull/241)).
 - All libcnb-data struct types now reject unrecognised fields when deserializing ([#252](https://github.com/Malax/libcnb.rs/pull/252)).
 - `BuildpackToml` has been replaced by `BuildpackDescriptor`, which is an enum with `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types. The new types now reject `buildpack.toml` files where both `stacks` and `order` are present ([#248](https://github.com/Malax/libcnb.rs/pull/248)).
+- Implement `Borrow<str>` for types generated using the `libcnb_newtype!` macro (currently `BuildpackId`, `LayerName`, `ProcessType` and `StackId`), which allows them to be used with `.join()` ([#258](https://github.com/Malax/libcnb.rs/pull/258)).
 
 ## [0.3.0] 2021-12-08
 
--- a/libcnb-data/src/newtypes.rs
+++ b/libcnb-data/src/newtypes.rs
@@ -103,6 +103,12 @@ macro_rules! libcnb_newtype {
             }
         }
 
+        impl ::std::borrow::Borrow<str> for $name {
+            fn borrow(&self) -> &str {
+                &self.0
+            }
+        }
+
         impl ::std::ops::Deref for $name {
             type Target = String;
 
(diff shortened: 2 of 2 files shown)


libcnb-data/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Switch `Buildpack.version` from type `semver::Version` to `BuildpackVersion`, in order to validate versions more strictly against the CNB spec ([#241](https://github.com/Malax/libcnb.rs/pull/241)).
 - All libcnb-data struct types now reject unrecognised fields when deserializing ([#252](https://github.com/Malax/libcnb.rs/pull/252)).
 - `BuildpackToml` has been replaced by `BuildpackDescriptor`, which is an enum with `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types. The new types now reject `buildpack.toml` files where both `stacks` and `order` are present ([#248](https://github.com/Malax/libcnb.rs/pull/248)).
+- Implement `Borrow<str>` for types generated using the `libcnb_newtype!` macro, which allows them to be used with `.join()` ([#258](https://github.com/Malax/libcnb.rs/pull/258)).
```
Malax, line 12: Is this a good entry for the end-user? They would have no visibility which types are created via that macro, only be reading the source.

Malax: In addition, do we want entry for `libcnb` too since most people use `libcnb-data` implicitly?

Malax (approved):

libcnb-data/CHANGELOG.md
edmorley: > In addition, do we want an entry for the libcnb CHANGELOG too since most people use libcnb-data implicitly?

It will be added later, see:
https://github.com/Malax/libcnb.rs/pull/256#issuecomment-1009829942

edmorley: > Is this a good entry for the end-user? They would have no visibility which types are created via that macro, only be reading the source.

Do you have a suggestion as to what else to put? I don't think we should list every single type.

edmorley: Actually, there are only four such types at the moment (I could have sworn there were more? 😆), so I'll list them (`BuildpackId`, `LayerName`, `ProcessType` and `StackId`).

PR #261, opened 2022-01-11
Test buildpack cross-compilation & packaging in CI

### LINKED ISSUE #259: Test building/packaging/running examples in CI
Whilst we currently lint the files under `examples/` and the `libcnb-cargo` package in CI, we don't actually have any tests that verify that running `cargo libcnb package` against those examples results in a buildpack that works as expected with Pack CLI.

We should add another CI job that:
1. Installs Pack CLI / cross-compiling deps
2. Runs `cargo libcnb package` against each of the example buildpacks
3. Performs a test Pack CLI build with each

Some of this will likely become easier when we have integration testing support for libcnb. Though even then, we may want to test running `cargo libcnb package` in CI, to check it works as a binary, and not just when consumed as a library from the integration testing crate.

edmorley:
Adds a new `integration` GitHub action job under the `CI` workflow, which tests that:
1. `cargo libcnb package` can run successfully with the hello-world buildpack.
2. That the resultant buildpack works when used with `pack build`.

This helps to validate both `libcnb-cargo` and the hello-world example.

The Ruby example buildpack remains without an integration test for now (for a start, it would need the addition of a suitable Ruby test app fixture), however it's something we can test in the future when `libcnb-test` is added, in order to test both `libcnb-test` itself, and library usage of `libcnb-cargo`.

The `cargo build` step has been removed from CI, since it's redundant given that:
(a) `cargo clippy` is a superset of `cargo check`,
(b) the integration job is cross-compiling anyway (which is more representative).

Lastly, branch builds are skipped for all branches apart from `main`, to prevent the duplicate builds on PRs (given that there would now otherwise be 3x2 builds per PR, with the new `integration` job).

Fixes #259.
GUS-W-10418196.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,11 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  # Only perform branch builds for `main`, to avoid duplicate builds on PRs.
+  push:
+    branches:
+      - main
+  pull_request:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,27 +14,50 @@ jobs:
...
+        # Default to a small non-libc image for faster CI + to validate the static musl cross-compilation.
+        # Adjust pull-policy to prevent redundant image-pulling, which slows CI and risks hitting registry rate limits.
+        run: |
+          pack config default-builder cnbs/sample-builder:alpine
+          pack config pull-policy if-not-present
+      - name: Compile and package example-01-basics
+        run: cargo run --package libcnb-cargo -- libcnb package
+        working-directory: ./examples/example-01-basics
+      - name: Pack build using example-01-basics
+        run: pack build example-01-basics --buildpack target/buildpack/debug/libcnb-examples_basics --path examples/
```
Malax, line 63: Do we want to run the image and check for the "Hello World" message?

Malax (approved):

.github/workflows/ci.yml
Malax: Ah, right. The README example does more than the example project. 🤔

PR #263, opened 2022-01-11
Enable caching in CI

### LINKED ISSUE #262: Enable caching in CI
Our CI runs take ~2 minutes, most of which is spent recompiling Rust dependencies from scratch every time.

This project doesn't have a `Cargo.lock` (by design, since it's a library) which makes cache invalidation more complex, however there is a GitHub action available that handles this intelligently for us:
https://github.com/Swatinem/rust-cache

edmorley:
Enables caching using:
https://github.com/Swatinem/rust-cache

This action caches the Cargo index cache, as well as dependencies of libcnb (but nothing else, for cache size efficiency), and automatically handles setting the cache key based on `**/Cargo.toml`, the Rust version and the GitHub job name. See more at:
https://github.com/Swatinem/rust-cache#cache-details

This action is much superior to the manual alternatives I had been investigating prior to finding that action.

This reduces CI end to end time from ~2.5 minutes to ~1 minute.

Fixes #262.
GUS-W-10419238.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,8 @@ jobs:
         # Most of the time this will be a no-op, since GitHub releases new images every week
         # which include the latest stable release of Rust, Rustup, Clippy and rustfmt.
         run: rustup update
+      - name: Rust Cache
+        uses: Swatinem/rust-cache@v1.3.0
       - name: Clippy
         # Using --all-targets so tests are checked and --deny to fail on warnings.
         # Not using --locked here and below since Cargo.lock is in .gitignore.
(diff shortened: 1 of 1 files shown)


Malax (approved):
It took me a moment to understand how this works TBH. Not because of the contents in this PR, just because I didn't realise that every step also has an implicit "post" step that runs at the end of the action. That post step is used to write the cache and the explicit one restores it.

I don't think we need to document this in our repo, just leaving this here should anyone comes across this and wonders the same. :)

edmorley:
Yeah agree it's not obvious from the outset - it's one of the few ways GitHub actions differs quite a bit from Circle CI. I guess longer term when we move this repo under an org, we should consider moving CI here to Circle CI too, just so have one fewer config syntax to remember (excl check-changelog, though we can always move that too if preferred).

PR #265, opened 2022-01-12
ProcessBuilder

Malax:
Adds a `ProcessBuilder` to simplify construction of `Process` values.

In addition, it changes some parts of #243 - `Process` fields are no longer wrapped in `Option` to better match the other types in this crate. This does not change what #243 is fixing: being able to deserialise `Process` values where some optional fields are missing. I also added tests for this.

GUS-W-10422974

### DIFF
--- a/libcnb-data/CHANGELOG.md
+++ b/libcnb-data/CHANGELOG.md
@@ -10,7 +10,8 @@
 - All libcnb-data struct types now reject unrecognised fields when deserializing ([#252](https://github.com/Malax/libcnb.rs/pull/252)).
 - `BuildpackToml` has been replaced by `BuildpackDescriptor`, which is an enum with `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types. The new types now reject `buildpack.toml` files where both `stacks` and `order` are present ([#248](https://github.com/Malax/libcnb.rs/pull/248)).
 - Implement `Borrow<str>` for types generated using the `libcnb_newtype!` macro (currently `BuildpackId`, `LayerName`, `ProcessType` and `StackId`), which allows them to be used with `.join()` ([#258](https://github.com/Malax/libcnb.rs/pull/258)).
-- `Process` can now be deserialized when optional fields are missing. ([#265](https://github.com/Malax/libcnb.rs/pull/265))
+- `Launch` and `Process` can now be deserialized when optional fields are missing, and omit default values when serializing ([#243](https://github.com/Malax/libcnb.rs/pull/243) and [#265](https://github.com/Malax/libcnb.rs/pull/265)).
+- `Process::new` has been replaced by `ProcessBuilder` ([#265](https://github.com/Malax/libcnb.rs/pull/265)).
 
 ## [0.3.0] 2021-12-08
 
--- a/libcnb-data/CHANGELOG.md
+++ b/libcnb-data/CHANGELOG.md
@@ -10,7 +10,7 @@
 - All libcnb-data struct types now reject unrecognised fields when deserializing ([#252](https://github.com/Malax/libcnb.rs/pull/252)).
 - `BuildpackToml` has been replaced by `BuildpackDescriptor`, which is an enum with `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types. The new types now reject `buildpack.toml` files where both `stacks` and `order` are present ([#248](https://github.com/Malax/libcnb.rs/pull/248)).
 - Implement `Borrow<str>` for types generated using the `libcnb_newtype!` macro (currently `BuildpackId`, `LayerName`, `ProcessType` and `StackId`), which allows them to be used with `.join()` ([#258](https://github.com/Malax/libcnb.rs/pull/258)).
-- `Process::args`, `Process::direct`, `Process::default` are now wrapped in `Option` in accordance with the CNB spec ([#243](https://github.com/Malax/libcnb.rs/pull/243)).
+- `Process` can now be deserialized when optional fields are missing. ([#265](https://github.com/Malax/libcnb.rs/pull/265))
 
 ## [0.3.0] 2021-12-08
 
--- a/libcnb-data/src/launch.rs
+++ b/libcnb-data/src/launch.rs
@@ -23,8 +23,9 @@ pub struct Launch {
 /// use libcnb_data::process_type;
 ///
 /// let mut launch_toml = launch::Launch::new();
-/// let web = launch::Process::new(process_type!("web"), "bundle", Some(vec!["exec", "ruby", "app.rb"]),
-/// Some(false), Some(false));
+/// let web = launch::ProcessBuilder::new(process_type!("web"), "bundle")
+///     .args(vec!["exec", "ruby", "app.rb"])
+///     .build();
 ///
 /// launch_toml.processes.push(web);
 /// assert!(toml::to_string(&launch_toml).is_ok());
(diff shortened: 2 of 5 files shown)


libcnb-data/src/launch.rs
```diff
@@ -168,4 +241,52 @@ mod tests {
...
+
+        let string = toml::to_string(&process).unwrap();
+        assert_eq!(
+            string,
+            r#"type = "web"
+command = "foo"
+default = true
+"#
+        );
+    }
```
Malax, line 291: These are tests that should've been added with #243. Since this PR changes the implementation form #243 slightly, I felt obligated to add them with this PR.

libcnb-data/src/launch.rs
```diff
@@ -65,30 +66,102 @@ pub struct Label {
...
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub direct: Option<bool>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub default: Option<bool>,
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub args: Vec<String>,
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub direct: bool,
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub default: bool,
```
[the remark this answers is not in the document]
edmorley: So from reading #243 it sounds like the upstream CNB spec does treat proceses with unset values differently from processes with values set to their default values.

That is, it seems we really do need `Option` in order to represent this?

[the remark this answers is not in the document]
edmorley: Although looking at the lifecycle implementation, it seems that they don't treat the unset case any differently?
https://github.com/buildpacks/lifecycle/blob/6f56a153034317a88f41109b76e026b93134eb89/launch/launch.go#L9-L16

(According to https://pkg.go.dev/encoding/json#Marshal `omitempty` will omit `false` or empty array values)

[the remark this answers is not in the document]
edmorley: > You're referring to #243 (comment), correct?
Yeah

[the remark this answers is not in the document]
hone: I don't think lifecycle treats it differently b/c this is how Go works.

hone (approved):

libcnb-data/CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 - All libcnb-data struct types now reject unrecognised fields when deserializing ([#252](https://github.com/Malax/libcnb.rs/pull/252)).
 - `BuildpackToml` has been replaced by `BuildpackDescriptor`, which is an enum with `Single` and `Meta` variants that wrap new `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` types. The new types now reject `buildpack.toml` files where both `stacks` and `order` are present ([#248](https://github.com/Malax/libcnb.rs/pull/248)).
 - Implement `Borrow<str>` for types generated using the `libcnb_newtype!` macro (currently `BuildpackId`, `LayerName`, `ProcessType` and `StackId`), which allows them to be used with `.join()` ([#258](https://github.com/Malax/libcnb.rs/pull/258)).
-- `Process::args`, `Process::direct`, `Process::default` are now wrapped in `Option` in accordance with the CNB spec ([#243](https://github.com/Malax/libcnb.rs/pull/243)).
+- `Process` can now be deserialized when optional fields are missing. ([#265](https://github.com/Malax/libcnb.rs/pull/265))
```
edmorley, line 13:
```suggestion
- `Launch` and `Process` can now be deserialized when optional fields are missing, and omit default values when serializing ([#243](https://github.com/Malax/libcnb.rs/pull/243) and [#265](https://github.com/Malax/libcnb.rs/pull/265)).
- `Process::new` has been replaced by `ProcessBuilder` ([#265](https://github.com/Malax/libcnb.rs/pull/265)).
```

libcnb-data/src/launch.rs
```diff
@@ -65,30 +66,102 @@ pub struct Label {
...
+                args: Vec::new(),
+                direct: false,
+                default: false,
+            },
         }
     }
+
+    /// Adds an argument to the process.
+    ///
+    /// Only one argument can be passed per use. So instead of:
```
edmorley, line 113: (nit)
"can be passed per use" suggests there is validation. libcnb doesn't validate this, so I presume it's more of a "should"? (Unless Pack/lifecycle do validate?)

```suggestion
    /// Only one argument should be passed per use. So instead of:
```

libcnb-data/src/launch.rs
```diff
@@ -65,30 +66,102 @@ pub struct Label {
...
+    /// # ;
+    /// ```
+    ///
+    /// To pass multiple arguments see [`args`](Self::args).
+    pub fn arg(&mut self, arg: impl Into<String>) -> &mut Self {
+        self.process.args.push(arg.into());
+        self
+    }
+
+    /// Adds multiple arguments to pass to the process.
```
edmorley, line 137: Should this be `s/process/command/` since the args are being passed to what's specified under `command`?

libcnb-data/src/launch.rs
```diff
@@ -65,30 +66,102 @@ pub struct Label {
...
+    /// To pass a single argument see [`arg`](Self::arg).
+    pub fn args(&mut self, args: impl IntoIterator<Item = impl Into<String>>) -> &mut Self {
+        for arg in args {
+            self.arg(arg);
+        }
+
+        self
+    }
+
+    /// Sets the direct flag on the process.
```
edmorley, line 148: (just since I think it may be non-obvious to people who haven't read the spec, what the implications of this bool are)

```suggestion
    /// Sets the `direct` flag on the process.
    ///
    /// If this is true, the lifecycle will launch the command directly, rather than via a shell.
```

libcnb-data/src/launch.rs
```diff
@@ -65,30 +66,102 @@ pub struct Label {
...
+    pub fn direct(&mut self, value: bool) -> &mut Self {
+        self.process.direct = value;
+        self
+    }
+
+    /// Sets the default flag on the process.
```
edmorley, line 154:
```suggestion
    /// Sets the `default` flag on the process.
    ///
    /// Indicates that the process type should be selected as the buildpack-provided
    /// default during the export phase.
```

edmorley (approved):

libcnb-data/src/launch.rs
Malax: This is a verbatim copy of the `Command` function from the standard lib with the same name. I liked that it 100% fits what we're doing here. If you don't feel strongly, I'd like to keep it the way it is.

libcnb-data/src/launch.rs
Malax: It's adding it to the `Process`.  What happens with that data inside the CNB lifecycle is a different beast I think. 🤔

libcnb-data/src/launch.rs
edmorley: Ah, then works for me :-)

lead4good:
@edmorley @Malax any comments on this?

Malax:
@lead4good what do you mean? :) This is merged and released as of `0.5.0`.

lead4good:
@Malax See the review I started: https://github.com/Malax/libcnb.rs/pull/265#discussion_r784149861

Malax:
@lead4good Nope, we can't see it in the "pending" state. You'd need to submit it I think.

lead4good:
Sorry my fault. Please check again.

libcnb-data/src/launch.rs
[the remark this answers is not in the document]
Malax: I see. This would be an issue for all other types in `libcnb-data` as well where we use non-`Option` defaults. This never came up before, but if we'd change `libcnb-data` to support that use-case, we should do it for all types for consistency.

I'm hesitant to make this change lightly, as it's a severely breaking change and in many cases complicates the usage  for the more common use-cases since a user would need to resolve the "missing value" case to the default value by hand.

Maybe we can have types suffixed with `Raw` in `libcnb-data` that use an enum for each optional TOML field that is almost identical to `Option` but conveys additional semantics:

```rust
enum<A> OptionalValue<A> {
     Missing,
     Value(A)
}
```

The types without the `Raw` suffix could stay the same and would resolve `OptionalValue`s during deserialization by replacing them with the default values like they do today.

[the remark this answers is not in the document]
edmorley: Also, whilst the CNB spec does state that there should be a difference between unset and "set to what happens to be the default value", it seems the implementation in `lifecycle` doesn't honour this? See:
https://github.com/Malax/libcnb.rs/pull/265#discussion_r784123333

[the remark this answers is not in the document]
Malax: @edmorley This isn't really about the lifecycle. From my understanding, @lead4good wants to parse to `launch.toml` files, and merge them together. Without knowing which field has the a default value, default values might override explicit values.

PR #267, opened 2022-01-12
Add LayerEnv::apply_to_empty

Malax:
It's a common pattern to apply a layer env and pass the resulting env to another layer. like so:

```rust
ruby_layer.env.apply(Scope::Build, &Env::new())
```

However, they're mostly applied to an empty env which is a little bit clunky. As a QOL enhancement, add `LayerEnv::apply_to_empty` which removes the second argument:

```rust
ruby_layer.env.apply_to_empty(Scope::Build)
```

GUS-W-10423858

### DIFF
--- a/examples/example-02-ruby-sample/src/main.rs
+++ b/examples/example-02-ruby-sample/src/main.rs
@@ -5,7 +5,7 @@ use libcnb::data::{layer_name, process_type};
 use libcnb::detect::{DetectContext, DetectResult, DetectResultBuilder};
 use libcnb::generic::GenericPlatform;
 use libcnb::layer_env::Scope;
-use libcnb::{buildpack_main, Buildpack, Env};
+use libcnb::{buildpack_main, Buildpack};
 
 use crate::util::{DownloadError, UntarError};
 use serde::Deserialize;
@@ -37,7 +37,7 @@ impl Buildpack for RubyBuildpack {
         context.handle_layer(
             layer_name!("bundler"),
             BundlerLayer {
-                ruby_env: ruby_layer.env.apply(Scope::Build, &Env::new()),
+                ruby_env: ruby_layer.env.apply_to_empty(Scope::Build),
             },
         )?;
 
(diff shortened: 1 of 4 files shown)


edmorley (commented):
Thank you for adding this!

I wonder if there is another name that might be more intuitive? Since it's using an empty env, it's not really applying an env to something else, but generating an env object from scratch - so having "apply" in the name seems strange. 

What we have is almost a builder, that we want to eventually resolve to an Env?

As in, we're really modelling two things here - the state of the `<layer>/env` directory, and the result of converting that to another form (an env which may/may not have been based on an empty or existing env).

edmorley (commented):
Separate to that, there are a number of places in the tests that could be switched to using this new method, rather than each having to `Env::new()`. (Mainly in `layer_env.rs` and one instance in `handling.rs`)

Malax:
I think I'm fine with `apply` and `apply_to_empty`. I want to ensure users realise that a LaverEnv is just a delta, on its own it has no meaning, but an `Env` does. In my head, this operation reads like "apply delta to env" and "apply delta to empty env".

edmorley:
Yeah makes sense. Happy to stick with that naming then. Could you move the additional test usages over to it, per https://github.com/Malax/libcnb.rs/pull/267#pullrequestreview-850634197 ? :-)

Malax:
@edmorley I will!

edmorley (approved):
What do you think about adding a cross-reference from `LayerEnv::apply()` to the new method (eg "If you are applying to an empty env you can use ... instead" etc).

Also, should the rustdoc example in `LayerEnv::new()` use the new method instead (it uses an empty env) or perhaps not have an example given the struct has one on it, or ...?

Malax:
@edmorley added extra RustDoc. For `LayerEnv::new()`, in the doctest, the initial and resulting envs are compared to show that applying an empty delta does not change the input `Env`. We can't just remove the initial one and use `apply_to_empty` if we want to keep that explicit example.

edmorley:
Ah I missed that aspect in the `new()` rustdoc, ty.

PR #282, opened 2022-01-18
Allow for multiple non-bin targets

Malax:
Currently `libcnb-cargo` allows exactly one target in the buildpack crate. When a crate uses integration tests, these will be an additional target, causing `cargo libcnb package` to always fail for buildpacks with integration tests.

To work around this, `libcnb-cargo` now only checks if there is a single _binary_ target.

GUS-W-10449039, extracted from #277

### DIFF
--- a/libcnb-cargo/src/lib.rs
+++ b/libcnb-cargo/src/lib.rs
@@ -5,7 +5,7 @@
 
 pub mod cross_compile;
 
-use cargo_metadata::MetadataCommand;
+use cargo_metadata::{MetadataCommand, Target};
 use libcnb_data::buildpack::SingleBuildpackDescriptor;
 use std::ffi::OsStr;
 use std::fs;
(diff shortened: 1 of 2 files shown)


libcnb-cargo/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## [Unreleased]
+- `cargo libcnb package` now allows multiple targets, as long as there is only one binary target. ([#282](https://github.com/Malax/libcnb.rs/pull/282))
```
edmorley, line 4: It's unlikely others are using libcnb-cargo as a library yet, but moving forwards we should probably mention all breaking changes in the changelog (with our open source maintainer hats on :-)).

```suggestion
- `cargo libcnb package` now allows multiple targets, as long as there is only one binary target. To support this change, the `BuildError` variants `NoTargetsFound` and `MultipleTargetsFound` have been replaced by `NoBinTargetsFound` and `MultipleBinTargetsFound` respectively. ([#282](https://github.com/Malax/libcnb.rs/pull/282))
```

edmorley (approved):

PR #283, opened 2022-01-18
Update buildpack assembly to no longer fail on existing output directories

Malax:
This is a remnant from when existing output directories always failed. For CLI usage, the directory is always removed before calling `assemble_buildpack_directory`. Therefore, this PR changes nothing for CLI usage.

This change was originally made for #277 to allow buildpack directory assembly into a `tempdir::TempDir` which will always exist when `assemble_buildpack_directory` is called.

Closes GUS-W-10449107

### DIFF
--- a/libcnb-cargo/CHANGELOG.md
+++ b/libcnb-cargo/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - `cargo libcnb package` now allows multiple targets, as long as there is only one binary target. To support this change, the `BuildError` variants `NoTargetsFound` and `MultipleTargetsFound` have been replaced by `NoBinTargetsFound` and `MultipleBinTargetsFound` respectively. ([#282](https://github.com/Malax/libcnb.rs/pull/282))
+- `assemble_buildpack_directory()` no longer fails if then output directory already exists. ([#283](https://github.com/Malax/libcnb.rs/pull/283))
 
 ## [0.2.0] 2022-01-14
 
--- a/libcnb-cargo/CHANGELOG.md
+++ b/libcnb-cargo/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 - `cargo libcnb package` now allows multiple targets, as long as there is only one binary target. To support this change, the `BuildError` variants `NoTargetsFound` and `MultipleTargetsFound` have been replaced by `NoBinTargetsFound` and `MultipleBinTargetsFound` respectively. ([#282](https://github.com/Malax/libcnb.rs/pull/282))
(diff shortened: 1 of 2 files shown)


libcnb-cargo/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- `assemble_buildpack_directory()` no longer fails if then output directory already exists. ([#283](https://github.com/Malax/libcnb.rs/pull/283))
```
edmorley, line 5:
```suggestion
- `assemble_buildpack_directory()` no longer fails if the output directory already exists. ([#283](https://github.com/Malax/libcnb.rs/pull/283))
```

libcnb-cargo/src/lib.rs
```diff
@@ -153,33 +152,26 @@ pub struct BuildpackData<BM> {
...
 
-        let bin_path = destination_path.as_ref().join("bin");
-        fs::create_dir_all(&bin_path)?;
+    let bin_path = destination_path.as_ref().join("bin");
+    fs::create_dir_all(&bin_path)?;
 
-        fs::copy(buildpack_binary_path.as_ref(), bin_path.join("build"))?;
-        create_file_symlink("build", bin_path.join("detect"))?;
+    fs::copy(buildpack_binary_path.as_ref(), bin_path.join("build"))?;
+    create_file_symlink("build", bin_path.join("detect"))?;
```
edmorley, line 172: Will symlink creation fail if the symlink already exists? If so, does this affect the #277 use-case?

Looking at:
https://doc.rust-lang.org/src/std/os/unix/fs.rs.html#899-901

...it calls:
https://doc.rust-lang.org/src/std/sys/unix/fs.rs.html#1136

And these libc docs seem to suggest that could fail if the file already exists:
https://ftp.gnu.org/old-gnu/Manuals/glibc-2.2.3/html_node/libc_269.html

> EEXIST
>    There is already an existing file named newname.

edmorley (approved):

libcnb-cargo/src/lib.rs
Malax: Sure, but a similar concern would also apply to all other FS/IO operations in that function. I think it's perfectly fine to not add additional error handling for these cases. `assemble_buildpack_directory` already returns `std::io::Result` which can signal any IO (and FS) error that's happening. It's not like the callsite can reasonably recover anyway.

#277 will always pass an empty directory, it will not run into the specific issue you're mentioning.

edmorley: I wasn't meaning additional error handling - it just wasn't clear whether the directory would be empty or not - though sounds like it will be, so this doesn't apply.

PR #292, opened 2022-01-28
Fix CI badge in README

edmorley:
Since the branch is now `main` not `master`, and the workflow name `CI` not `Rust`.


Malax (approved):
Weird. I spotted that yesterday and tried to do the exact same fix but it didn't do anything. 🤔  This seems to work though - 👍🏻 .

PR #293, opened 2022-01-28
Switch from `rand` to `fastrand` in libcnb layer tests

edmorley:
Since this usage doesn't need to be cryptographically secure, and `fastrand` is a smaller dependency, and already in the dependency tree (via `tempfile`).

Similar to that discussed in:
https://github.com/Malax/libcnb.rs/pull/277#discussion_r793516863

### DIFF
--- a/libcnb/src/layer/test.rs
+++ b/libcnb/src/layer/test.rs
@@ -25,10 +25,10 @@ use libcnb_data::buildpack::{BuildpackVersion, SingleBuildpackDescriptor, Stack}
 use libcnb_data::buildpack_plan::BuildpackPlan;
 use libcnb_data::layer::LayerName;
 use libcnb_data::layer_content_metadata::LayerContentMetadata;
-use rand::Rng;
 use serde::Deserialize;
 use serde::Serialize;
 use std::fs;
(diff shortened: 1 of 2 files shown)


libcnb/src/layer/test.rs
```diff
@@ -922,10 +922,8 @@ fn build_context(temp_dir: &TempDir) -> BuildContext<TestBuildpack> {
 }
 
 fn random_layer_name() -> LayerName {
-    rand::thread_rng()
-        .sample_iter(rand::distributions::Alphanumeric)
+    repeat_with(fastrand::lowercase)
```
Malax, line 925: FWIW: this changes the test slightly since the names are now only lowercased.

```suggestion
    repeat_with(fastrand::alphanumeric)
```

Malax (approved):

libcnb/src/layer/test.rs
edmorley: Yeah I noticed that, but had thought it didn't matter, since otherwise we'd have explicit tests for layer casing, rather than it be non-deterministic? However happy to change it back for parity.

PR #305, opened 2022-02-03
Support connecting to Docker daemon on a remote machine

Malax:
Previously, `libcnb-test` only used a local socket or named pipe to connect to Docker. This PR changes this to use the connection string from `DOCKER_HOST` to determine where the Docker API is hosted. This PR also adds HTTPS support for Docker connections.

This enables usage of libcnb-test on CircleCI where the Docker deamon is on a remote machine, configured by the `DOCKER_HOST` environment variable.

There is an additional issue on CircleCI where the test is not able to connect to the running container (see: https://circleci.com/docs/2.0/building-docker-images/#accessing-services). This will be addressed in a separate PR.

Closes GUS-W-10533383

### DIFF
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Use the `DOCKER_HOST` environment variable to determine the Docker connection strategy, adding support for HTTPS 
 connections in addition to local UNIX sockets. This enables using `libcnb-test` in more complex setups like on CircleCI 
-where the Docker deamon is on a remote machine.
+where the Docker deamon is on a remote machine. ([#305](https://github.com/Malax/libcnb.rs/pull/305))
 
 ## [0.1.0] 2022-02-02
 
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,10 +2,9 @@
 
 ## [Unreleased]
 
-## [0.1.1] 2022-02-03
-
-- Connect to Docker demon via HTTP by default, only fall back to local socket/named pipe if no `DOCKER_HOST`
-environment variable is set. ([#305](https://github.com/Malax/libcnb.rs/pull/305))
+- Use the `DOCKER_HOST` environment variable to determine the Docker connection strategy, adding support for HTTPS 
+connections in addition to local UNIX sockets. This enables using `libcnb-test` in more complex setups like on CircleCI 
+where the Docker deamon is on a remote machine.
 
 ## [0.1.0] 2022-02-02
 
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+## [0.1.1] 2022-02-03
+
+- Connect to Docker demon via HTTP by default, only fall back to local socket/named pipe if no `DOCKER_HOST`
+environment variable is set. ([#305](https://github.com/Malax/libcnb.rs/pull/305))
+
 ## [0.1.0] 2022-02-02
 
 - Initial release ([#277](https://github.com/Malax/libcnb.rs/pull/277)).
--- a/libcnb-test/src/lib.rs
+++ b/libcnb-test/src/lib.rs
@@ -21,6 +21,7 @@ use bollard::image::RemoveImageOptions;
 use bollard::Docker;
 
 use std::env;
+use std::env::VarError;
 use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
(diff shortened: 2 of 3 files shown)


libcnb-test/src/lib.rs
```diff
@@ -85,15 +86,39 @@ impl IntegrationTest {
...
+                    Err(bollard::errors::Error::HyperResponseError { err })
+                        if err.is_connect()
+                            && env::var("DOCKER_HOST") == Err(VarError::NotPresent) =>
+                    {
+                        Docker::connect_with_local_defaults()
+                    }
+                    Ok(_) => Ok(docker), // Ping successful, use connection.
+                    Err(err) => Err(err), // Ping failed for other reasons than connection errors,
+                                          // keep error as-is for later display to user.
+                }
```
edmorley, line 111: As discussed on Slack just now, I think this won't work since Circle CI's Docker config needs SSL. When expanding the "Setup a remote Docker engine" section on [this CI run](https://app.circleci.com/pipelines/github/heroku/procfile-cnb-rust/114/workflows/c022ba84-945e-4d58-90e1-6eb1643531e7/jobs/229/steps), the env vars set are:

```
  DOCKER_CERT_PATH=/tmp/docker-certs1511074844
  DOCKER_HOST=tcp://ec2-3-92-79-176.compute-1.amazonaws.com:2376
  DOCKER_MACHINE_NAME=172854067
  DOCKER_TLS_VERIFY=1
  NO_PROXY=127.0.0.1,localhost,circleci-internal-outer-build-agent,ec2-3-92-79-176.compute-1.amazonaws.com:237
```
...which includes `DOCKER_TLS_VERIFY=1`.

Also, I think I would prefer to try and pick the best connection type up front, rather than potentially try the wrong option and then fallback every time (particularly since it will be invisible to the user/us, so we'd never know to fix it).

Ideally this would be part of bollard (eg `connect_with_defaults()`) - I think we should file an issue, but implement here in the meantime.

From searching around, eg:
https://github.com/search?q=org%3Amoby+DOCKER_HOST&type=Code
https://github.com/moby/moby/blob/3c06ebd876687555fdf030a3307a66908c4fa57c/client/client_unix.go
https://github.com/moby/moby/blob/3c06ebd876687555fdf030a3307a66908c4fa57c/client/client_windows.go
https://github.com/moby/moby/blob/3c06ebd876687555fdf030a3307a66908c4fa57c/client/client_test.go

...it seems that typical protocols for `DOCKER_HOST` are:
- `unix://`
- `npipe:////` (Windows)
- `tcp://`
- `http://` / `https://` (less common; presumably because of [this reason](https://github.com/docker/cli/blob/c9f84735128fa5fbf3ffa76b8940ea42aa675011/e2e/global/cli_test.go#L38-L39))

Now bollard's `connect_with_local_defaults()` (which on unix calls `connect_with_unix_defaults()`) already uses `DOCKER_HOST` iff it's set to `unix://`:
https://github.com/fussybeaver/bollard/blob/c5d87a4934c70a04f9c649fedb241dbd4943c927/src/docker.rs#L629-L641

Therefore given the above, and the fact that I don't think we should only support HTTPS (and not HTTP, since HTTPS is mandatory for Docker by default), I think here all we need to do here, is:

- if `DOCKER_HOST` set and starts with `tcp://` or `https://`, then use `connect_with_ssl_defaults()`
- otherwise use `connect_with_local_defaults()`

What do you think? :-)

edmorley (changes requested):
In addition to the comment below, I'm curious to see if making requests to the container's exposed port works already, or if it needs further adjustments given:
https://circleci.com/docs/2.0/building-docker-images/#accessing-services

libcnb-test/src/lib.rs
Malax: > What do you think?

I wrote 7d17cad before reading this... it implements it almost exactly like that :D I'll iterate a bit more and test on CircleCI - but I expect this to be the most viable route. :)

joshwlewis: > I think here all we need to do here, is:
> - If DOCKER_HOST set and starts with tcp:// or https://, then use connect_with_ssl_defaults() (this will mostly be used only in CI, or for users who have advanced configs locally)
> - otherwise use connect_with_local_defaults()

I was digging through pack's source, and it seems to match this pretty closely.

libcnb-test/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+- Use the `DOCKER_HOST` environment variable to determine the Docker connection strategy, adding support for HTTPS 
+connections in addition to local UNIX sockets. This enables using `libcnb-test` in more complex setups like on CircleCI 
+where the Docker deamon is on a remote machine.
```
edmorley, line 7:
```suggestion
where the Docker deamon is on a remote machine. ([#305](https://github.com/Malax/libcnb.rs/pull/305))
```

libcnb-test/src/lib.rs
```diff
@@ -85,15 +86,37 @@ impl IntegrationTest {
...
+            {
+                Docker::connect_with_ssl_defaults()
+            }
+            Ok(docker_host) => panic!(
+                "Cannot connect to unsupported DOCKER_HOST '{}'",
+                docker_host
+            ),
+            Err(VarError::NotPresent) => Docker::connect_with_local_defaults(),
+            Err(VarError::NotUnicode(_)) => {
+                panic!("DOCKER_HOST environment variable is not unicode encoded!")
```
edmorley, line 108: (pedantic) I suppose the cause of this error could also be that it's unicode encoded, just not correctly? The [NotUnicode](https://doc.rust-lang.org/std/env/enum.VarError.html#variant.NotUnicode) docs use the phrase "did not contain valid unicode".

libcnb-test/src/lib.rs
```diff
@@ -85,15 +86,37 @@ impl IntegrationTest {
...
+            }
+        }
+        .expect("Could not connect to local Docker deamon");
```
edmorley, line 111:
```suggestion
        .expect("Could not connect to local Docker daemon");
```

libcnb-test/src/lib.rs
```diff
@@ -85,15 +86,37 @@ impl IntegrationTest {
...
+            Ok(docker_host) if docker_host.starts_with("unix://") => {
+                Docker::connect_with_unix_defaults()
+            }
+            Ok(docker_host)
+                if docker_host.starts_with("tcp://") || docker_host.starts_with("https://") =>
+            {
+                Docker::connect_with_ssl_defaults()
+            }
+            Ok(docker_host) => panic!(
+                "Cannot connect to unsupported DOCKER_HOST '{}'",
```
edmorley, line 103:
```suggestion
                "Cannot connect to unsupported DOCKER_HOST protocol '{}'",
```

edmorley (approved):
Thank you for persevering with this - looks great!

libcnb-test/src/lib.rs
Malax: Thing is, that string would be more than just the protocol. Since this is so minor, I will skip this. :)

libcnb-test/src/lib.rs
Malax: Is it unicode if it's not valid unicode? ;-)

PR #313, opened 2022-02-09
Make `libcnb_runtime_build` and `libcnb_runtime_detect` public

nokome:
This exposes `libcnb_runtime_build` and `libcnb_runtime_detect` (and associated types) so that they can be used by other crates.

For my use case, instead of building separate executable for each buildpack, I would like to combine several buildpacks into an existing CLI executable and delegate to their `detect` and `build` methods. This change allows me to do that without rewriting alot of the code contained in these two functions e.g.

```rust
/// A local trait for buildpacks that extends `libcnb::Buildpack`
///
/// Why? To provide some additional introspection and the ability
/// to compile several buildpacks into a single binary and call
/// their `detect` and `build` methods.
pub trait BuildpackTrait: libcnb::Buildpack {
    ...

    /// Run the buildpack's `detect` method
    fn detect_with(&self, platform_dir: &Path, build_plan: &Path) -> Result<i32>
    where
        Self: Sized,
    {
        env::set_var("CNB_STACK_ID", CNB_STACK_ID);

        let buildpack_dir = self.ensure_dir()?;
        env::set_var("CNB_BUILDPACK_DIR", buildpack_dir);

        match libcnb_runtime_detect(
            self,
            DetectArgs {
                platform_dir_path: PathBuf::from(platform_dir),
                build_plan_path: PathBuf::from(build_plan),
            },
        ) {
            Ok(code) => Ok(code),
            Err(error) => bail!(
                "While running `detect` for buildpack `{}`: {}",
                Self::id(),
                error
            ),
        }
    }
```

### DIFF
--- a/libcnb/src/layer_env.rs
+++ b/libcnb/src/layer_env.rs
@@ -510,9 +510,15 @@ impl LayerEnvDelta {
             // explicitly written in the spec, we read through the the reference implementation and
             // determined that it also treats the file contents as raw bytes.
             // See: https://github.com/buildpacks/lifecycle/blob/a7428a55c2a14d8a37e84285b95dc63192e3264e/env/env.go#L73-L106
-            use std::os::unix::ffi::OsStringExt;
             let path = dir_entry?.path();
-            let file_contents = OsString::from_vec(fs::read(&path)?);
+
+            #[cfg(target_family = "unix")]
+            let file_contents = {
+                use std::os::unix::ffi::OsStringExt;
+                OsString::from_vec(fs::read(&path)?)
+            };
+            #[cfg(not(target_family = "unix"))]
+            let file_contents = OsString::from(&fs::read_to_string(&path)?);
 
             // Rely on the Rust standard library for splitting stem and extension. Since paths
             // are not necessarily UTF-8 encoded, this is not as trivial as it might look like.
@@ -549,8 +555,6 @@ impl LayerEnvDelta {
     }
 
     fn write_to_env_dir(&self, path: impl AsRef<Path>) -> Result<(), std::io::Error> {
-        use std::os::unix::ffi::OsStrExt;
-
         if path.as_ref().exists() {
             // This is a possible race condition if the path is deleted between the check and
             // removal by this code. We accept this for now to keep it simple.
@@ -573,7 +577,14 @@ impl LayerEnvDelta {
 
             let file_path = path.as_ref().join(file_name);
 
-            fs::write(file_path, &value.as_bytes())?;
+            #[cfg(target_family = "unix")]
+            {
+                use std::os::unix::ffi::OsStrExt;
+                fs::write(file_path, &value.as_bytes())?;
+            }
+
+            #[cfg(not(target_family = "unix"))]
+            fs::write(file_path, &value.to_string_lossy().as_bytes())?;
         }
 
         Ok(())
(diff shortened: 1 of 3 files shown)


Malax:
Hey @nokome 👋🏻 

Thanks for your PR! In general, I can see us exposing `libcnb_runtime_build` and `libcnb_runtime_detect`.

Before we continue with this, I'd love to hear more about your use-case. Which benefits do you get by having multiple buildpacks in an existing binary and how are you packaging it? I never thought of doing that and maybe there is even more we can do to support your use-case if it's applicable to a broader audience?

nokome:
Thanks for your interest @Malax. I'm putting together a PR for our project, https://github.com/stencila/stencila, that will introduce `libcnb` and outline the rationale for doing so. That is probably the best place to answer your questions about our use case. I can ping you when that's ready (hopefully in a day or two).

edmorley (commented):
Thank you for the PR and the links to your project - it's very exciting to see other usages of libcnb! I'm curious, will your project interact with Pack CLI, or is it effectively an alternative implementation of the CNB [lifecycle](https://buildpacks.io/docs/concepts/components/lifecycle/)?

Re this PR, at first glance these changes seem great, I just wonder if it might be best to split them into a few PRs, since there are a few distinct changes here:
1. Moving args handling + the `process::exit`s up a level (to `libcnb_runtime`)
2. Making `libcnb_runtime_build`, `libcnb_runtime_detect`, `DetectArgs` and `BuildArgs` public
3. Improving support for non-Unix platforms in `LayerEnv` and `libcnb_runtime`

nokome:
Hi @edmorley. Sorry for taking so long to reply: I've had my head down with my other PR and somehow missed the notification.

> I'm curious, will your project interact with Pack CLI, or is it effectively an alternative implementation of the CNB [lifecycle](https://buildpacks.io/docs/concepts/components/lifecycle/)?

Both. Our users are data scientists and researchers (who often don't have software dev experience) and we want to provide a transparent and seam less as possible journey form a local project to a docker image. So the use case is perhaps a little different to the usual web dev use case for CNB's.

We're using `cargo libcnb package` and creating a CNB builder to use with `pack` but we also have a "lifecycle-lite" so that users can use the `stencila` CLI to preview which buildpacks match against a project (and perhaps guide them to making it more reproducible) e.g.

```sh
> stencila buildpacks plan fixtures/projects/polyglot/small

Buildpack 'stencila/apt'

- apt_packages        Install apt packages for Ubuntu 'impish' (Aptfile)

Buildpack 'stencila/node'

- node                Install Node.js lts
- node_modules        Install Node.js packages into node_modules (package-lock.json)

Buildpack 'stencila/python'

- python              Install Python ^3.10 (pyproject.toml)
- poetry              Install Poetry (pyproject.toml)
- venv                Install Python packages into virtual environment using Poetry (pyproject.toml)

Buildpack 'stencila/r'

- r                   Install R 4.1.2 (renv.lock)
- renv                Install R packages into renv by restoring from lockfile (renv.lock)

Buildpack 'stencila/stencila'

- stencila            Install Stencila >=1.1.0
```

> I just wonder if it might be best to split them into a few PRs, since there are a few distinct changes here:

I feel that 1 & 2 are closely connected so to avoid unnecessary extra work would prefer to combine those. Would it be OK to submit two PRs for each of my two commits:

- https://github.com/Malax/libcnb.rs/pull/313/commits/af5cfec194242c01acb3b22797e2b8fc634f8735
- https://github.com/Malax/libcnb.rs/pull/313/commits/638e74de99b3acbf7e7923d8128551bb41c4aadc

nokome:
I have splot this into 3 separate PRS: #368, #369, and #370. Will close this one.

PR #314, opened 2022-02-11
Package all binary targets

Malax:
This PR is the first of a series of PRs to support exec.d across libcnb.

Changes `libcnb-cargo` and `libcnb-test` to package all binaries of a buildpack crate into the final buildpack. Previously, a buildpack crate could only contain one binary target. Allowing multiple binaries enables use-cases like `exec.d` binaries written in Rust that exist in the buildpack that uses them. These binaries will be packaged into `.libcnb-cargo/additional-bin/$TARGET_NAME`.

An additional PR that is coming soon will make it easier to reference these binaries via macros at build time.

Closes GUS-W-10670969

### DIFF
--- a/libcnb-cargo/src/config.rs
+++ b/libcnb-cargo/src/config.rs
@@ -1,0 +1,68 @@
+use cargo_metadata::{Metadata, Target};
+use serde;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+pub struct Config {
+    #[serde(rename = "buildpack-target")]
+    pub buildpack_target: String,
+}
+
+/// # Errors
+pub fn config_from_metadata(cargo_metadata: &Metadata) -> Result<Config, ConfigError> {
+    let root_package = cargo_metadata
+        .root_package()
+        .ok_or(ConfigError::MissingRootPackage)?;
+
+    let explicit_config = root_package
+        .metadata
+        .as_object()
+        .unwrap_or(&serde_json::Map::default())
+        .get("libcnb")
+        .map(|libcnb_object| {
+            serde_json::from_value::<Config>(libcnb_object.clone())
+                .map_err(ConfigError::ConfigJsonError)
+        })
+        .transpose();
+
+    explicit_config.and_then(|maybe_config| {
+        maybe_config.map_or_else(
+            || {
+                let buildpack_bin_targets: Vec<&Target> = root_package
+                    .targets
+                    .iter()
+                    .filter(|target| target.kind == vec!["bin"])
+                    .collect();
+
+                match buildpack_bin_targets.as_slice() {
+                    [single_target] => Ok(Config {
+                        buildpack_target: single_target.name.clone(),
+                    }),
+                    [] => Err(ConfigError::NoBinTargetsFound),
+                    bin_target_names => {
+                        let has_bin_target_with_root_package_name = bin_target_names
+                            .iter()
+                            .any(|target_name| target_name.name == root_package.name);
+
+                        if has_bin_target_with_root_package_name {
+                            Ok(Config {
+                                buildpack_target: root_package.name.clone(),
+                            })
+                        } else {
+                            Err(ConfigError::MultipleBinTargetsFound)
+                        }
+                    }
+                }
+            },
+            Ok,
+        )
+    })
+}
+
+#[derive(Debug)]
+pub enum ConfigError {
+    MissingRootPackage,
+    ConfigJsonError(serde_json::Error),
+    NoBinTargetsFound,
+    MultipleBinTargetsFound,
+}
--- a/libcnb-cargo/src/config.rs
+++ b/libcnb-cargo/src/config.rs
@@ -3,13 +3,12 @@ use serde;
 use serde::Deserialize;
 
 #[derive(Debug, Deserialize)]
-pub struct Config {
+pub(crate) struct Config {
     #[serde(rename = "buildpack-target")]
     pub buildpack_target: String,
 }
 
-/// # Errors
-pub fn config_from_metadata(cargo_metadata: &Metadata) -> Result<Config, ConfigError> {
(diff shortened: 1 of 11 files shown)


libcnb-test/README.md
```diff
@@ -46,3 +46,27 @@ fn call_test_fixture_service(addr: std::net::SocketAddr, payload: &str) -> Resul
...
+
+## Configuration
+
+### Main buildpack binary
+
+A buildpack crate can contain multiple binaries. To determine which binary is the main buildpack binary, the convention is:
+1. If there is only one binary, use that.
+2. If there are multiple, use the one that has the same name as the crate.
+
+If this convention does not work for your buildpack, you can explicitly configure the main buildpack binary in `Cargo.toml`:
```
edmorley, line 58: What use-cases do we see for needing to configure `buildpack-target` via `Cargo.toml`? Unless there is something significant, should we wait until we need this, before adding support?

libcnb-test/README.md
```diff
@@ -46,3 +46,27 @@ fn call_test_fixture_service(addr: std::net::SocketAddr, payload: &str) -> Resul
    unimplemented!()
 }
 ```
+
+## Configuration
```
edmorley, line 50: It seems configuring the target via `Cargo.toml` works for both `libcnb-test` and `libcnb-cargo` CLI usage, however the config is only documented for `libcnb-test`. Should this be added to the `libcnb-cargo` README too? Somewhere else instead and linked to from both?

(Perhaps another reason why skipping allowing this configuration might be simpler for now, unless we really need it?)

libcnb-cargo/src/config.rs
```diff
@@ -0,0 +1,68 @@
...
+use serde;
+use serde::Deserialize;
+
+#[derive(Debug, Deserialize)]
+pub struct Config {
+    #[serde(rename = "buildpack-target")]
+    pub buildpack_target: String,
+}
+
+/// # Errors
```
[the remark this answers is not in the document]
Malax: I already removed it and made it private :)

libcnb-cargo/src/lib.rs
```diff
@@ -174,9 +77,27 @@ pub fn assemble_buildpack_directory(
...
+    fs::copy(
+        &buildpack_binaries.buildpack_target_binary_path,
+        bin_path.join("build"),
+    )?;
+
     create_file_symlink("build", bin_path.join("detect"))?;
 
+    let additional_binaries_dir = destination_path
+        .as_ref()
+        .join(".libcnb-cargo")
```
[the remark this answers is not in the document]
Malax: Woops, good catch! I will fix the wrong one. :) 

I wanted to have a prefix for these implicitly added files so that they'll never clash with something explicit. I was thinking about putting them in `bin` at first, but what happens if a user calls their binary `detect`? Or `env` which might be used by the CNB spec in the future?

By putting them in a `.libcnb-cargo` directory, we know these files from from `libcnb-cargo` and the change of clashing with something else is very low. I can see us using the directory for other things in the the future too. For example, we could put some build metadata there to aid with debugging.

Users, at least for the exec.d case will not have to construct this path by hand, a macro will be supplied that can be used to get a `PathBuf` to any Crate binary.

I don't see a problem with this deep path and would vote to keep it as-is. Thoughts?

libcnb-test/README.md
Malax: This is necessary when you don't put your `buildpack_main!` in `main.rs`, or configure your binary name via `Cargo.toml`. I feel like we should have this escape hatch from the get-go, otherwise folks would be stuck in such a case.

libcnb-test/README.md
Malax: I moved the main docs to `libcnb-cargo` and added a reference to them in `libcnb-test`'s README. I don't think we need to duplicate the docs, `libcnb-test` is only really useful if you use `libcnb-cargo` for your release buildpack too.

libcnb-test/README.md
edmorley: Yeah I figured those were the times where the option might be necessary, but that still doesn't answer the question of what use-cases would this be used for?

libcnb intends to be opinionated - I think it's fine for us to pick opinionated sensible defaults and wait until someone asks before adding additional customisation options - particularly for `libcnb-cargo` which is for a very specific cross-compiling/packaging workflow. The problem with adding options up front, is we'll never know if anyone actually needs them or if they could be removed in the future.

For this option in particular, I'm struggling to think of a time where someone might want to customise the buildpack binary name but yet still be using libcnb-cargo. ie: Either they are using libcnb-cargo since they want a fully packaged buildpack (in which case the binary name in the `target/` directory is moot, since libcnb-cargo will rename it to `build` anyway), or else are doing something custom and likely not using `libcnb-cargo`.

For the "the buildpack isn't the main crate binary" case, I'm also not sure this is ever likely to happen. If say someone wants to bundle multiple buildpacks into the same crate, libcnb-cargo doesn't support that anyway. Or if they say have some other CLI tool that they want as the main binary, with the buildpack as an additional binary, then libcnb-cargo is going to bundle that CLI tool into the package whether they like it or not. And if the CLI tool is meant to be packaged into the buildpack (because it's some utility or intended for exec.d etc), then why isn't it a secondary binary in the crate, given it's purpose is less important than the buildpack itself?

libcnb-cargo/src/lib.rs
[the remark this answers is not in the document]
Malax: @edmorley and I discussed this on Slack and decided to keep the path as-is.

libcnb-test/README.md
Malax: @edmorley and I discussed this for a bit in Slack. We decided to remove the explicit config (dfc90e8) for the reasons outlined by Ed above. It can be easily added back in should we ever need it.

libcnb-test/README.md
Malax: Obsolete after dfc90e8.

edmorley (approved):
Looks great - thank you :-)

PR #318, opened 2022-02-13
Make use of implicitly named `format!` args

edmorley:
In Rust 1.58, support for implicitly named arguments to `format!` (and related) was stabilised:
https://rust-lang.github.io/rfcs/2795-format-args-implicit-identifiers.html
https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1580-2022-01-13

This switches applicable usages to the new, more readable syntax, and bumps the MSRV to Rust 1.58. (We'll likely need to bump the MSRV again for the stabilised `strip` feature in Rust 1.59, which means it's not worth forgoing the bump here and saying we won't use the new format feature for now.)

I've bumped the MSRV for all crates (even those not using the new syntax) for consistency and to prevent us forgetting to adjust the remaining crates if they ever start using the new syntax in the future.

GUS-W-10727905.


Malax (approved):
👍🏻

PR #311, opened 2022-02-08
(Draft) - Add exec.d Support

Malax:
Draft - Will be split up in multiple PRs.

PR #320, opened 2022-02-14
Add `additional_buildpack_binary_path` macro

Malax:
Follow-up to: #314 

Resolves the path to an additional buildpack binary by Cargo target name when the buildpack is packaged with `libcnb-cargo` or `libcnb-test`.

This can be used to copy additional binaries to layers or use them for exec.d. To add an additional binary to a buildpack, add a new file with a main function to `bin/`. Cargo will [automatically configure it as a binary target](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery) with the name of file.

 Note: This only works properly if the buildpack is packaged with `libcnb-cargo`/`libcnb-test`.

## Implementation notes:

This follows the pattern for the existing proc macro where the proc macro itself only concerns itself with the logic that strictly requires a proc macro. In this case, verifying that the given binary target name is valid. Additional logic is implemented in a regular macro (here: [additional_buildpack_binary_path](https://github.com/Malax/libcnb.rs/pull/320/files#diff-7bd0cf110112f04611a6f19e1f6dc0248913c7493e0a45948258ad9333e23565R85-R117)).

Strictly speaking, the macro should live in `libcnb-cargo` as it only works when a buildpack is packaged with `libcnb-cargo`. However, since `libcnb-cargo` already is a library for implementing packaging, adding a dependency to it from an actual buildpack seemed weird. It also would would hinder discoverability. Therefore, I decided to add it to `libcnb` proper.

I'm slightly unhappy with the hardcoding of the path to the `additional-bin` directory in `libcnb`. I considered adding shared code between `libcnb-cargo` and `libcnb` that concerns itself with the construction of that path. To make that work, we would've needed to create a new shared crate to avoid circular dependencies (i.e. `libcnb-common`). For now, this seems to be overkill but should be reconsidered when we have more use-cases for a shared crate.

Closes GUS-W-10695287

### DIFF
--- a/libcnb/src/lib.rs
+++ b/libcnb/src/lib.rs
@@ -18,6 +18,10 @@ pub mod generic;
 pub mod layer;
 pub mod layer_env;
 
+// Internals that need to be public for macros
+#[doc(hidden)]
+pub mod internals;
+
 mod buildpack;
 mod env;
 mod error;
@@ -75,6 +79,39 @@ macro_rules! buildpack_main {
         fn main() {
             ::libcnb::libcnb_runtime(&$buildpack);
         }
+    };
+}
+
+/// Resolves the path to an additional buildpack binary by Cargo target name when the buildpack is
+/// packaged with `libcnb-cargo` or `libcnb-test`.
+///
+/// This can be used to copy additional binaries to layers or use them for exec.d.
+///
+/// To add an additional binary to a buildpack, add a new file with a main function to `bin/`.
+/// Cargo will [automatically configure it as a binary target](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery)
+/// with the name of file.
+///
+/// Note: This only works properly if the buildpack is packaged with `libcnb-cargo`/`libcnb-test`.
+#[macro_export]
+macro_rules! additional_buildpack_binary_path {
+    ($target_name:expr) => {
+        ::libcnb::internals::verify_bin_target_exists!(
+            $target_name,
+            {
+                ::std::env::var("CNB_BUILDPACK_DIR")
+                    .map(::std::path::PathBuf::from)
+                    .expect("Could not read CNB_BUILDPACK_DIR environment variable")
+                    .join(".libcnb-cargo")
+                    .join("additional-bin")
+                    .join($target_name)
+            },
+            {
+                compile_error!(concat!(
+                    $target_name,
+                    " is not a valid binary target in this buildpack crate!"
+                ))
+            }
+        )
     };
 }
 
--- a/libcnb/src/lib.rs
+++ b/libcnb/src/lib.rs
@@ -92,6 +92,18 @@ macro_rules! buildpack_main {
 /// with the name of file.
 ///
 /// Note: This only works properly if the buildpack is packaged with `libcnb-cargo`/`libcnb-test`.
+///
+/// ```no_run,compile_fail
+/// use libcnb::additional_buildpack_binary_path;
+/// # let layer_dir = std::path::PathBuf::from(".");
+///
+/// std::fs::copy(
+///     // This would not compile in this doctest since there is no `runtime_tool` binary target.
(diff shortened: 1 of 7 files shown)


edmorley:
> I'm slightly unhappy with the hardcoding of the path to the `additional-bin` directory in `libcnb`. I considered adding shared code between `libcnb-cargo` and `libcnb` that concerns itself with the construction of that path.

This seems fine to me for now, since:
- We at least have integration tests here that ensure the two constants are in sync.
- Even if we did split this shared code out, it wouldn't prevent issues with the values being changed in the future, since for example if a buildpack hadn't yet updated its dependency on libcnb, but pulled in latest libcnb-cargo (where the path had been changed), then they would see breakage (unless libcnb-cargo eventually has an allowlist of supported libcnb framework versions etc). As such, we'll need to think carefully should we ever change the path in the future anyway, so hardcoded vs shared code in separate crate doesn't really buy us much.

libcnb/src/lib.rs
```diff
@@ -78,6 +82,39 @@ macro_rules! buildpack_main {
...
+#[macro_export]
+macro_rules! additional_buildpack_binary_path {
+    ($target_name:expr) => {
+        ::libcnb::internals::verify_bin_target_exists!(
+            $target_name,
+            {
+                ::std::env::var("CNB_BUILDPACK_DIR")
+                    .map(::std::path::PathBuf::from)
+                    .expect("Could not read CNB_BUILDPACK_DIR environment variable")
+                    .join(".libcnb-cargo")
```
edmorley, line 104: Should we add a comment here and in `libcnb-cargo` about keeping these in sync?

libcnb/src/lib.rs
```diff
@@ -78,6 +82,39 @@ macro_rules! buildpack_main {
...
+///
+/// This can be used to copy additional binaries to layers or use them for exec.d.
+///
+/// To add an additional binary to a buildpack, add a new file with a main function to `bin/`.
+/// Cargo will [automatically configure it as a binary target](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#target-auto-discovery)
+/// with the name of file.
+///
+/// Note: This only works properly if the buildpack is packaged with `libcnb-cargo`/`libcnb-test`.
+#[macro_export]
+macro_rules! additional_buildpack_binary_path {
```
edmorley, line 96: I presume it's not possible to add unit tests for this macro since:
- the failing case won't compile (so can't be tested without having some kind of runner that invokes Rust)
- the passing case won't work in this crate, since there are no binary targets we can test this with?

Also, even if we could test the passing case at least compiles, it wouldn't be testing the runtime aspect anyway.

As such, testing this only via a future integration test (eg the new examples-03 I see drafted in #311) seems fine to me :-)

libcnb/src/lib.rs
```diff
@@ -78,6 +82,39 @@ macro_rules! buildpack_main {
     };
 }
 
+/// Resolves the path to an additional buildpack binary by Cargo target name when the buildpack is
```
edmorley, line 85: This sentence is quite hard for me to read. I think some additional punctuation or moving the `libcnb-cargo` / `libcnb-test` out of the opening line, might help?

Also, would it be worth showing an example usage here? I presume it would need to be `no_run` (since it wouldn't compile with a made up binary name), but it might at least make the purpose clearer? :-)

edmorley (approved):

libcnb/src/lib.rs
edmorley: What about `compile_fail`? (TIL)
https://doc.rust-lang.org/rustdoc/documentation-tests.html?#attributes

Malax: I added an example with an explanation why it is marked as failing. (There is a big red exclamation mark next to the example in the docs when `compile_fail` is used).

Malax: I removed the libcnb-cargo info from the opening line. The libcnb-cargo/test caveat is already a note in the body and IMO enough. :)

PR #323, opened 2022-02-15
Remove numbering of examples

Malax:
Removes the numbered prefix of all examples. With the two examples right now the numbering might make sense, but as soon as we have specialised examples for single features, numbering them is less obvious.


edmorley (approved):
Thank you! The redundant `examples-` prefix on the directory names catches me out every time I tried to tab complete, so glad it's gone :-)

PR #324, opened 2022-02-16
Add exec.d related types

Malax:
Adds exec.d related types, these will be used to implement helpers for implementing exec.d programs with libcnb.

See: https://github.com/buildpacks/spec/blob/main/buildpack.md#execd

This isn't a breaking change since the existing `exec_d` module was never referenced in `lib.rs` and therefore neither exposed nor compiled.

### DIFF
--- a/libcnb-data/src/exec_d.rs
+++ b/libcnb-data/src/exec_d.rs
@@ -39,7 +39,7 @@ libcnb_newtype!(
     /// let key: ExecDProgramOutputKey = exec_d_program_output_key!("PATH");
     /// ```
     exec_d_program_output_key,
-    /// A key of from exec.d program output
+    /// A key of exec.d program output
     ///
     /// It MUST only contain numbers, letters, and the characters `_` and `-`.
     ///
--- a/libcnb-data/src/exec_d.rs
+++ b/libcnb-data/src/exec_d.rs
@@ -2,6 +2,9 @@ use crate::newtypes::libcnb_newtype;
 use serde::Serialize;
 use std::collections::HashMap;
 
+/// Output of a CNB exec.d program.
+///
+/// See [Cloud Native Buildpack specification](https://github.com/buildpacks/spec/blob/main/buildpack.md#execd)
 #[derive(Serialize, Clone)]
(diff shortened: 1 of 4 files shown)


libcnb-data/src/exec_d.rs
```diff
@@ -1,5 +1,111 @@
...
+    ///
+    /// # Examples:
+    /// ```
+    /// use libcnb_data::exec_d::ExecDProgramOutputKey;
+    /// use libcnb_data::exec_d_program_output_key;
+    ///
+    /// let key: ExecDProgramOutputKey = exec_d_program_output_key!("PATH");
+    /// ```
+    exec_d_program_output_key,
+    /// A key of from exec.d program output
```
edmorley, line 42: "key of from" seems like a typo?

Should this rustdoc either link to https://github.com/buildpacks/spec/blob/main/buildpack.md#execd-output-toml or include a bit more detail? From reading the rustdoc at the moment I'd have no idea what this should be used for, without also having read the spec.

libcnb-data/src/exec_d.rs
```diff
@@ -1,5 +1,111 @@
+use crate::newtypes::libcnb_newtype;
 use serde::Serialize;
 use std::collections::HashMap;
 
-#[derive(Serialize)]
-pub type ExecD = HashMap<String, String>;
+#[derive(Serialize, Clone)]
```
edmorley, line 5: Should `ExecDProgramOutput` have a rustdoc too?

libcnb-data/src/exec_d.rs
```diff
@@ -1,5 +1,111 @@
+use crate::newtypes::libcnb_newtype;
 use serde::Serialize;
 use std::collections::HashMap;
 
-#[derive(Serialize)]
-pub type ExecD = HashMap<String, String>;
```
edmorley, line 5: Could you mention in the PR description that the reason this isn't a breaking change, is that `ExecD` was never compiled/available originally?

edmorley (approved):

libcnb-data/src/exec_d.rs
Malax: Typo fixed, link to spec added to `ExecDProgramOutput`.

PR #325, opened 2022-02-16
Improve crate metadata used on crates.io

edmorley:
* Updates crate descriptions now that we have several crates.
* Adjusts `repository` for the ancillary crates to deep link to their directory in the monorepo, rather than the repository root. This may help with tools like Dependabot, which parse these URLs to find out whether a dependency exists in a subdirectory or not.
* Adds explicit `documentation` links since otherwise docs links are  missing from search results, due to rust-lang/crates.io#1484.
* Adds `keywords` for all user-facing crates (ie everything but the proc-macros crate, since that isn't really buildpacks/CNB specific).


README.md
```diff
@@ -9,7 +9,7 @@
 [Rustc Version 1.56+]: https://img.shields.io/badge/rustc-1.56+-lightgray.svg
 [rustc]: https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html
 
-`libcnb.rs` is a Rust framework for writing [Cloud Native Buildpacks](https://buildpacks.io) in Rust. It is an opinionated implementation adding language constructs and convenience methods for working with the spec. It values strong adherence to the spec and data formats.
```
Malax, line 12: Should we mention it's written in Rust? :P

Malax (approved):

PR #331, opened 2022-02-21
Prevent unnecessary runtime validation for newtype generated macros

### LINKED ISSUE #330: Newtype macros for compile-time validation also perform validation at runtime
The `libcnb_newtype!` macro allows defining a newtype backed by `String` that is validated by a regular expression.

There are two ways to then use this type:
1. Via the generated macro that performs compile-time validation using the regex
2. Via runtime parsing/validation thanks to the `FromStr` implementation

As found in https://github.com/Malax/libcnb.rs/pull/326#issuecomment-1043270814 it turns out that when (1) is used, the string is validated both at compile time, and at runtime. This:
- is redundant, since the string has already been validated at compile time
- means `fancy-regex` and its subdependencies are required at runtime, causing a significant increase in the example exec.d binary size (0.4MB -> 1.7MB in a release+stripped binary)

To resolve this, the compile-time validation macro needs to be changed to no longer use `.parse()`:
https://github.com/Malax/libcnb.rs/blob/bca7f027214f5f84b6b46c2356f4023d6ecdcb23/libcnb-data/src/newtypes.rs#L137-L143

However the string backing the newtype is currently private (by design), and since the compile-time macro expands in user code, trying to use it as is gives an error. One way to avoid this, whilst also trying to discourage manual creation of these types would be to add a function like `.new_unchecked()` (for which there seems to be precedence, eg https://doc.rust-lang.org/std/index.html?search=new_unchecked) that would be harder for end-users to accidentally use than `.new()` or making the tuple struct contents public.

GUS-W-10727850.

edmorley:
Adjusts the newtype generated compile-time validation macros so that they don't perform redundant validation at runtime, given that the value was already validated at compile time.

In cases where only compile-time validation is being performed (for example `exec.d` scripts), this results in significant reduction in binary size (1.7MB -> 0.4MB in a release+stripped binary), due to no longer needing the `fancy-regex` dependency at runtime.

A `new_unchecked()` method was added, so that the types (whose backing string is intentionally private) can be constructed by the code generated by the macro (which will exist outside of the module). The name was based on:
https://doc.rust-lang.org/std/index.html?search=new_unchecked

Fixes #330.
GUS-W-10727850.

### DIFF
--- a/libcnb-data/CHANGELOG.md
+++ b/libcnb-data/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add `#[must_use]` to `BuildPlan` and `BuildPlanBuilder` ([#288](https://github.com/Malax/libcnb.rs/pull/288)).
 - Add `exec_d` module with types representing the output of an `exec.d` program ([#324](https://github.com/Malax/libcnb.rs/pull/324)).
 - Increase minimum supported Rust version from 1.56 to 1.58 ([#318](https://github.com/Malax/libcnb.rs/pull/318)).
+- Adjust newtype generated compile-time validation macros so that they don't also perform redundant validation at runtime. In cases where only compile-time validation is being performed (for example `exec.d` scripts), this results in a significant reduction in binary size. ([#331](https://github.com/Malax/libcnb.rs/pull/331))
 
 ## [0.4.0] 2022-01-14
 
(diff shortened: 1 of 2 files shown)


Malax (approved):
Nice. :)

PR #342, opened 2022-02-22
Consistently use `tests` as the module name for tests (again)

edmorley:
To be consistent with our other tests, and so people copying the examples end up using the idiomatic naming too (which is `tests`):
https://doc.rust-lang.org/book/ch11-03-test-organization.html#unit-tests
https://doc.rust-lang.org/rust-by-example/testing/unit_testing.html

Fixes new instances added since #191.


edmorley:
@Malax Out of curiosity, do you hand type these test blocks, or use an IDE template? I wondered if it might be the latter that was using `test`, but it seems the IntelliJ Rust plugin does use `tests` (presuming this is the plugin you are using)?
https://github.com/intellij-rust/intellij-rust/blob/42deeaea73cb677c2f7f9bf2e5a7ae4191ca5723/src/main/resources/org/rust/ide/liveTemplates/test.xml#L45-L46

Malax:
@edmorley I don't use a template for that - I just write it out quickly. :)

Malax (approved):

PR #344, opened 2022-02-23
Set `publish = false` and `version = "0.0.0"` in `Cargo.toml` of examples

### LINKED ISSUE #341: Use `publish = false` and `version = "0.0.0"` in `Cargo.toml` of examples
The majority of buildpacks crates will not need to be published to crates.io, since:
1. They do not have a public library
2. Or they do have a public library, but it's part of a separate crate in the monorepo, so the buildpack crate itself still isn't a library

As such, it makes sense to add `publish = false` and also zero out the `version` (to reduce confusion with the `buildpack.toml` version) in the `Cargo.toml` of our examples.

See:
https://github.com/heroku/procfile-cnb/pull/38

Malax:
Closes #341

### DIFF
--- a/examples/basics/Cargo.toml
+++ b/examples/basics/Cargo.toml
@@ -1,8 +1,9 @@
 [package]
 name = "examples-basics"
-version = "0.1.0"
+version = "0.0.0"
 edition = "2021"
 rust-version = "1.58"
+publish = false
 
 [dependencies]
 libcnb = { path = "../../libcnb" }
--- a/examples/execd/Cargo.toml
+++ b/examples/execd/Cargo.toml
@@ -1,8 +1,9 @@
 [package]
 name = "examples-execd"
-version = "0.1.0"
+version = "0.0.0"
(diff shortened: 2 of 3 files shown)


examples/basics/Cargo.toml
```diff
@@ -1,8 +1,9 @@
 [package]
 name = "examples-basics"
-version = "0.1.0"
+version = "0.0.0"
```
edmorley, line 3: Do we also want the explanation?
https://github.com/heroku/procfile-cnb/pull/38/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R3

(Happy either way)

edmorley (approved):

examples/basics/Cargo.toml
Malax: I don't think we really need it. I'll go ahead and merge, we could add it later if we want to.

PR #346, opened 2022-02-23
Allow setting environment variables in integration tests

Malax:
Adds the ability to set environment variables in `libcnb-test` integration tests for both build and runtime.

This allows us to write integration tests for code paths that are triggered by environment variables. For build-time, this is a non-breaking change since `IntegrationTest` already used a builder-like pattern:

```rust
IntegrationTest::new("heroku/buildpacks:20", "test-fixtures/empty-app")
    .env("FOO", "SOME_VALUE")
    .run_test(|context| {
        // ...
    });
```

For runtime, `start_container` was replaced by `prepare_container` which is now also a builder-like structure. Besides allowing for future expansion when we want to set additional metadata for the container, exposing ports is now optional (users were previously forced to pass at least an empty array to `start_container`):

```rust
context
    .prepare_container()
    .env("PORT", TEST_PORT.to_string())
    .expose_port(TEST_PORT)
    .start(|container| {
      // ...
    });
```

Fixes https://github.com/Malax/libcnb.rs/issues/338, GUS-W-10765593

### DIFF
--- a/libcnb-test/src/container_context.rs
+++ b/libcnb-test/src/container_context.rs
@@ -33,12 +33,40 @@ impl<'a> PrepareContainerContext<'a> {
     }
 
     /// Inserts or updates an environment variable mapping for the container.
+    ///
+    /// # Example
+    /// ```no_run
+    /// use libcnb_test::IntegrationTest;
+    ///
+    /// IntegrationTest::new("heroku/buildpacks:20", "test-fixtures/app").run_test(|context| {
+    ///     context
+    ///         .prepare_container()
+    ///         .env("FOO", "FOO_VALUE")
+    ///         .start(|container| {
+    ///             // ...
+    ///         })
+    /// });
+    /// ```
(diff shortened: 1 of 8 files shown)


libcnb-test/src/container_context.rs
```diff
@@ -1,10 +1,128 @@
...
+        Self {
+            integration_test_context,
+            exposed_ports: Vec::new(),
+            env: HashMap::new(),
+        }
+    }
+
+    /// Exposes a given port of the container to the host machine.
+    ///
+    /// The given `exposed_port` is mapped to random ports on the host machine. Use
```
edmorley, line 28: There is no such variable called `exposed_port` in the signature of the method. Was this meant to be `port` or else for it to use prose ("exposed port") rather than code backticks?

edmorley (approved):
Awesome!

libcnb-test/src/container_context.rs
Malax: Good catch, this is a leftover from a previous version of the docs. I'll fix it. :)

PR #362, opened 2022-02-28
Extract `libcnb-cargo`'s library to a new `libcnb-package` crate

### LINKED ISSUE #332: Move the library part of `libcnb-cargo` to a separate crate
Currently the `libcnb-cargo` crate is both a binary (that powers the cargo plugin), and a library (that's used by both the binary, and separately by `libcnb-test` and any other external consumers).

End-user projects will have `libcnb-test` in their `dev-dependencies`, which itself depends upon the library part of `libcnb-cargo`. Whilst these projects will also use the CLI part, they do so by `cargo install`ing it, which doesn't use the project's `target/` directory.

As such, by having both the CLI and library in the same crate, we make these projects unnecessarily compile the CLI-specific dependencies, such as `clap`, `pathdiff`, `size_format` and `stderrlog` (and all of their transitive dependencies). (Which on `main` is currently 23 extra dependencies; this will go up after #329 since Clap v3 has a few more deps.)

To prevent this, we could move the library parts of `libcnb-cargo` to another crate. Then `libcnb-cargo` would contain only a binary (the cargo plugin), and both it and `libcnb-test` would depend on the new crate.

In addition to reducing the dependency tree, such a move would make it easier to keep track of breaking changes/writing the changelog for the library vs the binary parts of `libcnb-cargo` (for example we've accidentally released breaking library changes in a patch release of `libcnb-cargo` before, because it wasn't a breaking change for the CLI).

Potential names for this crate:
- `libcnb-build`
- `libcnb-compile`
- `libcnb-packaging`
- `libcnb-package`

Note: I've intentionally not considered a `libcnb-common` or similar type name, since the crate will be a development only dependency, so won't want to be in the same crate as non-dev dependencies.

@Malax @hone Thoughts?

edmorley:
The `libcnb-cargo` crate was previously both a binary that users would install via `cargo install`, and a library that was used by that binary, by `libcnb-test`, and also by any other external crates.

Now, the library part of `libcnb-cargo` has been extracted to a new crate, `libcnb-package`, which:

- Means consumers of the library only (such as `libcnb-test`) don't need to compile the 20 CLI-related dependencies, improving compile times.
- Improves separation of concerns:
  - Makes the name of each of the respective crates more accurately reflect their purpose.
  - It's now clearer what changes will affect the CLI vs library use-cases, allowing for finer grained and less error prone major version bumps (for example, `libcnb-cargo@0.2.1` was released as a minor version bump, when it was actually breaking for the library).
- Makes it possible for both `libcnb-cargo` and `libcnb-package` to start using the `unused_crate_dependencies` lint again (which doesn't work well with dual binary/library crates), meaning one less thing we need to worry about in code reviews.

The effect on dependency tree size and compile times (using a project that depends on `libcnb` and `libcnb-test`)...

Before:
```
$ cargo tree --prefix none --no-dedupe | sort -u | wc -l
138
$ cargo clean && cargo check --all-targets
    Finished dev [unoptimized + debuginfo] target(s) in 45.12s
```

After:
```
$ cargo tree --prefix none --no-dedupe | sort -u | wc -l
118
$ cargo clean && cargo check --all-targets
...
    Finished dev [unoptimized + debuginfo] target(s) in 41.76s
```

Closes #332.
GUS-W-10766598.


edmorley:
For alternative names for the new crate, see #332.

Malax (approved):
Main main concern with the name of the crate is what would happen when there is more than just packaging that is shared between `libcnb-test` and `libcnb-cargo`? However, I would not come up with a realistic example at this point. Therefore, I think this is fine.

edmorley:
> what would happen when there is more than just packaging that is shared between `libcnb-test` and `libcnb-cargo`

Yeah true. 

In terms of future functionality we may want to add to libcnb:
- asset downloading/bundling -> seems like it might belong in `libcnb-package` anyway
- buildpack releasing/publishing automation -> seems like it wouldn't be used by `libcnb-test` and only `libcnb-cargo`, so would want to be a separate crate from `libcnb-package` anyway?
- <some feature that's used at runtime> -> wouldn't want to be a dev-dependency, so would have to be a separate crate too (or just in `libcnb` with feature flags)

...so it seems like using `libcnb-package` should be ok. 

Worst case if we do think of additional use-cases, we just add another crate alongside `libcnb-package`, that's used by both `libcnb-test` and `libcnb-cargo`. Since it (and `libcnb-package`) would be transitive dependencies, end-users won't have any additional overhead of extra dependency versions to manage, so it seems less of an issue.

PR #366, opened 2022-03-02
Allow starting a container without using the default process type

Malax:


libcnb-test/Cargo.toml
```diff
@@ -20,3 +20,4 @@ libcnb-cargo = { path = "../libcnb-cargo", version = "0.3.0" }
 tempfile = "3.3.0"
 tokio = "1.17.0"
 tokio-stream = "0.1.8"
+serde = "1.0.136"
```
edmorley, line 23:
```suggestion
serde = "1.0.136"
tempfile = "3.3.0"
tokio = "1.17.0"
tokio-stream = "0.1.8"
```

libcnb-test/src/container_context.rs
```diff
@@ -129,6 +222,71 @@ pub struct ContainerContext<'a> {
...
+    /// Note: This method will block until the container stops. If the container never stops by
+    /// itself, your test will hang indefinitely. This is common when the container hosts an HTTP
+    /// service.
+    ///
+    /// See: [`logs`](Self::logs) for a non-blocking alternative.
+    ///
+    /// # Panics
+    /// - When the log output could not be consumed/read.
+    #[must_use]
+    pub fn logs_follow(&self) -> LogOutput {
```
edmorley, line 263: I'm not 100% sure about the naming of this method, but I don't know if I can think of something better either.

The name suggests that one can actually follow the logs, which isn't the case, and I think might not be intuitive when reading tests calling it.

If I'm understanding correctly this is more like a `logs_blocking()`, or to focus on the desired outcome of calling the function (rather than the side-effect behaviour), `logs_until exit()`? Though both of those names aren't perfect either.

Any other ideas? :-)

libcnb-test/src/container_context.rs
```diff
@@ -154,7 +312,7 @@ impl<'a> ContainerContext<'a> {
     }
 
     /// # Panics
-    pub fn shell_exec(&self, command: impl AsRef<str>) -> ContainerExecResult {
+    pub fn shell_exec(&self, command: impl AsRef<str>) -> LogOutput {
```
edmorley, line 315: Could you add rustdocs that clarifies what this method does? In particular, that it runs inside the existing container, rather than starting a new container?

libcnb-test/src/container_context.rs
```diff
@@ -78,12 +80,101 @@ impl<'a> PrepareContainerContext<'a> {
         self
     }
 
-    /// Creates and starts the container configured by this context.
+    /// Creates and starts the container configured by this context with the image's default
+    /// CNB process.
+    ///
```
edmorley, line 85: I wonder if it would be worth mentioning here (and in the docs for the related methods below) how container teardown is handled?

libcnb-test/CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
...
 - Added the ability to set environment variables for the container via `PrepareContainerContext::env` and `PrepareContainerContext::envs`. ([#346](https://github.com/Malax/libcnb.rs/pull/346))
+- Allow starting a container without using the default process type. Removed `PrepareContainerContext::start`, added `PrepareContainerContext::start_with_default_process`, `PrepareContainerContext::start_with_default_process_args`, `PrepareContainerContext::start_with_process`, `PrepareContainerContext::start_with_process_args`, `PrepareContainerContext::start_with_shell_command` ([#366](https://github.com/Malax/libcnb.rs/pull/366))
+- Add `ContainerContext::logs` and `ContainerContext::logs_follow` to access the logs of the container. Useful when used in conjunction with the new `PrepareContainerContext::start_with_shell_command` method to get the shell command output. ([#366](https://github.com/Malax/libcnb.rs/pull/366))
+- Replaced `container_context::ContainerExecResult` with `log::LogOutput` which is now shared with `ContainerContext::logs` and `ContainerContext::logs_follow`. ([#366](https://github.com/Malax/libcnb.rs/pull/366))
```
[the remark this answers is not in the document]
Malax: Yes. I'll fix up the wording.

libcnb-test/Cargo.toml
Malax: 2d9b20d

libcnb-test/src/container_context.rs
```diff
@@ -78,12 +80,101 @@ impl<'a> PrepareContainerContext<'a> {
         self
     }
 
-    /// Creates and starts the container configured by this context.
+    /// Creates and starts the container configured by this context with the image's default
```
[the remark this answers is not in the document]
Malax: I like `s/with/using/`! 👍🏻

libcnb-test/src/container_context.rs
```diff
@@ -78,12 +80,101 @@ impl<'a> PrepareContainerContext<'a> {
...
+    ///
+    /// # Panics
+    /// - When the container could not be created
+    /// - When the container could not be started
+    pub fn start_with_default_process<F: FnOnce(ContainerContext)>(&self, f: F) {
+        self.start_internal(None, None, f);
+    }
+
+    /// Creates and starts the container configured by this context with the image's default
+    /// CNB process.
```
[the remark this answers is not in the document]
Malax: Good point - I think I wanted to do that and then forgot. :) 43fac40

libcnb-test/src/container_context.rs
Malax: I don't have a strong opinion on this other than this is IMO out of scope for this PR since the way we do teardown didn't change. If we want this as RustDoc, we can do a separate PR or issue. :)

libcnb-test/src/container_context.rs
Malax: I went with `follow` as this is the Docker terminology on both the CLI and API. The fact that it blocks and doesn't allow real-time streaming of the logs is IMO well conveyed through the types (since this doesn't return a `Stream`).

I would leave it as-is unless you feel strongly the name should be changed.

libcnb-test/src/container_context.rs
edmorley: Ah true. I'll file an issue for it. However I'd love for us to start making progress though the long tail of follow-up issues, and not just the big feature PRs at some point too :-)

libcnb-test/src/container_context.rs
Malax: I added some basic docs in f209796.

libcnb-test/src/container_context.rs
Malax: > and not just the big feature PRs at some point too :-)

I guess that's a matter of priorities. Missing features block our progress with the actual buildpacks, doc improvements do not (granted that better docs would reduce friction). :-) But I think we're close to the point where the actual blockers are gone. :)

libcnb-test/src/container_context.rs
Malax: @edmorley do you want to propose a name?  As you said,`logs_blocking` and `logs_until_exit` aren't amazing either.

edmorley: I'm also curious - in what scenarios will `.logs()` not include all logs, for containers that do exit? From reading `start_internal()` it seems we block on the `create_container` and `start_container` steps, so presumably by the time the test-provided function runs, the container will have at least started? But I presume the issue is that it's not guaranteed that the started command will have finished running and outputting to stdout/stderr by the time the test block runs?

It just seems like `.logs()` is perhaps something that should rarely be used (unless querying the logs of a web server, and even then careful use of sleeps etc will be needed), and instead `.logs_follow()` should be preferred in most cases?

If so, then I worry that having the names as `.logs()` and `.logs_follow()` doesn't make it clear that the former is a often a footgun and the latter is often the best practice, since:
- To me, when there are several methods with similar names, some with suffixes and some without, then the ones without the suffixes seem like the standard/usual case.
- "Follow" makes me think it's some interactive thing, therefore not what I need, therefore the "other" method is probably what I need.

Now obviously reading the rustdocs does make things clearer, however when people are say reviewing a PR via the GitHub code interface, I think it would be very easy to forget that a `.logs()` usage in a test is likely a footgun.

What about something like:
- `logs_follow()` -> `logs_wait()`
- `logs()` -> `logs_now()`

Or even this, if we see "follow" as being the safer (as in, less prone to races)/more common usage:
- `logs_follow()` -> `logs()`
- `logs()` -> `logs_now()`

(I'm not overly attached to the exact names, more-so the idea of not having the footgun method be the one without the suffix)

Malax: > Or even this, if we see "follow" as being the more common usage

Thing is, `logs_follow` isn't without it's own foot-guns either. If you call it on a container reference that runs a never-terminating process such as a web service (which isn't uncommon), that call will hang indefinitely without any indication what is going on.

With `logs`, the assert will fail when the test runs into timing issues, but that will at least directly point to the assert as the source of the issue.

I'd say which of the two you want to use is really dependent on the use-case: if it's a long running process, you probably want `logs` with some extra timing logic. If we're talking about a one-off command,`logs_follow` is easier to use since the test doesn't need any custom timing logic. There is no "default".

Honestly, I still don't think the naming of the log functions is bad and/or misleading. The naming and behaviour exactly mirrors the Docker CLI when getting logs from a running container and the Docker API (which is used programatically, not as an interactive process) uses the same naming too.

Even if there is a slightly better name, we always pay the price of using a different name than the underlying (and not abstracted away) technology. I feel like we already spent way too much time on this minor naming discussion which will probably have no measurable impact - I'm sure people will run into the timing/blocking issues with either name when they don't read the docs that mention the caveats.

Malax: That said, to cut this short I'm willing to rename them as proposed if that ends this bikeshedding thread:

```
logs_follow() -> logs_wait()
logs() -> logs_now()
```

edmorley: > If you call it on a container reference that runs a never-terminating process such as a web service (which isn't uncommon), that call will hang indefinitely without any indication what is going on.

I agree, but this issue is (a) deterministic (will hang every time), (b) will be visible whilst writing/developing the test, not at some random point later.

> I'd say which of the two you want to use is really dependent on the use-case: if it's a long running process, you probably want `logs` with some extra timing logic.

I absolutely agree, I was just wondering if this case was less common.

> Honestly, I still don't think the naming of the log functions is bad and/or misleading. The naming and behaviour exactly mirrors the Docker CLI when getting logs from a running container and the Docker API (which is used programatically, not as an interactive process) uses the same naming too.

So whilst sometimes looking at the underlying terminology can be useful (ie: we don't want to make concepts up when adding layers of abstraction, otherwise it can add to the mental overhead to understand the overall model), I think in this case, the users of libcnb-test aren't thinking in terms of the Docker CLI or API, but a test runner letting them make assertions in a testcase. I think that changes things slightly.

> That said, to cut this short I'm willing to rename them as proposed if that ends this bikeshedding thread:

Thank you for making the change. However I think it's a little unfair to call a genuine concern about users mis-using an API due to naming (which can result in either hangs or flaky tests; this isn't just a vanity naming thing), merely "bikeshedding" :-)

libcnb-test/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
...
 - Use `--pull-policy if-not-present` when running `pack build` ([#373](https://github.com/Malax/libcnb.rs/pull/373)).
+- Allow starting a container without using the default process type. Removed `PrepareContainerContext::start`, added `PrepareContainerContext::start_with_default_process`, `PrepareContainerContext::start_with_default_process_args`, `PrepareContainerContext::start_with_process`, `PrepareContainerContext::start_with_process_args`, `PrepareContainerContext::start_with_shell_command` ([#366](https://github.com/Malax/libcnb.rs/pull/366))
+- Add `ContainerContext::logs` and `ContainerContext::logs_follow` to access the logs of the container. Useful when used in conjunction with the new `PrepareContainerContext::start_with_shell_command` method to get the shell command output. ([#366](https://github.com/Malax/libcnb.rs/pull/366))
+- Replaced `container_context::ContainerExecResult` with `log::LogOutput` which is now also returned by `ContainerContext::logs` and `ContainerContext::logs_follow`. ([#366](https://github.com/Malax/libcnb.rs/pull/366))
```
edmorley, line 12: The method names need updating here for the rename

edmorley (approved):

libcnb-test/CHANGELOG.md
edmorley: d28965e911ebc9a88b36855e355c686dbcc622df

schneems:
FWIW this was a meh upgrade experience. I got this error:

```
error[E0599]: no method named `start` found for mutable reference `&mut PrepareContainerContext<'_>` in the current scope
  --> tests/integration_test.rs:23:18
   |
23 |                 .start(|container| {
   |                  ^^^^^ method not found in `&mut PrepareContainerContext<'_>`
```

And so I check the readme which shows this code:

```
// In $CRATE_ROOT/tests/integration_test.rs
use libcnb_test::{IntegrationTest, BuildpackReference, assert_contains};

#[test]
fn test() {
    IntegrationTest::new("heroku/buildpacks:20", "test-fixtures/app")
        .buildpacks(vec![
            BuildpackReference::Other(String::from("heroku/openjdk")),
            BuildpackReference::Crate,
        ])
        .run_test(|context| {
            assert_contains!(context.pack_stdout, "---> Maven Buildpack");
            assert_contains!(context.pack_stdout, "---> Installing Maven");
            assert_contains!(context.pack_stdout, "---> Running mvn package");

            context.prepare_container().expose_port(12345).start(|container| {
                assert_eq!(
                    call_test_fixture_service(
                        container.address_for_port(12345).unwrap(),
                        "Hagbard Celine"
                    )
                    .unwrap(),
                    "enileC drabgaH"
                );
            });
        });
}

fn call_test_fixture_service(addr: std::net::SocketAddr, payload: &str) -> Result<String, ()> {
   unimplemented!()
}
```

So it looks like start method does exist, but there is some other problem.

I spelunk to find the context.rs file and see this change as the most recent history. Then kinda intuit from the commit name and the all the `start_` methods that I should use `start_with_default_process` instead. 

My thoughts:

- Readme needs to be tested as this is invalid code that will not compile
- Just because we're statically typed doesn't mean interface changes won't be intuitive. I'm wanting some kind of a "this is deprecated, use this instead" experience which I would expect from **even** a language like Ruby. Granted there you have to manually do it, and not everyone does. Maybe there are some better UX patterns or macros or crates or something that exist for when we rename something (not just methods, structs and implementations as well). 


Overall it's not a HUGE deal. Just wanted to give stream-of-concious and in-the-moment feedback as I think it's valuable. Whether we choose to act on it or not.

edmorley:
@schneems Thank you for the feedback!

Agree the libcnb-test README should be fixed - the code examples in repo root README are already tested in CI - this one is supposed to be too, but maybe the `[test]` annotation is interfering with that? I've filed #403.

These breaking changes were mentioned in the changelog (and the changelog gets inlined into Dependabot PRs for example, though you won't have such a PR since the Ruby Rust changes are not merged). However I agree this requires someone reading the changelog, so best if there is a more automatic solution.

Rust offers the ability to show such a warning, which we could use for future breaking changes:
https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

That all said, it's worth bearing in mind that much of libcnb (particularly libcnb-test) is still at the very early pre-v1.0.0 stage (our team's buildpacks are the early adopters/guinea pigs for them), so we expect there to be more breaking changes/rough edges now, than there will be in the future. The changelog is there to try and make this slightly less confusing, so I would recommend reading it in the future :-)

schneems:
> That all said, it's worth bearing in mind that much of libcnb (particularly libcnb-test) is still at the very early pre-v1.0.0 stage (our team's buildpacks are the early adopters/guinea pigs for them), so we expect there to be more breaking changes/rough edges now, than there will be in the future. The changelog is there to try and make this slightly less confusing, so I would recommend reading it in the future :-)

I think there's two components for me 1) is a pure technical "can we do this" and the other is the implementation side. You've answered the "can we do this" thank you. 

I think the second component is more of a cultural "should we do this". I certainly understand the cost/time tradeoff. I personally fall into habits where I "play" much like I "practice". I.e. I tend to do things always or never. If I don't make a habit of going through deprecation cycles then when it's really critical to do so I'll likely forget. I believe you have more self control as well as awareness and also are good at reviewing PRs. I need more black and white policies as well as automated tooling.

I do want to push back on the "it's just us" narrative. While that's true for now, the longer we act that way, the more hostile the codebase will be to adoption by others (self fulfilling prophecy). Also even "just us" is a hefty cost/time tradeoff. I spent maybe 60 min (perhaps rounding up) investigating this. If we multiply by number of team mebers then taking an extra hour to reduce multiplicative time spent might not always come out to be the best investment but it's an investment and does have payback.

The other thing of value I gain from choosing to use deprecation cycles is that it informs the implementation and also forces us to think about "after this PR is merged how will we deprecate or remove this code if it's no longer needed". In the same way I believe in shipping tests with code as it's written as the tests will inform the API. Not so dissimilar to making plans to deprecate and remove a stack before it's even released. 

Coming from Ruby the main thing's I'm hoping to gain from Rust are improved maintainability and part of that is library/ecosystem deprecations.

I don't think we disagree in a large number of these areas I am somewhat typing to get my thoughts out and clear on the subject. I'm curious in talking more about the path forward for some of this stuff regarding when a good time to flip the switch from "not now" to "now" might be. Alternatively, maybe we could make me the "deprecation czar" (or something) and I can do some of the legwork of trying to figure out the deprecation/upgrade path on breaking changes.

PR #367, opened 2022-03-02
Update README.md

schneems:
Putting this in the `test` directory will not trigger runs via:

```
$ RUST_BACKTRACE=1 cargo test --all-features --locked -- --include-ignored
```

It needs to be `tests` (with an `s` a the end)


Malax (approved):
Good catch! Thanks!

PR #368, opened 2022-03-02
Fix compilation on non-`unix` family systems

nokome:
Previously, there were a couple of pieces of code that failed to compile on Windows e.g. `use std::os::unix::ffi::OsStringExt;`.

This PR adds `#[cfg(target_family = "unix")]` attributes to avoid those compile errors and where necessary implements analogous functionality for non-unix platforms (checked using cross-compile on Linux).

Note: This commit has been extracted out from https://github.com/Malax/libcnb.rs/pull/313 (which will be superseded by this an another PR).

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -54,7 +54,6 @@ pub fn libcnb_runtime<B: Buildpack>(buildpack: &B) {
         .and_then(Path::file_name)
         .and_then(OsStr::to_str);
 
-    #[cfg(any(target_family = "unix"))]
     let result = match current_exe_file_name {
         Some("detect") => libcnb_runtime_detect(buildpack),
         Some("build") => libcnb_runtime_build(buildpack),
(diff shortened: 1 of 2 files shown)


libcnb/src/runtime.rs
```diff
@@ -54,7 +54,6 @@ pub fn libcnb_runtime<B: Buildpack>(buildpack: &B) {
         .and_then(Path::file_name)
         .and_then(OsStr::to_str);
 
-    #[cfg(any(target_family = "unix"))]
```
Malax, line 57: I think removing this to allow compilation on non-unix systems is fine. However, since we're matching the unix specific binary names, this would never work on i.e. Windows where `detect` would be `detect.exe` for example.

Since we don't support Windows buildpacks this isn't a huge deal, but maybe we can add a comment here to remind our future selves about this?

Malax: Actually, I think don't even need a comment here. Adding Windows support would require us to go over everything anyway and the executable name matching will be probably the most obvious thing to fix. :)

Malax (approved):

PR #369, opened 2022-03-02
Centralize arg handling, writing to stderr and process exit in `libcnb_runtime`

nokome:
This change takes a "Functional Core - Imperative Shell" type of approach by centralizing argument handling, writing of error messages to stderr and calls to `process::exit` in the main `libcnb_runtime` function.

This allows for other functions such as `libcnb_runtime_detect` and `libcnb_runtime_build` to be re-used in other runtimes.

Note: This commit has been extracted out from https://github.com/Malax/libcnb.rs/pull/313 (which will be superseded by this an other PRs).

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -11,7 +11,6 @@ use std::env;
 use std::ffi::OsStr;
 use std::fmt::Debug;
 use std::path::{Path, PathBuf};
-use std::process;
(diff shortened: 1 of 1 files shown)


Malax (approved):
Thanks for the PR and the split-up! 👍🏻  I took the liberty and replaced the `&str` error type with an enum so the runtime output is concentrated in one place.

PR #373, opened 2022-03-06
Use `--pull-policy if-not-present` when running `pack build`

### LINKED ISSUE #306: Use `--pull-policy if-not-present` with Pack CLI in libcnb-test
By default (if the global config hasn't been updated), Pack CLI will attempt to pull the builder image every time a `pack build` is performed. Even if this image is up to date, it still:
- adds ~2 seconds to the `pack build` time
- counts against the container registry rate limits (ie Docker Hub or ECR)

This is not ideal when running `pack build` multiple times in a short period of time - as happens when running `libcnb-test`.

Currently to prevent around this in CI, we enable the `if-not-present` pull policy via Pack global config:
https://github.com/Malax/libcnb.rs/blob/552edb954cfc069814141baf4945921edb4e7a47/.github/workflows/ci.yml#L75-L80

However:
1. This means every project using libcnb.rs (a) has to remember to set this in their CI (which is easy to forget, eg https://github.com/heroku/procfile-cnb-rust/pull/24), (b) now has extra boilerplate in their CI config.
2. Solving for CI doesn't help usages locally, where users (a) might not know about the pull policy config, (b) might not want to set it globally since it affects all other pack usages outside of integration tests too.

One way to avoid both of these issues, would be if `libcnb-test` were to pass `--pull-policy if-not-present` to `pack build` as part of running the integration tests.

One downside to doing that, would be that if a user runs `libcnb-test` powered integration tests on a machine that has old builder images, then they would never get updated unless the user is also running `pack build` outside of integration tests too. However (a) users will presumably be running `pack build` reasonably often, (b) if there are any builder image changes that cause issues then CI would catch it anyway (since it will be pulling in new images each time, since none will be present).

Ideally Pack CLI would support an `if-not-pulled-recently` option, which would skip pulling if that builder image was pulled recently - and so give the best of both worlds. However in lieu of having`if-not-pulled-recently` (we should file an issue), I think `libcnb-test` should pass `--pull-policy if-not-present` to Pack for now, as it's the least worst of the two approaches.

edmorley:
Since it:
1. Saves ~2 seconds per integration test re-pulling an image that's already up to date.
2. Helps prevent hitting Docker Hub or ECR rate limits from duplicate (and redundant) image pulling (that counts agains the rate limit even when it's a no-op).

Longer term, if Pack CLI supports a periodic pulling mode (buildpacks/pack#1368), we can switch to that, however for now this is the lesser of two evils - and in most cases Pack usage outside of `libcnb-test` will ensure that newer builder images are pulled from time to time.

See #306 for more details.

See also:
https://buildpacks.io/docs/tools/pack/cli/pack_config_pull-policy/

Fixes #306.
GUS-W-10799284.

### DIFF
--- a/libcnb-test/src/pack.rs
+++ b/libcnb-test/src/pack.rs
@@ -7,10 +7,11 @@ use tempfile::TempDir;
 #[derive(Clone, Debug)]
 pub(crate) struct PackBuildCommand {
     builder: String,
-    path: PathBuf,
-    image_name: String,
     buildpacks: Vec<BuildpackReference>,
     env: BTreeMap<String, String>,
+    image_name: String,
+    path: PathBuf,
+    pull_policy: PullPolicy,
     verbose: bool,
 }
 
@@ -38,18 +39,32 @@ impl From<String> for BuildpackReference {
     }
 }
 
+#[derive(Clone, Debug)]
+/// Controls whether Pack should pull images.
+#[allow(dead_code)]
+pub(crate) enum PullPolicy {
+    /// Always pull images.
+    Always,
+    /// Use local images if they are already present, rather than pulling updated images.
+    IfNotPresent,
+    /// Never pull images. If the required images are not already available locally the pack command will fail.
+    Never,
+}
+
 impl PackBuildCommand {
     pub fn new(
         builder: impl Into<String>,
         path: impl Into<PathBuf>,
         image_name: impl Into<String>,
+        pull_policy: PullPolicy,
     ) -> PackBuildCommand {
         PackBuildCommand {
             builder: builder.into(),
-            path: path.into(),
-            image_name: image_name.into(),
-            buildpacks: vec![],
+            buildpacks: Vec::new(),
             env: BTreeMap::new(),
+            image_name: image_name.into(),
+            path: path.into(),
+            pull_policy,
             verbose: false,
         }
     }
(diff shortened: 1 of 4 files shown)


joshwlewis (approved):

joshwlewis:
This probably means consumers should run a `docker pull builder-image` in CI prior to starting a suite of integration tests. Is there anything this library could/should do to help with that?

edmorley:
CI won't have any cached images (unless using the paid circle ci docker layer caching feature, which we'd never want to use with pack), so will always pull the latest image in the first test run in a job anyway. This PR just means subsequent tests in that job don't needlessly pull again after that.

libcnb-test/src/pack.rs
```diff
@@ -76,6 +76,9 @@ impl From<PackBuildCommand> for Command {
             pack_build_command.builder,
             String::from("--path"),
             pack_build_command.path.to_string_lossy().to_string(),
+            // Adjust pull-policy to prevent redundant image-pulling, which slows CI and risks hitting registry rate limits.
+            String::from("--pull-policy"),
+            String::from("if-not-present"),
```
Malax, line 81: Nit-ish: `PackBuildCommand` was meant to abstract away the building of the `pack` command from out actual logic which flags should be passed, etc.

I would love to see a `pull_policy` field on `PackBuildCommand` that is of type `PullPolicy` and the "logic" (even though it a static value at this point) moved to the call-site to keep the separation of concerns.

Malax (changes requested):

Malax:
Marking this as "breaking change". Although unlikely, if users rely on automatic pulling to keep up-to-date, this will break them. I don't think this shouldn't be a huge concern for us at this point though.

Malax (approved):
:shipit: Thanks!

PR #382, opened 2022-03-10
BuildPlan Metadata Support

### LINKED ISSUE #350: `BuildPlanBuilder` doesn't provide an API for setting `requires.metadata`
In the build plan, `requires` entries can have associated `metadata`:
https://github.com/buildpacks/spec/blob/main/buildpack.md#build-plan-toml

Whilst the `Requires` type does have a `metadata` field:
https://github.com/Malax/libcnb.rs/blob/0ca9e3ed828f174633367c748c33164331513cf9/libcnb-data/src/build_plan.rs#L117

...`BuildPlanBuilder::require()` doesn't currently support setting it:
https://github.com/Malax/libcnb.rs/blob/0ca9e3ed828f174633367c748c33164331513cf9/libcnb-data/src/build_plan.rs#L53-L56

And it's not possible to set it by hand since the fields on `BuildPlanBuilder` are (understandably) private.

One option might be to have `requires` accept a second `Option` argument. However that will mean anyone who isn't passing metadata will need to pass `None` as the second argument.

Another option would be to add another method, eg `BuildPlanBuilder::require_with_metadata()`, to support this use-case without making the UX worse in the presumably more common no-metadata case.

hone:
This fixes #350.

### DIFF
--- a/libcnb-data/src/build_plan.rs
+++ b/libcnb-data/src/build_plan.rs
@@ -37,8 +37,8 @@ impl BuildPlanBuilder {
         self
     }
 
-    pub fn requires(mut self, name: impl AsRef<str>) -> Self {
-        self.current_requires.push(Require::new(name.as_ref()));
+    pub fn requires(mut self, require: impl Into<Require>) -> Self {
+        self.current_requires.push(require.into());
         self
     }
 
@@ -123,10 +123,22 @@ impl Require {
 
             Ok(())
         } else {
-            Err(toml::ser::Error::Custom(
-                "Could not be serialized as a TOML Table.".to_string(),
-            ))
+            Err(toml::ser::Error::Custom(String::from(
+                "Could not be serialized as a TOML Table.",
+            )))
         }
+    }
+}
+
+impl From<String> for Require {
+    fn from(s: String) -> Require {
+        Require::new(s)
+    }
+}
+
+impl From<&str> for Require {
+    fn from(s: &str) -> Require {
+        Require::new(s)
     }
 }
 
@@ -152,13 +164,13 @@ mod tests {
 
         let mut require = Require::new("foo");
         let metadata = Metadata {
-            foo: "bar".to_string(),
+            foo: String::from("bar"),
         };
         let result = require.metadata(metadata);
         assert!(result.is_ok());
         assert_eq!(
             require.metadata.get("foo"),
-            Some(&toml::Value::String("bar".to_string()))
+            Some(&toml::Value::String(String::from("bar")))
         );
     }
 }
--- a/libcnb-data/src/build_plan.rs
+++ b/libcnb-data/src/build_plan.rs
@@ -130,16 +130,14 @@ impl Require {
     }
 }
 
-impl From<String> for Require {
-    fn from(s: String) -> Require {
-        Require::new(s)
+impl<S: Into<String>> From<S> for Require {
+    fn from(s: S) -> Self {
+        Require {
+            name: s.into(),
+            metadata: Default::default(),
+        }
     }
 }
-
-impl From<&str> for Require {
-    fn from(s: &str) -> Require {
-        Require::new(s)
-    }
 }
 
 #[cfg(test)]
--- a/libcnb-data/src/build_plan.rs
+++ b/libcnb-data/src/build_plan.rs
@@ -42,7 +42,7 @@ impl BuildPlanBuilder {
         self
     }
 
-    /// Add Require to the BuildPlan
+    /// Add Require to the `BuildPlan`
     pub fn requires_with_struct(mut self, require: Require) -> Self {
         self.current_requires.push(require);
         self
@@ -113,6 +113,10 @@ impl Require {
     }
 
     /// Convert a Serializable struct and store it as a toml Table for metadata
+    ///
+    /// # Errors
+    /// This will return error for any normal TOML serialization error as well if it's not
+    /// possible to serialize as a TOML Table.
     pub fn metadata<T: Serialize>(&mut self, metadata: T) -> Result<(), toml::ser::Error> {
         if let toml::Value::Table(table) = toml::Value::try_from(metadata)? {
             self.metadata = table;
@@ -120,7 +124,7 @@ impl Require {
             Ok(())
         } else {
             Err(toml::ser::Error::Custom(
-                "Could not be serialized as a map.".to_string(),
+                "Could not be serialized as a TOML Table.".to_string(),
             ))
         }
     }
@@ -155,6 +159,6 @@ mod tests {
         assert_eq!(
             require.metadata.get("foo"),
             Some(&toml::Value::String("bar".to_string()))
-        )
+        );
     }
 }
--- a/libcnb-data/src/build_plan.rs
+++ b/libcnb-data/src/build_plan.rs
@@ -42,12 +42,6 @@ impl BuildPlanBuilder {
         self
     }
 
-    /// Add Require to the `BuildPlan`
-    pub fn requires_with_struct(mut self, require: Require) -> Self {
-        self.current_requires.push(require);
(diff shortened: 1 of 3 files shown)


libcnb-data/src/buildpack_plan.rs
```diff
@@ -16,6 +16,17 @@ pub struct Entry {
...
 }
 
+impl Entry {
+    /// Deserializes Metadata to a type T that implements Deserialize
+    pub fn metadata<'de, T>(&self) -> Result<T, toml::de::Error>
+    where
+        T: Deserialize<'de>,
+    {
+        // serde::de::Deserializer is not implemented for toml::map::Map, so need to clone() here
+        toml::Value::Table(self.metadata.clone()).try_into()
```
hone, line 26: @Malax not sure we can remove `clone()` here. :(

libcnb-data/src/build_plan.rs
```diff
@@ -42,6 +42,12 @@ impl BuildPlanBuilder {
         self
     }
 
+    /// Add Require to the `BuildPlan`
+    pub fn requires_with_struct(mut self, require: Require) -> Self {
```
Malax, line 46: I find the `requires_with_struct` name a little clunky. Maybe we can change the `requires` signature to this instead:

```rust
pub fn requires<R: Into<Require>>(mut self, require: R) -> Self {
    self.current_requires.push(require.into());
    self
}
```

and then provide `From<String> for Require` and `From<&str> for Require`. Since `From` is reflexive (`From<T> for T` is always available), this would allow the user to use either the full struct or just a string in `requires` without an additional function that is hard to name.

```rust
impl<S: Into<String>> From<S> for Require {
    fn from(s: S) -> Self {
        Require {
            name: s.into(),
            metadata: Default::default(),
        }
    }
}

impl<S: Into<String>> From<S> for Provide {
    fn from(s: S) -> Self {
        Provide { name: s.into() }
    }
}
```

WDYT?

libcnb-data/src/buildpack_plan.rs
Malax: Since `try_into` consumes `self` (which is `self.metadata` in the context of this function) and `fn metadata` only has a reference to `self`, this seems unlikely to be possible.

Nit: the comment is therefore slightly incorrect. We don't need to clone because `Deserialize` isn't implemented, we need to clone because `try_into` will consume the value that is a field of `self`.

libcnb-data/src/buildpack_plan.rs
```diff
@@ -50,4 +64,24 @@ name = "rust"
...
+        metadata.insert("foo".to_string(), toml::Value::String("bar".to_string()));
+        let entry = Entry {
+            name: "foo".to_string(),
+            metadata,
+        };
+
+        let result = entry.metadata::<Metadata>();
+        assert!(result.is_ok());
+        let metadata = result.unwrap();
+        assert_eq!(metadata.foo, "bar".to_string());
```
Malax, line 85: Nit (doesn't need to be changed): This could be written like so if we change the attribute of `Metadata` to `#[derive(Deserialize, Eq, PartialEq, Debug)]`:

```rust
 assert_eq!(
    entry.metadata(),
    Ok(Metadata {
        foo: String::from("bar")
    })
);
```

Test failures would be little easier to read since we would see the whole expected and actual structure in the test output.

libcnb-data/src/buildpack_plan.rs
hone: While that's true I got here because you can't do `serde::de::Deserialize::deserialize(self.metadata))`, you'll run into:

```
error[E0277]: the trait bound `toml::map::Map<std::string::String, Value>: bom::_::_serde::Deserializer<'_>` is not satisfied
  --> libcnb-data/src/buildpack_plan.rs:24:9                                                                                             
   |                                                             
24 |         serde::de::Deserialize::deserialize(self.metadata)                                                                  
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `bom::_::_serde::Deserializer<'_>` is not implemented for `toml::map::Map<std::string::String, Value>`
```

That's why I need to use `.try_into` and convert it into a `toml::Value`.

Malax: Ah, that is because you're trying to work with the `serde` primitives that don't know anything about `TOML`. Since `Deserializer` is not implemented for `Table`, serde can't read its primitives from our `Table`.

All toml `Value`s have `try_into()` which converts them to a `T` if `Deserialize` is implemented for `T`. Sadly, the `Table` type we use in `Entry` is not a `Value` so we need to make it one by wrapping it. As outlined earlier, we can't get around the clone since we're only borrowing the `metadata`.

Long story short, I think the code you have in here is perfectly fine.

libcnb-data/src/buildpack_plan.rs
hone: @Malax i feel like a lot of things we write should probably have these things derived on them in `libcnb-data`

libcnb-data/src/build_plan.rs
```diff
@@ -37,8 +37,14 @@ impl BuildPlanBuilder {
...
-    pub fn requires(mut self, name: impl AsRef<str>) -> Self {
-        self.current_requires.push(Require::new(name.as_ref()));
+    pub fn requires(mut self, require: impl Into<Require>) -> Self {
+        self.current_requires.push(require.into());
+        self
+    }
+
+    /// Add Require to the `BuildPlan`
+    pub fn requires_with_struct(mut self, require: Require) -> Self {
+        self.current_requires.push(require);
```
Malax, line 47: This can be removed now, right?

libcnb-data/src/build_plan.rs
```diff
@@ -105,6 +111,35 @@ impl Require {
...
+impl From<String> for Require {
+    fn from(s: String) -> Require {
+        Require::new(s)
+    }
+}
+
+impl From<&str> for Require {
+    fn from(s: &str) -> Require {
+        Require::new(s)
+    }
```
Malax, line 142: This would cover both cases + more:

```suggestion
impl<S: Into<String>> From<S> for Require {
    fn from(s: S) -> Self {
        Require {
            name: s.into(),
            metadata: Default::default(),
        }
    }
}
```

libcnb-data/src/build_plan.rs
```diff
@@ -119,4 +154,23 @@ mod tests {
...
+        let mut require = Require::new("foo");
+        let metadata = Metadata {
+            foo: String::from("bar"),
+        };
+        let result = require.metadata(metadata);
+        assert!(result.is_ok());
+        assert_eq!(
+            require.metadata.get("foo"),
+            Some(&toml::Value::String(String::from("bar")))
+        );
```
Malax, line 174: Can we streamline the assert here so it works the same as in `it_deserializes_metadata`?

Malax (changes requested):

libcnb-data/src/build_plan.rs
hone: oh right yes.

libcnb-data/src/build_plan.rs
hone: @Malax what were you thinking here? Unlike the other test, I can derive traits on the `Metadata` struct b/c that's what I'm checking against, but I have a `Table` here.

Malax (approved):
Feel free to ignore the comment on the test. 👍🏻  Thanks for the PR!

libcnb-data/src/build_plan.rs
hone: I tried adding `Eq` to `Require` but sadly it's not implemented on `toml::map::Map`. :(

https://docs.rs/toml/latest/toml/map/struct.Map.html

PR #385, opened 2022-03-27
Only create layer `env` directories if env vars are being written

### LINKED ISSUE #384: Layer `env` directories are always created even if no layer env vars are set
If a layer is defined in a buildpack that does not set any environment variables, libcnb still creates the `env`, env.build` and `env.launch` directories within `<layer dir>`.

For example:

```
$ docker run --rm -it my-image ls -al /layers/heroku_python/pip
total 36
drwxr-xr-x 9 heroku heroku 4096 Jan  1  1980 .
drwxr-xr-x 3 heroku heroku 4096 Jan  1  1980 ..
drwxr-xr-x 2 heroku heroku 4096 Jan  1  1980 bin
drwxr-xr-x 2 heroku heroku 4096 Jan  1  1980 env
drwxr-xr-x 2 heroku heroku 4096 Jan  1  1980 env.build
drwxr-xr-x 2 heroku heroku 4096 Jan  1  1980 env.launch
...
```

These unnecessary directories could be prevented by adding an early return iff `self.entries.is_empty()` is true , after the old directory clean-up, but before the new directory creation, here:
https://github.com/Malax/libcnb.rs/blob/06b33cca7fd845fcc2a07284321a9613749cca73/libcnb/src/layer_env.rs#L541-L548

edmorley:
Prevents empty `env`, `env.build` and `env.launch` directories from being created within each layer's directory, if no environment variables are set within them.

Fixes #384.
GUS-W-10905886.

### DIFF
--- a/libcnb/CHANGELOG.md
+++ b/libcnb/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Allow compilation of libcnb.rs buildpacks on Windows. Please note that this does not imply Windows container support, it's meant to allow running unit tests without cross-compiling. ([#368](https://github.com/Malax/libcnb.rs/pull/368))
 - Expose `runtime::libcnb_runtime_detect`, `runtime::libcnb_runtime_build` and their related types for advanced use-cases. Buildpack authors should not use these. ([#375](https://github.com/Malax/libcnb.rs/pull/375)).
+- Only create layer `env`, `env.build` and `env.launch` directories when environment variables are being set within them ([#385](https://github.com/Malax/libcnb.rs/pull/385)).
 
 ## [0.6.0] 2022-02-28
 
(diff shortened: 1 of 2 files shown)


edmorley:
There wasn't an obvious existing test to adapt or duplicate - I'll let you decide whether we need a new test for this or not :-)

Malax (approved):
Thanks for the fix! 👍🏻  

I was thinking about recommending to move the `fs::create_dir_all` into the `for` to avoid the extra check for emptiness, but in the end it does not really matter and might even have slightly worse runtime characteristics.

If anything, the comment threw me off at first since it (for me at least) comments the whole block, not only the if. But we can keep it there. :)

edmorley:
> and might even have slightly worse runtime characteristics

I would think it would be definitely worse - one in-memory `BTreeMap` empty check vs multiple duplicate disk stats to confirm the directory already exists :-) (no doubt mitigated by the OS's disk cache, but still)

PR #387, opened 2022-03-30
Add explicit error for missing exec.d paths

Malax:
I recently messed up packaging my buildpack and all exec.d binaries were not properly packaged. The error messages made it hard to understand what was wrong since the error only complains about "file not found" but doesn't tell you anything about which file.

This PR adds an explicit error for missing exec.d paths so this error-case is much easier to debug. This is technically a breaking change for users that manually match `libcnb::Error`. Most users won't be affected if they rely on libcnb handling internal errors.

Closes GUS-W-10919490

### DIFF
--- a/libcnb/src/layer/handling.rs
+++ b/libcnb/src/layer/handling.rs
@@ -213,6 +213,9 @@ pub enum WriteLayerError {
 
     #[error("Error while writing layer content metadata TOML: {0}")]
     TomlFileError(#[from] TomlFileError),
+
+    #[error("Cannot find exec.d file for copying: {0}")]
+    MissingExecDFile(PathBuf),
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


libcnb/src/layer/handling.rs
```diff
@@ -262,7 +265,17 @@ fn write_layer<M: Serialize, P: AsRef<Path>>(
...
+                    // IO errors when the path does not exist. We're using an explicit error variant
+                    // for a missing exec.d binary makes it easier to debug issues with packaging
+                    // since the usage of exec.d binaries often relies on implicit packaging the
+                    // buildpack author might not be aware of.
+                    Some(&path)
+                        .filter(|path| path.exists())
+                        .ok_or_else(|| WriteLayerError::MissingExecDFile(path.clone()))
+                        .and_then(|path| {
+                            fs::copy(path, exec_d_dir.join(name)).map_err(WriteLayerError::IoError)
+                        })?;
```
edmorley, line 278: Implementing using a conditional rather than a combinator seems a bit simpler, since:
- it avoids the need to wrap in `Some()`
- means the `map_err` is no longer required
- means the `.clone()` is no longer required

```suggestion
                    if !path.exists() {
                        return Err(WriteLayerError::MissingExecDFile(path));
                    }
                    fs::copy(path, exec_d_dir.join(name))?;
```

However I'm fine with it either way.

edmorley (approved):

edmorley:
Thank you for fixing this - quick UX win! :-)

PR #388, opened 2022-04-05
Replace malax/libcnb.rs with heroku/libcnb.rs

Malax:
Since the repo moved from malax/libcnb.rs to heroku/libcnb.rs

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # libcnb.rs [![Build Status]][ci] [![Docs]][docs.rs] [![Latest Version]][crates.io] [![Rustc Version 1.58+]][rustc]
 
-[Build Status]: https://img.shields.io/github/workflow/status/Malax/libcnb/CI/main
+[Build Status]: https://img.shields.io/github/workflow/status/heroku/libcnb.rs/CI/main
 [ci]: https://github.com/heroku/libcnb.rs/actions/workflows/ci.yml?query=branch%3Amain
 [Docs]: https://img.shields.io/docsrs/libcnb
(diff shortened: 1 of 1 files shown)


edmorley:
I just remembered we have URLs in the historic GitHub releases, so I've updated those too:
https://github.com/heroku/libcnb.rs/releases

Whilst doing that, I spotted that the monorepo cross-dependency change entries had broken links, since they used relative links which are no longer valid from the release page. I've changed those to absolute URLs instead.

edmorley:
The `GUS-W` autolink URLs were broken - I've adjusted the target from `...?bugorworknumber=<num>` to `...?bugorworknumber=W-<num>` since `<num>` only contains the number, and the ticket redirector needs the `W-` prefix.

edmorley:
@dzuelke @hone @lead4good @lillianzhang331 @michaelhoefer @nokome @schneems

Hi! 

Sorry for the mass at-mention, but it looks like as part of moving this repository under the `heroku` org, all of its forks have become disconnected from this one (due to this repo being private for a few mins as part of the move, due to some of the strict settings enabled for this org), causing multiple search results for `libcnb.rs` to be returned.

I don't suppose you could delete your repos below and recreate as new forks off this one, to reduce confusion as to which is the canonical repo? 

Many thanks :-)

https://github.com/dzuelke/libcnb.rs
https://github.com/hone/libcnb.rs
https://github.com/lead4good/libcnb.rs
https://github.com/lillianzhang331/libcnb.rs
https://github.com/michaelhoefer/libcnb.rs
https://github.com/nokome/libcnb.rs
https://github.com/schneems/libcnb.rs

PR #409, opened 2022-06-17
libcnb-test: Pass `--trust-builder` to `pack build`

### LINKED ISSUE #407: `libcnb-test` should pass `--trust-builder` to `pack build`
Pack has the concept of trusted builders:
https://buildpacks.io/docs/tools/pack/concepts/trusted_builders/

Any of the builders on Pack's "suggested builders" list (which will include the Heroku stacks) are automatically trusted, however if tests either use a different builder, or if Pack hasn't yet been updated to include a new builder (such as our newly created `heroku/builder:22`), then that builder won't be trusted.

If a builder is not trusted, then
1. The build stages are each run in a separate container (which is slightly slower)
2. The log output changes to have a `[<stage>]` prefix on every line.

Example logs for a trusted builder:

```
===> DETECTING
heroku/procfile 1.0.1
===> RESTORING
===> BUILDING
 
[Discovering process types]
Procfile declares types -> web, worker
===> EXPORTING
Adding layer 'launch.sbom'
...
```

Example logs for an untrusted builder:

```
===> DETECTING
[detector] heroku/procfile 1.0.1
===> RESTORING
===> BUILDING
[builder] 
[builder] [Discovering process types]
[builder] Procfile declares types -> web, worker
===> EXPORTING
[exporter] Adding layer 'launch.sbom'
...
```

In the latter there are now line prefixes like `[detector]` and `[builder]`.

If multi-line `libcnb-test` log output assertions are used (such as in https://github.com/heroku/procfile-cnb/pull/75), tests that passed on one machine (where the builder was trusted) can then fail when run on another machine where the Pack config doesn't have that builder marked as trusted.

I believe we should get `libcnb-test` to pass `--trust-builder` to its `pack build` calls (which overrides the trusted builders list just for that build), in order to:
- normalise build log output in tests across environments
- avoid having to add `pack config trusted-builders add ZZZZ` boilerplate to all of our CI configs
- speed up test execution (eg: trusting the builder makes the `procfile-cnb` integration tests take 25% less time on my local machine; albeit this is a best case, since these are simple/fast tests)

This should be safe to do, since:
- `libcnb-test` doesn't publish images or pass in credentials (one of the concerns of using an untrusted builder)
- projects using `libcnb-test` will have chosen/vetted their test images (so it's a bit different from the "end user experimenting with random builders they found on the internet" case)
- by compiling/running a Rust `libcnb.rs`-using project on your local machine you are already running a bunch of untrusted code, so you either trust the author of that buildpack or you don't.

GUS-W-11312254.

edmorley:
In order to:
- Normalise build log output in tests across environments - so multi-line log output assertions don't pass on one machine and fail on another (as happened in `procfile-cnb` for local vs CI), due to the untrusted-builder-only `[<stage>]` log output prefixes.
- Avoid having to add `pack config trusted-builders add ZZZZ` boilerplate to all of our CI configs.
- Speed up test execution.

This should be safe to do, since:
- `libcnb-test` doesn't publish images or pass in credentials (one of the concerns of using an untrusted builder)
- projects using `libcnb-test` will have chosen/vetted their test images (so it's a bit different from the "end user experimenting with random builders they found on the internet" case)
- by compiling/running a Rust `libcnb.rs`-using project on your local machine you are already running a bunch of untrusted code, so you either trust the author of that buildpack or you don't.

See the issue description for more details.

Fixes #407.
GUS-W-11312254.

### DIFF
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Pass `--trust-builder` to `pack build` to ensure the builders used in tests are always trusted ([#409](https://github.com/heroku/libcnb.rs/pull/409)).
 - Add `IntegrationTest::app_dir_preprocessor`, allowing users to modify the app directory before an integration test run ([#397](https://github.com/heroku/libcnb.rs/pull/397)).
 
 ## [0.3.1] 2022-04-12
--- a/libcnb-test/src/pack.rs
+++ b/libcnb-test/src/pack.rs
@@ -12,6 +12,7 @@ pub(crate) struct PackBuildCommand {
     image_name: String,
     path: PathBuf,
     pull_policy: PullPolicy,
+    trust_builder: bool,
     verbose: bool,
 }
 
@@ -65,6 +66,7 @@ impl PackBuildCommand {
             image_name: image_name.into(),
             path: path.into(),
             pull_policy,
+            trust_builder: true,
             verbose: false,
         }
     }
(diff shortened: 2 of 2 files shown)


Malax (approved):
👍🏻 Thanks for the detailed description in the issue!

PR #412, opened 2022-06-20
Leverage `Into` trait for `String`/`&str` arguments in `ContainerContext`

### LINKED ISSUE #405: Make `start_with_process` accept `Into<String>` not just `String`
Currently libcnb-test's `start_with_process()` only accepts `String`, so we have to do something like:

```rust
            context
                .prepare_container()
                .start_with_process(String::from("worker"), |container| {
                    // ...
                });
```

It would be more ergonomic if it accepted `&str`, eg by using `Into<String>`.

See:
https://github.com/heroku/libcnb.rs/blob/e4e75c40816000312cfe56290a049d8df9dbb72b/libcnb-test/src/container_context.rs#L122-L124

Malax:
Functions in `ContainerContext` used concrete string types in some places that could easily replaced with a generic `Into<String>`. In most cases, conversion to `String` was happening anyway. This change just exposes this via the API, allowing the user to pass in anything that can be turned into a `String`, improving DX.

Fixes #405

### DIFF
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Leverage `Into` trait for `String`/`&str` arguments in `ContainerContext` ([#412](https://github.com/heroku/libcnb.rs/pull/412)).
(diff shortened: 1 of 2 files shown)


edmorley (approved):
Thank you :-)

PR #414, opened 2022-06-20
Make `Buildpack::on_error` phase aware and fix `Buildpack::on_error` default implementation

### LINKED ISSUE #286: Default `on_error` exit code for internal errors is incorrect
The default `Buildpack::on_error` implementation (as well as the custom implementation we have in libherokubuildpack) uses an exit code of 100 if an internal buildpack error occurred:
https://github.com/Malax/libcnb.rs/blob/abe9b153535c543ec5f1a691a1f8c58b3c1413dd/libcnb/src/buildpack.rs#L46-L51

During the build phase this is fine, since all non-zero exit codes have the same meaning in the CNB spec:
https://github.com/buildpacks/spec/blob/main/buildpack.md#build

However during detect, an exit code of 100 means "fail detect" rather than "error":
https://github.com/buildpacks/spec/blob/main/buildpack.md#detection

Therefore for any buildpack where such an error occurs during detect (eg an IO error reading `buildpack.toml`), Pack will end up displaying a generic "no buildpacks passed detect" error, rather than surfacing the error message in the output.

Malax:
`Buildpack::on_error` now gets passed a `libcnb::BuildpackPhase` value. This allows error handling based on the currently executing phase. This is particularly useful since the CNB specification assigns different meanings for exit codes based on the current phase. 

The default `Buildpack::on_error` implementation now uses exit code `1` instead of `100`. The previously used exit code signalled a failed detect instead of an error during the detect phase. See #286 for details.

Fixes #286

### DIFF
--- a/libcnb/CHANGELOG.md
+++ b/libcnb/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+- `Buildpack::on_error` now gets passed a `libcnb::BuildpackPhase` value. This allows error handling based on the currently executing phase. This is particularly useful since the CNB specification assigns different meanings for exit codes based on the current phase. ([#414](https://github.com/heroku/libcnb.rs/pull/414)).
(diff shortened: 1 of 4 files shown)


Malax:
Superseded by #415

PR #416, opened 2022-06-20
Replace magic exit code values with constants

Malax:
Replace magic exit code values with constants.

In https://github.com/heroku/libcnb.rs/pull/415#pullrequestreview-1012086445, @edmorley brought up that we might not even want to return `i32` exit codes from `libcnb_runtime_detect` and `libcnb_runtime_build`.

Both functions are public, but really only used in _very advanced_ use-cases where `libcnb_runtime` isn't used. If we consider these functions practically private, then I don't think adding a layer of types in-between is of high-value. If we would copy their implementations directly to `libcnb_runtime` (their only internal call-site), we would not introduce additional types.

Considering these functions _are_ public, it might even be beneficial to keep their return values as-is, since this frees users of these functions to map between types to CNB compliant exit codes, removing CNB specific logic from the users code.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -5,7 +5,7 @@ use crate::detect::{DetectContext, InnerDetectResult};
 use crate::error::Error;
 use crate::platform::Platform;
 use crate::toml_file::{read_toml_file, write_toml_file};
-use crate::LIBCNB_SUPPORTED_BUILDPACK_API;
+use crate::{exit_code, LIBCNB_SUPPORTED_BUILDPACK_API};
 use serde::de::DeserializeOwned;
 use std::env;
 use std::ffi::OsStr;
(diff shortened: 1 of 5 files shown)


libcnb/src/runtime.rs
```diff
@@ -65,7 +65,7 @@ pub fn libcnb_runtime<B: Buildpack>(buildpack: &B) {
                     eprintln!(
                         "https://github.com/buildpacks/spec/blob/main/buildpack.md#detection"
                     );
-                    exit(1);
+                    exit(exit_code::GENERIC_UNSPECIFIED_ERROR);
```
edmorley, line 68: Should these (the "Usage" errors) get their own exit code, given the only reason they would occur is due to a libcnb issue (eg incorrect packaging of the buildpack or incorrect args handling for a new future buildpack API version), or else an end user trying to run `bin/build --help` or something?

That is, these failure modes seem similar to the other internal ones that currently have their own exit code too - so I'm unsure as to why these are inconsistent?

edmorley (approved):

libcnb/src/runtime.rs
Malax: Probably inconsistent due to being introduced at different times and maybe different authors? I agree that exit codes are a bit messy right now. At least that surfaces now with the constants. :)

PR #421, opened 2022-06-21
Make the Buildpack API version check more robust

edmorley:
Previously, if a `buildpack.toml` either:
1. Didn't match the specific custom buildpack type (used when custom fields are specified under `[[metadata]]`)
2. Didn't match the CNB spec

...then a generic "failed to deserialise buildpack.toml" message was shown, rather than any error about the buildpack API version not matching.

In cases where the chosen API version is actually the cause of the error (for example if a future API version makes breaking changes to the schema), then this means the underlying cause isn't mentioned.

Now, the check uses a minimal `BuildpackDescriptorApiOnly` type when deserialising `buildpack.toml`, to ensure the API version can still be determined in those cases.

The `BuildpackDescriptorApiOnly` type was used instead of a raw `toml::value::Table`, since otherwise there are a myriad of `Option`s and related to handle, from both the potential lack of `api` field and it not having a value of type string etc.

GUS-W-11324328.

### DIFF
--- a/libcnb-data/CHANGELOG.md
+++ b/libcnb-data/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## [Unreleased]
 
-- Add `BuildpackDescriptorApiOnly` for representing only the API version of a `buildpack.toml` ([#421](https://github.com/heroku/libcnb.rs/pull/421)).
 - Add `Launch::processes` function to add multiple `Process`es to a `Launch` value at once. ([#418](https://github.com/heroku/libcnb.rs/pull/418)) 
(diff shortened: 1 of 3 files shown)


Malax (approved):
Do we need to make `BuildpackDescriptorApiOnly` public? I'm fine with it, but it feels like an internal-only data structure.

edmorley:
@Malax Yeah I wasn't sure if it should be or not. I can move it into `runtime.rs` and make it private perhaps?

Malax:
@edmorley Yeah, that is what I had in mind. 👍🏻

PR #422, opened 2022-06-22
Allow running subsequent integration test builds

### LINKED ISSUE #278: Allow running subsequent integration test builds
This does allow for testing caching logic in integration tests.

Malax:
To achieve a consistent API, breaking changes had to be made to allow sharing of `TestConfig` (formerly `IntegrationTest`) for both initial and subsequent test runs. The biggest change is the addition of `TestRunner` which is now the way to run tests.

Even though it's a breaking change, migration to this new API should be simple. See examples in this PR for an idea of required changes.

## Changelog

- Remove `IntegrationTest::run_test`, to run a test use the new `TestRunner::run_test` function.
- Rename `IntegrationTest` to `TestConfig`.
- Rename `IntegrationTestContext` to `TestContext`.
- Add `Clone` implementation for `TestConfig`, allowing it to be shared across tests.
- Add `TestContext::run_test` and `TestContext::run_test_inherit_config`, allowing you to run subsequent integration tests with the image from a previous test. These functions allow testing of subsequent builds, including caching logic and buildpack behaviour when build environment variables change, stacks are upgraded and more.

## Examples

```rust
#[test]
fn basic() {
    TestRunner::default().run_test(
        TestConfig::new("heroku/buildpacks:20", "test-fixtures/simple-ruby-app"),
        |context| {
            assert_contains!(context.pack_stdout, "---> Ruby Buildpack");
            assert_contains!(context.pack_stdout, "---> Installing bundler");
            assert_contains!(context.pack_stdout, "---> Installing gems");

            context.run_test_inherit_config(|context| {
                assert_not_contains!(context.pack_stdout, "---> Installing bundler");
                assert_not_contains!(context.pack_stdout, "---> Installing gems");
            });
        },
    );
}

#[test]
fn clear_bundler_cache() {
    let base_config = TestConfig::new("heroku/buildpacks:20", "test-fixtures/simple-ruby-app");

    TestRunner::default().run_test(&base_config, |context| {
        assert_contains!(context.pack_stdout, "---> Ruby Buildpack");
        assert_contains!(context.pack_stdout, "---> Installing bundler");
        assert_contains!(context.pack_stdout, "---> Installing gems");

        context.run_test(
            // Subsequent tests can be run using a different config, allowing testing of stack
            // changes, different environment variables, etc. Since TestConfig is now Clone,
            // configuration can be easily shared.
            base_config.clone().env("CLEAR_BUNDLER_CACHE", "true"),
            |context| {
                assert_contains!(context.pack_stdout, "---> Installing bundler");
                assert_contains!(context.pack_stdout, "---> Installing gems");
            },
        );
    });
}
```

Closes #278, [GUS-W-10539929](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-10539929)

### DIFF
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -12,7 +12,7 @@
 - Rename `IntegrationTest` to `TestConfig`. ([#422](https://github.com/heroku/libcnb.rs/pull/422)).
 - Rename `IntegrationTestContext` to `TestContext`. ([#422](https://github.com/heroku/libcnb.rs/pull/422)).
 - Add `Clone` implementation for `TestConfig`, allowing it to be shared across tests. ([#422](https://github.com/heroku/libcnb.rs/pull/422)).
-- Add `TestContext::run_test` and `TestContext::run_test_inherit_config`, allowing you to run subsequent integration tests with the image from a previous test. These functions allow testing of subsequent builds, including caching logic and buildpack behaviour when build environment variables change, stacks are upgraded and more. ([#422](https://github.com/heroku/libcnb.rs/pull/422)).
+- Add `TestContext::run_test`, allowing you to run subsequent integration tests with the image from a previous test. These functions allow testing of subsequent builds, including caching logic and buildpack behaviour when build environment variables change, stacks are upgraded and more. ([#422](https://github.com/heroku/libcnb.rs/pull/422)).
 
 ## [0.3.1] 2022-04-12
 
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,12 +2,18 @@
 
 ## [Unreleased]
 
-- Leverage `Into` trait for `String`/`&str` arguments in `ContainerContext` ([#412](https://github.com/heroku/libcnb.rs/pull/412)).
-- Pass `--trust-builder` to `pack build` to ensure the builders used in tests are always trusted ([#409](https://github.com/heroku/libcnb.rs/pull/409)).
-- Add `IntegrationTest::app_dir_preprocessor`, allowing users to modify the app directory before an integration test run ([#397](https://github.com/heroku/libcnb.rs/pull/397)).
-- Update `bollard` dependency from 0.12.0 to 0.13.0 ([#419](https://github.com/heroku/libcnb.rs/pull/419)).
-- Update `cargo_metadata` dependency from 0.14.2 to 0.15.0 ([#423](https://github.com/heroku/libcnb.rs/pull/423)).
+- Leverage `Into` trait for `String`/`&str` arguments in `ContainerContext` ([#412](https://github.com/heroku/libcnb.rs/pull/412))
+- Pass `--trust-builder` to `pack build` to ensure the builders used in tests are always trusted ([#409](https://github.com/heroku/libcnb.rs/pull/409))
+- Add `IntegrationTest::app_dir_preprocessor`, allowing users to modify the app directory before an integration test run ([#397](https://github.com/heroku/libcnb.rs/pull/397))
+- Update `bollard` dependency from 0.12.0 to 0.13.0 ([#419](https://github.com/heroku/libcnb.rs/pull/419))
+- Update `cargo_metadata` dependency from 0.14.2 to 0.15.0 ([#423](https://github.com/heroku/libcnb.rs/pull/423))
 - Add `assert_not_contains!` macro, an inverted version of `assert_contains!`. ([#424](https://github.com/heroku/libcnb.rs/pull/424))
+- Remove `IntegrationTest::run_test`, to run a test use the new `TestRunner::run_test` function. ([#422](https://github.com/heroku/libcnb.rs/pull/422))
+- Rename `IntegrationTest` to `TestConfig`. ([#422](https://github.com/heroku/libcnb.rs/pull/422))
+- Rename `IntegrationTestContext` to `TestContext`. ([#422](https://github.com/heroku/libcnb.rs/pull/422))
+- Add `Clone` implementation for `TestConfig`, allowing it to be shared across tests. ([#422](https://github.com/heroku/libcnb.rs/pull/422))
+- Add `TestContext::run_test` and `TestContext::run_test_inherit_config`, allowing you to run subsequent integration tests with the image from a previous test. These functions allow testing of subsequent builds, including caching logic and buildpack behaviour when build environment variables change, stacks are upgraded and more. ([#422](https://github.com/heroku/libcnb.rs/pull/422))
+
 
 ## [0.3.1] 2022-04-12
 
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,18 +2,17 @@
 
 ## [Unreleased]
 
-- Leverage `Into` trait for `String`/`&str` arguments in `ContainerContext` ([#412](https://github.com/heroku/libcnb.rs/pull/412))
-- Pass `--trust-builder` to `pack build` to ensure the builders used in tests are always trusted ([#409](https://github.com/heroku/libcnb.rs/pull/409))
-- Add `IntegrationTest::app_dir_preprocessor`, allowing users to modify the app directory before an integration test run ([#397](https://github.com/heroku/libcnb.rs/pull/397))
-- Update `bollard` dependency from 0.12.0 to 0.13.0 ([#419](https://github.com/heroku/libcnb.rs/pull/419))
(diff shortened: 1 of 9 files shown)


libcnb-test/src/test_runner.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+
+        let temp_crate_buildpack_dir = if config.buildpacks.contains(&BuildpackReference::Crate) {
+            Some(
+                build::package_crate_buildpack(&config.target_triple)
+                    .expect("Could not package current crate as buildpack"),
+            )
+        } else {
+            None
+        };
```
edmorley, line 161: (Optional) As of Rust 1.50+ we have [bool::then()](https://doc.rust-lang.org/std/primitive.bool.html#method.then). I can't decide whether I prefer it for this case or not? (Though wondered if you might leap at the chance to use a combinator 😆 😄)

```suggestion
        let temp_crate_buildpack_dir =
            config
                .buildpacks
                .contains(&BuildpackReference::Crate)
                .then(|| {
                    build::package_crate_buildpack(&config.target_triple)
                        .expect("Could not package current crate as buildpack")
                });
```

libcnb-test/tests/integration_test.rs
```diff
@@ -22,7 +22,10 @@ ERROR: failed to build: failed to fetch builder image 'index.docker.io/libcnb/vo
 ")]
 fn panic_on_unsuccessful_pack_run() {
```
edmorley, line 23: Are there any more cases where it would be useful to have integration tests here? Either related to this PR (in which case potentially in scope for this PR), or for existing libcnb-test features (in which case it would be good to file a GitHub issue to track adding them later).

libcnb-test/README.md
```diff
@@ -15,35 +15,36 @@ Please use the same tag for feature requests.
...
 ```rust,no_run
 // In $CRATE_ROOT/tests/integration_test.rs
-use libcnb_test::{IntegrationTest, BuildpackReference, assert_contains};
+use libcnb_test::{assert_contains, BuildpackReference, TestRunner, TestConfig};
 
 #[test]
 fn test() {
-    IntegrationTest::new("heroku/builder:22", "test-fixtures/app")
-        .buildpacks(vec![
-            BuildpackReference::Other(String::from("heroku/openjdk")),
```
edmorley, line 24: (No action needed) I wonder if this README example of using `.buildpacks()` is the reason why some of the other projects had been using them unnecessarily (when only testing the current buildpack)? If so, I think removing the `.buildpacks()` usage here (as you have done) is a good idea, to keep things simple / avoid unnecessary usage :-)

libcnb-test/src/lib.rs
```diff
@@ -13,373 +13,131 @@ mod container_port_mapping;
...
-    /// Prepares a new container with the image from the integration test.
+    /// Starts a subsequent integration test run with inherited configuration.
     ///
-    /// This will not create nor run the container immediately. Use the returned
-    /// `PrepareContainerContext` to configure the container, then call
-    /// [`start_with_default_process`](PrepareContainerContext::start_with_default_process) on it
-    /// to actually create and start the container.
+    /// This function is the same as [`TestContext::run_test`] but automatically inherits the
+    /// configuration from the previous test run. See [`TestContext::run_test`] for details.
     ///
```
edmorley, line 115: Should `run_test_inherit_config()` have an example in its rustdoc too? 

I had to copy-paste from the PR description to construct an example in one of the other review comments, since I couldn't find one in the codebase :-)

libcnb-test/src/lib.rs
```diff
@@ -13,373 +13,131 @@ mod container_port_mapping;
...
-    /// });
-    /// ```
-    #[must_use]
-    pub fn prepare_container(&self) -> PrepareContainerContext {
-        PrepareContainerContext::new(self)
+    /// # Panics
+    /// - When the app could not be copied
+    /// - When this crate could not be packaged as a buildpack
+    /// - When the `pack` command unexpectedly fails
+    pub fn run_test_inherit_config<F: FnOnce(TestContext)>(self, f: F) {
```
edmorley, line 120: So just thinking out loud, but if we didn't have `run_test_inherit_config()`, then users would instead do something like this? (ie just pass `config` twice unchanged; maybe with a clone, I didn't run the below)

```rust
#[test]
fn clear_bundler_cache() {
    let config = TestConfig::new("heroku/buildpacks:20", "test-fixtures/simple-ruby-app");

    TestRunner::default().run_test(&config, |context| {
        assert_contains!(context.pack_stdout, "---> Ruby Buildpack");
        assert_contains!(context.pack_stdout, "---> Installing bundler");
        assert_contains!(context.pack_stdout, "---> Installing gems");

        context.run_test(&config, |context| {
            assert_contains!(context.pack_stdout, "---> Installing bundler");
            assert_contains!(context.pack_stdout, "---> Installing gems");
        });
    });
}
```

If so, then I wonder if we even need `run_test_inherit_config()` at all? 

Whilst it does save a small amount of boilerplate, (a) re-using the exact config only applies to a very small subset of testing scenarios (much of the time you need to change something before performing a new build, eg adjusting runtime version or whatever), (b) having a single API rather than two seems like it might be easier for people to learn/remember?

If we decided to not add it now, we could also add it later, if we find a lot of tests would benefit from it?

Anyway, I'm open to ideas here :-)

edmorley (approved):

libcnb-test/README.md
```diff
@@ -15,35 +15,36 @@ Please use the same tag for feature requests.
...
-        ])
-        .run_test(|context| {
+    TestRunner::default().run_test(
+        TestConfig::new("heroku/builder:22", "test-fixtures/app"),
+        |context| {
             assert_contains!(context.pack_stdout, "---> Maven Buildpack");
             assert_contains!(context.pack_stdout, "---> Installing Maven");
             assert_contains!(context.pack_stdout, "---> Running mvn package");
 
-            context.prepare_container().expose_port(12345).start(|container| {
```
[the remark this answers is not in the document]
Malax: Yeah, it should compile now. 🙌🏻

libcnb-test/src/lib.rs
Malax: As discussed in https://github.com/heroku/libcnb.rs/pull/422#discussion_r905110888, the function will be removed altogether.

libcnb-test/CHANGELOG.md
```diff
@@ -2,12 +2,18 @@
 
 ## [Unreleased]
 
-- Leverage `Into` trait for `String`/`&str` arguments in `ContainerContext` ([#412](https://github.com/heroku/libcnb.rs/pull/412)).
```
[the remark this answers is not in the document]
Malax: Yes, for consistency. But if this doesn't fix anything I'll revert this and we can have a dedicated PR for changelog hygenie. That makes more sense anyway. 👍🏻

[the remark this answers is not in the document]
Malax: Addressed in [35da518](https://github.com/heroku/libcnb.rs/pull/422/commits/35da518425e261b1b906c147b3031fe0fed462a3)

libcnb-test/src/test_runner.rs
Malax: 🙌🏻  [45bda1f](https://github.com/heroku/libcnb.rs/pull/422/commits/45bda1f4d59d4a2fc5d83ea43ed5e0f55f806cf7)

libcnb-test/src/test_runner.rs
```diff
@@ -0,0 +1,216 @@
...
+        TestRunner::new(tokio_runtime, docker)
+    }
+}
+
+impl TestRunner {
+    /// Creates a new runner that uses the given Tokio runtime and Docker connection.
+    ///
+    /// This function is meant for advanced use-cases where fine control over the Tokio runtime
+    /// and/or Docker connection is required. For the common use-cases, use `Runner::default`.
+    pub fn new(tokio_runtime: tokio::runtime::Runtime, docker: Docker) -> Self {
```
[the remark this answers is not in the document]
Malax: I like the idea of a default TestRunner and the ability to create your own if you ever need to. Since this isn't a huge maintenance burden (we simply expose something we need in any case) I'll keep it.

libcnb-test/tests/integration_test.rs
Malax: This is an interesting problem. So far, I added the "tests" for `libcnb-test` to the examples since we would need custom buildpacks to test specific features of `libcnb-test` and I have no idea where to put them. For example, the Ruby example now runs a subsequent test.

Problem with this approach that it isn't really clear we rely on the examples to test critical functionality.

PR #431, opened 2022-06-24
Remove `TestContext::app_dir`, only copy `app_dir` when preprocessor is configured

Malax:
GitHub code search did not yield any results that showed that people are using this field.

## Changelog

- Remove `TestContext::app_dir` to encourage the use of `TestConfig::app_dir_preprocessor`.
- Improve performance when no `TestConfig::app_dir_preprocessor` is configured by skipping application directory copying.

### DIFF
--- a/libcnb-test/src/test_runner.rs
+++ b/libcnb-test/src/test_runner.rs
@@ -147,12 +147,12 @@ impl TestRunner {
             // Copy the app to a temporary directory if an app_dir_preprocessor is specified and run the
             // preprocessor. Skip app copying if no changes to the app will be made.
             if let Some(app_dir_preprocessor) = &config.app_dir_preprocessor {
-                let app_dir = app::copy_app(&normalized_app_dir_path)
+                let temporary_app_dir = app::copy_app(&normalized_app_dir_path)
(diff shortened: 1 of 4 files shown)


libcnb-test/src/test_runner.rs
```diff
@@ -134,21 +134,29 @@ impl TestRunner {
...
+            if let Some(app_dir_preprocessor) = &config.app_dir_preprocessor {
+                let app_dir = app::copy_app(&normalized_app_dir_path)
+                    .expect("Could not copy app to temporary location");
 
-        if let Some(app_dir_preprocessor) = &config.app_dir_preprocessor {
-            (app_dir_preprocessor)(PathBuf::from(temp_app_dir.path()));
-        }
+                (app_dir_preprocessor)(app_dir.as_path().to_owned());
+
+                app_dir
```
edmorley, line 155: (Optional) Would renaming `app_dir` (just inside this block) to `temporary_app_dir` make it clearer what's different about what's being returned in this codepath? 

It's just at the moment, if it were not for the `.expect()` message, it wouldn't be clear to me that this is a temporary directory.

edmorley (approved):
Thank you for this!

PR #433, opened 2022-06-24
Compile README tests as a part of doctest

schneems:
In https://github.com/heroku/libcnb.rs/pull/366#issuecomment-1151577355 I left feedback that stated the README was broken. I found we actually DO have doctests on it but even if you add a non-existent method in there it still passes. Why? Because the method has the `#[test]` line on it so it only compiles on the test environment and doctest does not use the test environment.

I removed the config declaration and added a comment explaining why.

In `lib.rs` I moved the ordering to match https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html but I don't think it has an impact.

### DIFF
--- a/libcnb-test/README.md
+++ b/libcnb-test/README.md
@@ -17,7 +17,9 @@ Please use the same tag for feature requests.
 // In $CRATE_ROOT/tests/integration_test.rs
 use libcnb_test::{assert_contains, BuildpackReference, TestRunner, TestConfig};
 
-#[test]
+// In your code you'll want to mark your function as a test with `#[test]`.
+// It is removed here for compatability with doctest so this code in the readme
+// tests for compilation.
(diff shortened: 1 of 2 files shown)


schneems:
Conflicts with https://github.com/heroku/libcnb.rs/commit/83356f9777e8351e7d7f77cf2e5bbb0dafa6faab. Which is updated, but still not tested due to the issue with `#[test]`. I'll update this with that new code in a bit. In the mean time weigh in on the approach of moving the code out of a function. Maybe there are alternatives.

Malax (changes requested):
Nice find! Could we add at least the `fn` back in and maybe comment out the `#[test]` attribute with a comment telling people that they need in a real-world scenario?

Malax (approved):

Issue #365, opened 2022-03-02
Consider disabling some of `fancy-regex`'s default features


edmorley:
The [fancy-regex](https://crates.io/crates/fancy-regex) crate is used in a few parts of libcnb.rs:
https://github.com/Malax/libcnb.rs/search?q=fancy

By default, the crate (like the underlying `regex` crate), has many additional features enabled by default:
https://github.com/fancy-regex/fancy-regex/blob/cccff7c089091917e1c9a890296799e5ccf83f04/Cargo.toml#L13-L22

These map to the features offered by the [regex](https://crates.io/crates/regex) crate:
https://docs.rs/regex/latest/regex/#crate-features

For the simple use-cases in libcnb we likely don't need the unicode related features. 

We also may not need the perf related features, given that (a) our regexes are simple expressions, (b) the haystacks are small (eg validating fields in the CNB spec), (c) validation isn't on the critical path of buildpack speed.

For example, dropping both the unicode and perf features improves:
- Cold `cargo libcnb packge --release` build times by 9% for `example/basics`
- Buildpack binary size by 38% for release+stripped `example/basics`

PR #435, opened 2022-06-29
Stop using regex to validate `BuildpackVersion`

edmorley:
The previous implementation used regex due to the spec requirement of redundant leading zeros not being permitted. 

However, implementing this without regex is just as easy, avoids using a heavy dependency (though sadly we still need it elsewhere in the codebase, at least for now), is closer to the approach used for `BuildpackApi`, and IMO slightly cleaner.

It also means one fewer regex usage that needs to be tweaked for #365.

### DIFF
--- a/libcnb-data/src/buildpack/version.rs
+++ b/libcnb-data/src/buildpack/version.rs
@@ -1,4 +1,3 @@
-use fancy_regex::Regex;
 use serde::Deserialize;
 use std::convert::TryFrom;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


Malax (approved):
I'm not sure if I would value the not-quite-removal of `fancy_regex` higher than the added complexity of the implementation. The existing implementation is IMO easier to understand. 
I'm not blocking this PR since I can live with it and the perceived complexity is isolated in a function.

PR #440, opened 2022-06-30
Fix rustdocs for `LogOutput`

### LINKED ISSUE #404: Fix docs for `libcnb-test`'s `LogOutput`
When looking at https://docs.rs/libcnb-test/0.3.1/libcnb_test/struct.ContainerContext.html#method.logs_now the `LogOutput` return type isn't a link.

Also if I manually search the docs for that type, I get zero results?
https://docs.rs/libcnb-test/0.3.1/libcnb_test/index.html?search=LogOutput

Maybe it's because there isn't currently a rustdoc for the type?
https://github.com/heroku/libcnb.rs/blob/e4e75c40816000312cfe56290a049d8df9dbb72b/libcnb-test/src/log.rs#L3-L9

edmorley:
The rustdocs for `LogOutput` were previously not being shown since the module was not publicly exported, even though it was exposed via other modules.

Fixes #404.

### DIFF
--- a/libcnb-test/src/lib.rs
+++ b/libcnb-test/src/lib.rs
@@ -19,6 +19,7 @@ mod test_runner;
 mod util;
 
 pub use crate::container_context::*;
+pub use crate::log::*;
 pub use crate::test_config::*;
 pub use crate::test_context::*;
 pub use crate::test_runner::*;
--- a/libcnb-test/CHANGELOG.md
+++ b/libcnb-test/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Fix rustdocs for `LogOutput`. ([#440](https://github.com/heroku/libcnb.rs/pull/440)).
+
 ## [0.4.0] 2022-06-24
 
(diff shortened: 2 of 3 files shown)


libcnb-test/src/lib.rs
```diff
@@ -19,6 +19,7 @@ mod test_runner;
 mod util;
 
 pub use crate::container_context::*;
+pub use crate::log::*;
```
Malax, line 22: Strictly speaking, do we need this? From the description, `crate::log::LogOutput` was exposed already? Is the idea there that we might as well expose it here?

Malax (approved):

libcnb-test/src/lib.rs
edmorley: It fixed the generated rustdocs - I don't know if there is a better way to fix it, or if if fact it was already supposed to work the way it was and we were instead running into a rustdocs bug?

Malax: Ah! Now I see what you mean - we expose values of this type, but since the type itself is not public, it will not be in the generated rustdocs. Gotcha. Yeah - then we have to do this. Thanks for clarifying! 👍🏻

PR #442, opened 2022-07-01
Fix the buildpack size reported by `cargo libcnb package`

### LINKED ISSUE #336: Buildpack size output by `cargo libcnb package` is incorrect
The `cargo libcnb package` command outputs the size of the built+packaged buildpack,  eg:

```
basics $ cargo libcnb package
...
INFO - Writing buildpack directory...
INFO - Successfully wrote buildpack directory: ../../target/buildpack/debug/libcnb-examples_basics (50.5M)
```

However this size is incorrect:

```
basics $ du -hs ../../target/buildpack/debug/libcnb-examples_basics/
25M	../../target/buildpack/debug/libcnb-examples_basics/
```

The size calculation is performed here:
https://github.com/Malax/libcnb.rs/blob/0b9b8214b56e41ad1a59635effb4198d9886143d/libcnb-cargo/src/main.rs#L188-L190

It looks like [fs_extra::dir::get_size()](https://docs.rs/fs_extra/latest/fs_extra/dir/fn.get_size.html) is buggy, and iterates all entries in a directory without taking into account whether they are symlinks or not:
https://github.com/webdesus/fs_extra/blob/9b972027bac127236d00da3f807891d1972ff49c/src/dir.rs#L754-L765

I'll file an upstream issue for this, however the `fs-extra` crate looks to be unmaintained (judging by open issues/PRs, no merges to `master` since Feb 2021, and no new release since Aug 2020). 

One option might just be to remove the printing of the buildpack size. This would also let us drop the dependency on both `fs-extra` and `size_format` (and its 8 deps) from `libcnb-cargo`.

edmorley:
Previously the size of the packaged buildpack reported by `cargo libcnb package` was incorrect, since `fs_extra::dir::get_size` doesn't handle symlinks correctly (webdesus/fs_extra#59). This meant the reported sizes counted the buildpack binary size twice, due to the symlink of `bin/detect` to `bin/build`.

In addition, the size was previously displayed in SI units (and with only an `M` suffix, not `MB`), when we really want it to use binary sizes (ie: MiB; divided by `1024*1024` rather than `1000*1000`).

Now, both of these issues are fixed, and the dependency on both the `fs_extra` (which is unmaintained/not super reliable) + `size_format` (which is bloated for what we need) crates have been dropped, reducing the total number of dependencies of `libcnb-cargo` from 62 to 54.

Lastly, if an IO error occurred during calculation of the buildpack size, this is no longer ignored (by listing the size as "unknown size"), but instead results in a non-zero exit code. Whilst the size is a nice to have, calculation of the size should never fail in the common case, so covering up failures seems more likely to hide things than be an overall benefit.

Fixes #336.
GUS-W-11375184.


libcnb-cargo/src/main.rs
```diff
@@ -173,15 +173,20 @@ fn handle_libcnb_package(args: PackageArgs) {
...
 
+    let size_in_bytes = calculate_dir_size(&output_path).unwrap_or_else(|io_error| {
+        error!("IO error while calculating buildpack directory size: {io_error}");
+        std::process::exit(exit_code::UNSPECIFIED_ERROR);
+    });
+
+    // Precision will only be lost for sizes bigger than 52 bits (~4 Petabytes), and even
+    // then will only result in a less precise figure, so is not an issue.
+    #[allow(clippy::cast_precision_loss)]
+    let size_in_mb = size_in_bytes as f64 / (1024.0 * 1024.0);
```
Malax, line 184: (Optional) Ultra pedantic mode: ENGAGED! 😂 

> In addition, the conversion of the size from bytes to MB was incorrect, since it used SI units (ie: divided by 1000\*1000 rather than 1024\*1024).

Technically, it was correct before since `MB` refers to the SI value where a megabyte is exactly a million bytes. The value here should be called a Mebibyte and using `MiB` as the abbreviation.

libcnb-cargo/src/main.rs
```diff
@@ -195,3 +200,26 @@ fn setup_logging() {
         std::process::exit(exit_code::UNSPECIFIED_ERROR);
     }
 }
+
+/// Recursively calculate the size of a directory and its contents in bytes.
+// Not using `fs_extra::dir::get_size` since it doesn't handle symlinks correctly:
+// https://github.com/webdesus/fs_extra/issues/59
+fn calculate_dir_size(path: impl AsRef<Path>) -> io::Result<u64> {
```
Malax, line 207: (Optional, Observation) Since this is not tail-recursive, we might overflow the stack when the directory structure is deep. Probably nothing to worry about for a standard buildpack. Looking at `fs_extra`, their implementation isn't better in this regard, so I'm fine with keeping it as-is.

Malax (approved):

libcnb-cargo/src/main.rs
edmorley: Ah true. Though before it just said `M` not `MB` so from an end user perspective it wasn't displaying what we wanted it to say either way. I'll change it to `MiB` though :-)

PR #443, opened 2022-07-04
libcnb-cargo: Disable unused `stderrlog` features

edmorley:
As of `stderrlog` v0.5.3 (https://github.com/cardoe/stderrlog-rs/pull/45), it's possible to disable the timestamps feature of `stderrlog` (which we don't use), which otherwise requires the `chrono` dependency. 

Doing so reduces the number of transitive dependencies by 5, and also removes Chrono from the dependency tree, which currently has reported CVEs so can cause security audit false positives.

GUS-W-11378865.


edmorley:
Between this and #442, `libcnb-cargo` now has 13 fewer dependencies (62 -> 49). This should help slightly with installation times eg in CI, where the "install libcnb-cargo" step can take up to 2 minutes.

Malax (approved):

PR #455, opened 2022-07-06
Improve the error message shown when Pack CLI is not installed

### LINKED ISSUE #296: Validate pack install before attempting to run it
Currently, `libcnb-test` tries to run `pack` and does not:

- Validate its installed
- Validate the installed version satisfies the version requirement

In such cases, errors will pop up but they're not very descriptive. By running a validation step first, we can output better errors.

edmorley:
Before:

```
---- basic_build stdout ----
thread 'basic_build' panicked at 'Could not spawn external 'pack' process: Os { code: 2, kind: NotFound, message: "No such file or directory" }', /Users/emorley/src/libcnb.rs/libcnb-test/src/test_runner.rs:188:14
```

After:

```
---- basic_build stdout ----
thread 'basic_build' panicked at 'External `pack` command not found. Install Pack CLI and ensure it is on PATH: https://buildpacks.io/docs/install-pack', /Users/emorley/src/libcnb.rs/libcnb-test/src/test_runner.rs:184:21
```

No integration test has been added, since writing one would involve manipulating the `PATH` of the integration test process which would be prone to intermittent test failures given concurrent test execution occurs in threads.

In addition all panic messages that reference "pack" now do so inside backticks rather than single quotes, since the panic error message itself is inside single quotes, so it looked like the quoted section had ended prematurely.

Fixes #296.
GUS-W-11390631.

### DIFF
--- a/libcnb-test/src/test_runner.rs
+++ b/libcnb-test/src/test_runner.rs
@@ -2,10 +2,10 @@ use crate::pack::PackBuildCommand;
 use crate::{app, build, util, BuildpackReference, PackResult, TestConfig, TestContext};
 use bollard::Docker;
 use std::borrow::Borrow;
-use std::env;
 use std::env::VarError;
 use std::path::PathBuf;
 use std::process::{Command, Stdio};
+use std::{env, io};
 
 /// Runner for libcnb integration tests.
 ///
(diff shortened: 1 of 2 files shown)


libcnb-test/src/test_runner.rs
```diff
@@ -185,9 +185,15 @@ impl TestRunner {
...
+            .unwrap_or_else(|io_error| {
+                if io_error.kind() == io::ErrorKind::NotFound {
+                    panic!("External `pack` command not found. Install Pack CLI and ensure it is on PATH: https://buildpacks.io/docs/install-pack");
+                } else {
+                    panic!("Could not spawn external `pack` process: {io_error}");
+                };
+            })
             .wait_with_output()
-            .expect("Error while waiting on external 'pack' process");
+            .expect("Error while waiting on external `pack` process");
```
Malax, line 196: (joke) We're outputting Markdown now? :D

Malax (approved):

PR #470, opened 2022-07-19
libcnb-test: Rename `Test{Runner,Context}::run_test` and `TestConfig`

edmorley:
Performs the following renames:
- `TestRunner::run_test` -> `TestRunner::build`
- `TestConfig` -> `BuildConfig`
- `TestContext::run_test` -> `TestContext::rebuild`

Whilst there were arguments for using `run_test` at the top level (such as the closure being the whole test, not just the build):
- using `run_test` again for the rebuild case didn't fit as well
- it obscures the fact that inside the closure is initially only the result of the build, and not anything else (until containers are started etc).

Similarly, the rename of `TestConfig` to `BuildConfig` should make it clearer that options there (such as `BuildConfig.env`) only apply  to the build, and don't also apply to the started containers.

GUS-W-11415596.

### DIFF
--- a/examples/execd/tests/integration_test.rs
+++ b/examples/execd/tests/integration_test.rs
@@ -7,13 +7,13 @@
 // https://rust-lang.github.io/rust-clippy/stable/index.html
 #![warn(clippy::pedantic)]
 
-use libcnb_test::{assert_contains, assert_empty, TestConfig, TestRunner};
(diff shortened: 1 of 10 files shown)


Malax (approved):
I don't have strong opinions here, I'm fine with how it is right now and I am fine how it will be after this is merged. :)

PR #472, opened 2022-07-19
libcnb-test: Expand integration test coverage

edmorley:
Adds coverage for:
- `ContainerContext::address_for_port`
- `ContainerConfig::envs` (only `env` was tested previously)
- `ContainerConfig::app_dir_preprocessor` when rebuilding

Also replaces the hardcoded sleep with a retrying HTTP client. Longer term we may want to add support for a retrying test HTTP client to `libcnb-test` behind an opt-in Cargo feature flag.

GUS-W-11451593.

### DIFF
--- a/libcnb-test/Cargo.toml
+++ b/libcnb-test/Cargo.toml
@@ -24,6 +24,7 @@ tokio-stream = "0.1.9"
 
 [dev-dependencies]
 indoc = "1.0.6"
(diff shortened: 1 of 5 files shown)


edmorley:
Tests are failing due to this needing the fix from #471. I'll rebase on that once it merges.

dzuelke (approved):

Issue #486, opened 2022-08-17
Add newtype for `launch::Slice` path globs


Malax:
As per [spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#launchtoml-toml), these are Go standard library globs and must follow that syntax. This is potentially more complex to implement as a newtype than the ones we currently have.

edmorley:
I'd be inclined to wontfix this to be honest. Usage of slice layers is likely going to be pretty rare, and people using it can just read the docs. Whilst it's nice that libcnb.rs can prevent a subset of errors at runtime though typing and nice DX, I don't think it's viable for it to prevent every class of issue.

Issue #166, opened 2021-11-17
Ensure API consistency for builders and builder-like types


Malax:
Some types don't provide builder-like functionality but have dedicated `*Builder` types, others have it with the normal type, some have nothing. We should ensure API consistency.

Malax:
At this point, I'm leaning towards explicit `Builder` suffixes and non-consuming builder implementations. Explicit names aid with discoverability and non-consuming builders are the "gold standard" in the Rust ecosystem. They allow our users to do chained calls _and_ mutable values with the same API.

PR #487, opened 2022-08-17
Add LaunchBuilder

### LINKED ISSUE #485: Add support for slice layers
To implement caching for a few heroku buildpack features with libcnb.rs (listed in https://github.com/heroku/buildpacks-ruby/tree/schneems/fresh-start-rust#application-contract-compile) I would like an implementation of slice layers https://github.com/buildpacks/spec/blob/main/buildpack.md#slice-layers.

A slice layer allows a buildpack to specify a layer that is within the application directory. So this will allow me to cache `<app-dir>/public/assets` without having to move files on disk around.

@Malax previously said this is a thing he thinks is easy to implement. Maybe if we're running short on time, if he can share thoughts on desired interface someone can try to stub out an implementation.

GUS-W-11486542.

Malax:
This PR originally started as a fix for #485. As it turns out, adding slices to `Launch` was already possible, albeit more complicated as it should be since there were no builder functions available for slices.

This PR adds a dedicated `LaunchBuilder` that allows setting all fields for `Launch`, removes the builder-style functions from `Launch` as they were inconsistent with the rest of libcnb.rs anyways. I still want to tackle #166, but this isn't the PR for that.

I also added some docs for `Slice` that clarify what `paths` actually is (and renamed it). I opted to not implement a newtype for Go standard library path globs for now as we can spend our time better in other areas. I created an issue for it as #486.

## Changelog

- Replace builder style functions from `Launch` with a dedicated `LaunchBuilder` to be more consistent with other builders in the library. Additionally, all fields of `Launch` can now be modified via the builder pattern. 
- Rename `paths` field in `launch::Slice` to `path_globs` and add docs to make it clearer that these strings are Go standard library globs.

Closes #485, [GUS-W-11486542](https://gus.my.salesforce.com/apex/ADM_WorkLocator?bugorworknumber=W-11486542)

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ Modify the project's `src/main.rs` file to contain the following:
 
 ```rust,no_run
 use libcnb::build::{BuildContext, BuildResult, BuildResultBuilder};
-use libcnb::data::launch::{Launch, ProcessBuilder};
+use libcnb::data::launch::{LaunchBuilder, ProcessBuilder};
 use libcnb::data::process_type;
 use libcnb::detect::{DetectContext, DetectResult, DetectResultBuilder};
 use libcnb::generic::{GenericError, GenericMetadata, GenericPlatform};
@@ -153,12 +153,13 @@ impl Buildpack for HelloWorldBuildpack {
 
         BuildResultBuilder::new()
             .launch(
-                Launch::new().process(
+                LaunchBuilder::new().process(
                     ProcessBuilder::new(process_type!("web"), "echo")
                         .arg("Hello World!")
                         .default(true)
                         .build(),
-                ),
+                )
+                .build(),
             )
             .build()
     }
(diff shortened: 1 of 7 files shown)


Malax:
Currently looking into a fix for the integration test failures that are triggered by us having multiple `heroku/procfile` buildpacks in the builder used for the integration tests. These failures are very likely not related to this PR.

edmorley (approved):
Thank you!

schneems:
A little late on the draw, but I reviewed this. Looks good. 

I see that the `slice` and `slices` functions take something that implements Into `Slice`. I don't see where we're implementing that anywhere other than the default that any struct can be made into itself (though perhaps I missed it). 

Do we have examples of building a slice or should we just manually create the struct for now https://github.com/heroku/libcnb.rs/blob/d0a35f7adbfaaf814b2c6643aeb6f463aae8d777/libcnb-data/src/launch.rs#L225-L232 ?

PR #489, opened 2022-08-23
Update for CNB spec `0.8` (Including SBOM support)

### LINKED ISSUE #269: API 0.7: Adjust buildpack version parts to be `u64`
One of the changes in buildpack API 0.7 is:

>  Clarifies buildpack API must be u64 numbers ([#265](https://github.com/buildpacks/spec/pull/265)).

We'll want to change each of the major/minor/patch version parts to be `u64` here:
https://github.com/Malax/libcnb.rs/blob/7a8092a6d879c48ee0ee623002b77ad9b69bcf66/libcnb-data/src/buildpack/version.rs#L14-L18

### LINKED ISSUE #270: API 0.7: Support for SBOM
In Buildpack API 0.7 there are several SBOM related changes.

> Build phase expects Software Bill of Materials (SBOM) to be output as files with the `application/vnd.cyclonedx+json`, `application/spdx+json`, or `application/vnd.syft+json` media type ([#269](https://github.com/buildpacks/spec/pull/269), [RFC 0095](https://github.com/buildpacks/rfcs/blob/main/text/0095-sbom.md)). The `[bom]` tables in launch.toml and build.toml are no longer supported.

There was also a followup change not mentioned in the release entry:
https://github.com/buildpacks/spec/pull/281

GUS-W-10426512.

### LINKED ISSUE #271: API 0.7: Adjust supported API version
We'll need to adjust the API version here from 0.6 to 0.7:
https://github.com/Malax/libcnb.rs/blob/7a8092a6d879c48ee0ee623002b77ad9b69bcf66/libcnb/src/lib.rs#L40-L41

We'll also likely want to bump the API version used in various test fixtures, for completeness.

GUS-W-10426513.

### LINKED ISSUE #294: API 0.8: Support for Process Specific Working Directories
For any given process, the buildpack should be able to specify a working directory that is not CNB_APP_DIR.

https://github.com/buildpacks/spec/issues/212

GUS-W-10502884.

### LINKED ISSUE #298: Identify changes required to support Buildpack API 0.8
Buildpack API 0.8 will be out later this year:
https://github.com/buildpacks/spec/milestone/19

This issue is a placeholder for the work to identify what libcnb changes will be required to support it. Each change will need to be filed as a new issue under the [Buildpack API 0.8](https://github.com/Malax/libcnb.rs/milestone/7) GitHub milestone.

So far we have identified:
- #294
- #299

GUS-W-10502876.

### LINKED ISSUE #299: API 0.8: Adjust supported API version
We'll need to adjust the API version here from 0.7 to 0.8:
https://github.com/Malax/libcnb.rs/blob/7a8092a6d879c48ee0ee623002b77ad9b69bcf66/libcnb/src/lib.rs#L40-L41

We'll also likely want to bump the API version used in various test fixtures, for completeness.

GUS-W-10502896.

Malax:
Updates libcnb.rs to target CNB spec `0.8`. Changes between `0.6` and `0.8` are minimal with the exception of SBOM support which takes up the majority of this PR. I opted to **not** open a `0.8` branch and target multiple PRs to it to speed up the process. Look at individual commits for easier review.

## Relevant Documentation

- [Upstream CNB Buildpack API 0.8 Release Notes](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.8)
- [Upstream CNB Buildpack API 0.7 Release Notes](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.7)
- [Upstream CNB Buildpack Specification: Software-Bill-Of-Materials](https://github.com/buildpacks/spec/blob/b4eeb4390676419ce4fe794eab50a169e834ffc2/buildpack.md#software-bill-of-materials)

## Note on Cloud Native Buildpacks [RFC 100](https://github.com/buildpacks/rfcs/blob/main/text/0100-buildpack-input-vars.md)

In version `0.8`, a change has landed that added environment variables for values that were previously command line arguments to `/bin/detect` and `/bin/build`. This is an optional change since the classic command line arguments are still passed in and will never be removed ([ref](https://github.com/buildpacks/rfcs/blob/main/text/0100-buildpack-input-vars.md#how-it-works)). I decided to not switch over to the environment variables for this PR to keep it smaller. We can easily decide later to make the switch. Right now, I see no benefits for libcnb.rs itself or its users.

## Highlight: SBOM Support

Since the CNB specification supports multiple SBOM formats and there are many ways to generate and represent them, libcnb.rs treats them as raw bytes with attached metadata about the format.

This means there is no direct support for generating an SBOM file with libcnb.rs. Buildpack authors are expected to leverage external tooling such as build tool plugins or Rust crates to generate SBOM files. 

We expect most buildpacks to use language specific tooling directly instead of implementing SBOM generation themselves. Examples are [cyclonedx-maven-plugin](https://github.com/CycloneDX/cyclonedx-maven-plugin), [sbt-bill-of-materials](https://github.com/lightbend/sbt-bill-of-materials), [cyclone-dx-gradle-plugin](https://github.com/CycloneDX/cyclonedx-gradle-plugin), [cyclonedx-ruby](https://rubygems.org/gems/cyclonedx-ruby/versions/1.1.0), [cyclonedx-rust-cargo](https://github.com/CycloneDX/cyclonedx-rust-cargo). Alternatively, `syft` has support for many ecosystems built-in: https://github.com/anchore/syft#supported-ecosystems.

libcnb.rs offers `From` implementations for Rust SBOM libraries (currently only [cyclonedx-bom](https://github.com/CycloneDX/cyclonedx-rust-cargo/tree/main/cyclonedx-bom)) so that these library types can be used directory with libcnb.rs without an explicit conversion step. Enable the appropriate features to gain access to them.

As the Rust SBOM ecosystem expands, we will add support for additional libraries. Support is quite sparse right now: [crates.io: Crates with 'SBOM' keyword](https://crates.io/keywords/sbom).

## Highlight: SBOM Testing Support

`libcnb-test` was updated to allow downloading of SBOM files from a built OCI image. This feature uses `pack sbom download` under the hood and provides an easy way to lookup SBOM files in the rather complex directory structure `pack sbom download` creates.

```rust
TestRunner::default().build(
    BuildConfig::new("heroku/builder:22", "test-fixtures/app"),
    |context| {
        context.download_sbom_files(|sbom_files| {
            assert!(sbom_files
                .path_for(
                    buildpack_id!("heroku/jvm"),
                    SbomType::Launch,
                    SbomFormat::SyftJson
                )
                .exists());
        });
    },
);
```

Fixes #270, Fixes #271, Fixes #299, Fixes #298, Fixes #294, Fixes #269, Fixes GUS-W-10502884


libcnb-test/src/test_context.rs
```diff
@@ -199,6 +244,20 @@ impl<'a> TestContext<'a> {
...
+/// Downloaded SBOM files.
+pub struct SbomFiles {
+    sbom_files_directory: PathBuf,
+}
+
+/// The type of SBOM.
+///
+/// Not to be confused with [`libcnb_data::sbom::SbomFormat`].
+pub enum SbomType {
+    Launch,
```
edmorley, line 256: Should the `Launch` variant also have rustdocs, for completeness?

libcnb-test/src/test_context.rs
```diff
@@ -199,6 +244,20 @@ impl<'a> TestContext<'a> {
     }
 }
 
+/// Downloaded SBOM files.
```
edmorley, line 247: Should these new types be added below the `impl<'a> Drop for TestContext<'a>` block, so the different types and their implementations aren't intermixed?

libcnb/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Fix permission issues during layer handling when the layer contains read-only directories. ([#488](https://github.com/heroku/libcnb.rs/pull/488)).
 - Add explicit `DeleteLayerError` to provide more context when debugging layer handling problems. ([#488](https://github.com/heroku/libcnb.rs/pull/488)).
+- libcnb.rs now targets [Buildpack API 0.8](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.8). Buildpacks need to upgrade the `api` key to `0.8` in their `buildpack.toml`. ([#489](https://github.com/heroku/libcnb.rs/pull/489))
```
edmorley, line 7: Do we need to mention the new methods on `BuildResultBuilder` and `LayerResultBuilder` (given they are `libcnb` specific, not `libcnb-data` specific), or perhaps a brief mention of SBOM support?

libcnb/src/build.rs
```diff
@@ -190,4 +197,22 @@ impl BuildResultBuilder {
         self.store = Some(store);
         self
     }
+
+    /// Adds a build SBOM to the build result.
+    ///
+    /// Entries in this SBOM represent materials in the build container(sic!) for auditing purposes.
```
edmorley, line 203: I think we should remove the `(sic!)` and use a corrected description of what the SBOM represents, rather than use the wording from upstream verbatim.

I also found this alternative description upstream, which doesn't have that mistake (no need to use this verbatim either; just mentioning for inspiration):

> Write a partial Software Bill of Materials to <layers>/build.sbom.<ext> describing any provided build dependencies not associated with a layer.

https://github.com/buildpacks/spec/blob/main/buildpack.md#purpose-2

libcnb/src/layer/handling.rs
```diff
@@ -747,6 +788,7 @@ mod tests {
                 metadata: GenericMetadata::default(),
             },
             ExecDPrograms::Overwrite(HashMap::new()),
+            Sboms::Keep,
```
edmorley, line 791: There are no ~unit tests added here. I presume that's because we're favouring the testing via the test buildpack?

If so, in the future were we going to migrate some of the other IO-heavy "unit" tests in this file to test-buildpacks too?

edmorley (changes requested):
This is great, thank you!

Requesting changes pending the discussion about the `WorkingDirectory` enum vs using `Option<PathBuf>`.

libcnb-test/src/test_context.rs
Malax: Happy to take suggestions what to put there. I wanted to add some Rustdocs but couldn't come up with something that wasn't "well, launch like it says, duh." :)

libcnb-test/src/test_context.rs
Malax: Yeah, makes sense. Fixed in d6e74a6,

libcnb/src/build.rs
Malax: Oh, this is not a mistake. At least I didn't understand it as such. From my understanding, the `build` SBOM is strictly for things that got added to the build _container_ (hence the sic! as it might be unexpected). The SBOMs added for build aren't in any way connected to the resulting image.

The description you linked isn't wrong, but is omitting the fact that that BOM only describes the build container and that it won't be associated with the launch image at all.

edmorley: I think the use of "sic" here makes the meaning of this more confusing, not less. Could you reword without it?

edmorley: I think in general we might need to give some examples of the types of things that belong in each category of SBOM, given buildpack authors might not know otherwise. Though I know this is partly subjective, and maybe something we'll just have to figure out over time :-)

libcnb-test/src/test_context.rs
edmorley: As someone who hasn't read the SBOM RFC recently, I would struggle to guess what belongs in a launch SBOM and what it's really for. As such, some rustdocs here seems necessary - and not obvious to me?

Malax: I added changed the rustdocs for both via [1a9ac2c](https://github.com/heroku/libcnb.rs/pull/489/commits/1a9ac2cec66aeac47398a4b9e89f8c47f009111b). In the end I still believe the RustDoc here is just stating the obvious. The CNB spec itself does not describe the SBOM types beyond this in any meaningful way. 

Since I don't have any recommendations how to use these SBOM building blocks in a production buildpack, I'm hesitant adding any more documentation here that goes beyond the spec.

libcnb/CHANGELOG.md
Malax: Yeah, I added an entry for those in [1a9ac2c](https://github.com/heroku/libcnb.rs/pull/489/commits/1a9ac2cec66aeac47398a4b9e89f8c47f009111b). 👍🏻

libcnb/src/build.rs
Malax: Fixed (I think?) in [1f99649](https://github.com/heroku/libcnb.rs/pull/489/commits/1f99649540732d2f07d0275f782d3b7e01949157). Agreed that it might be confusing - you did comment on it for a reason. ;-)

libcnb/src/layer/handling.rs
Malax: Yeah, these changes here are just to make them work again. Unsure about the permutations we should check as well. I agree that we should have some strategy when to test what with which tooling. "If you need to create temporary directories and/or files it should be an integration test via test buildpacks" seems sensible to me. Thoughts?

edmorley: Yeah that works for me.

libcnb-data/src/launch.rs
```diff
@@ -123,6 +115,54 @@ pub struct Process {
     pub direct: bool,
     #[serde(default, skip_serializing_if = "std::ops::Not::not")]
     pub default: bool,
+    #[serde(
+        rename = "working-directory",
+        default,
+        skip_serializing_if = "WorkingDirectory::is_app"
+    )]
+    pub working_directory: WorkingDirectory,
```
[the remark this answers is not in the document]
Malax: Sure! It all boils down to `working_directory: None` being less explicit what will happen than `working_directory: WorkingDirectory::App`. I think that being explicit here adds DX value for users of libcnb.rs since they don't need to go and look up what `None` means in the CNB spec. That information could be added as a RustDoc on the field and all "getter"/"setter" style functions, but I think that types are the most explicit form here and don't require docs in multiple places.

I actually did refactor it once while working on this PR. I started with `WorkingDirectory` as seen here, went to `Option<PathBuf>` thinking "man, this additional code, is it really worth it?" but went back when I saw the result.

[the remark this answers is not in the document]
edmorley: Yeah I get why, I just don't think it's consistent with how we've modelled everything else. Most of the other types are very thin layers over the spec. eg Launch commands run from the app dir, and we don't have a type to represent that.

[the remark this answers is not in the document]
edmorley: I think it's mainly since we're trying to model the absence of something, it just feels strange to have a type for it.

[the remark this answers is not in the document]
Malax: Do you have an example that shows where you think we're inconsistent when we add `WorkingDirectory`? I did a quick search in `libcnb-data` and this is the only use of `PathBuf` so far, so there is at least no precendent how we handle paths. I do think that adding that layer around `PathBuf` here is consistent with the rest of `libcnb-data`, our newtypes being the clearest case, but `Provide` is another case where we simply wrap another type for clarity.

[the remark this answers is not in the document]
edmorley: So thinking about this more, `WorkingDirectory: None` is kinda weird, and yes as you say the `Option`s we have are for things where the `None` case is more intuitive. 

As such happy to leave this as an enum.

libcnb-data/src/launch.rs
```diff
@@ -155,6 +195,7 @@ impl ProcessBuilder {
                 args: Vec::new(),
                 direct: false,
                 default: false,
+                working_directory: WorkingDirectory::App,
```
edmorley, line 198: The rustdocs for `ProcessBuilder` needs updating to mention this default value (like it does for `direct`, `default` etc).

edmorley (approved):

Issue #358, opened 2022-02-27
Prevent compatibility issues with mismatched `libcnb-cargo`, `libcnb-test` and `libcnb` versions 


edmorley:
As part of getting set up with a libcnb.rs powered buildpack, users will run `cargo install libcnb-cargo`, which installs the latest `libcnb-cargo` available at the time.

Over time we will end up adding new features to the libcnb related crates, that require corresponding support in `libcnb-cargo`.

However, whilst users will have Dependabot to tell them to update their libcnb deps in `Cargo.toml`, they won't have anything to remind them to update the globally installed `libcnb-cargo`.

This may mean they see unexpected errors, or missing features, when trying to use their older version of `libcnb-cargo` with the newer libcnb deps. For example that seen in #357.

There are a few ways we can deal with this:
1. Have `libcnb-cargo` check for updates from time to time and output an "update available" message.
2. Have `libcnb-cargo` check that the libcnb crate versions in the `Cargo.toml` of the project being packaged match an allowlist of supported libcnb versions.

I feel like (2) might be the best approach, since:
- it handles both the forwards and backwards compatibility cases (whereas encouraging users to update libcnb-cargo via "update available" messages might actually mean they update to a version that's too new for that project)
- it would be the simplest to implement, given `libcnb-cargo` already loads cargo metadata
- it would also benefit `libcnb-test`, since it also needs to be compatible with the `libcnb` version being used (as they are peer deps not parent/child, there's a chance they might get out of sync).

edmorley:
Hmm though I can now see a bunch of issues with (2), since how can `libcnb-cargo` know in advance which future releases of `libcnb` will/won't be compatible, without making every major release incompatible by default? (Which wouldn't be great for UX)

PR #492, opened 2022-08-30
Synchronous Crate Versions

### LINKED ISSUE #335: Document project release process
We should document the release process used for this repository, to ensure we don't forget steps. (In the future we may want to automate parts of it, but for now some docs would be a great start.)

The process is roughly:
1. Open a PR to bump the versions of the crates that are to be published (and any other crates that depend on those crates), making sure to check for any breaking changes that might require a major version bump.
2. Obtain review of the PR and then merge it.
3. Locally, ensure you are on `main` (not the branch for the PR) and have pulled. Then and Git tag HEAD of main with the relevant tags for each package (using lightweight tags, not annotated) - following the new tag naming scheme from #327 (see https://github.com/Malax/libcnb.rs/tags for examples).
4. Push those tags to the remote.
5. `cargo publish` each crate, starting at the leaf crates (those at the bottom of the dependency tree).
6. Create GitHub releases for each tag (the quickest way to do this is to visit https://github.com/Malax/libcnb.rs/tags and use "create release"), using the markdown from the respective changelog entries. The GitHub release title should be the same as the GitHub tag name, but with the `/` replaced with a space.

Malax:
Currently, all libcnb.rs crates use dedicated versions. This has some drawbacks:

- Hard to remember which version of which crate is the most current one.
- No indication that i.e. `libcnb-test 0.6.0` is intended to be used with `libcnb 0.9.0`, similar stories with the Cargo plugin and the `libcnb-data` crate.
- Implementing compatibility verification into the Cargo plugin and `libcnb-test` is harder since it would require a mapping file. See #358.
- Cutting a release requires copy&pasting or at least cross-linking of CHANGELOGs to include cross-crate changes. For example, if `libcnb-data` changes, almost all crates will be affected by that change implicitly and need CHANGELOG mentioning.
- When releasing, different version numbers for every crate makes it easy to make mistakes.

The idea with this PR is to release all crates at the same time and with the same version. Additionally, a unified CHANGELOG at the repository root is used, making it easier to discover all changes for users. This will avoid all of the issues mentioned before. However, this solution is not perfect either:

- We might release a new version of a crate without any changes. This is especially likely with `libcnb-proc-macros` which changes rarely and does not depend on another libcnb package. For the other crates, this is less of a problem as they all depend depend on each other and all of them on `libcnb-data`, which is the crate that changes most often.
- A unified CHANGELOG bears the question what happens with the old CHANGELOG entries. For this PR, I kept them where they were, renamed them to `HISTORICAL_CHANGELOG.md`, added a comment on top, explaining the change to a unified changelog and mentioning the historical CHANGELOGs in the new one.

In the end, I think the upsides outweigh the downsides.

Closes GUS-W-11631951, Fixes #335

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,37 @@
+# Changelog
+
+This is the new, unified, changelog that contains changes from across all libcnb.rs crates. Before version `0.10.0`,
+separate changelogs for each crate were used. If you need to refer to these old changelogs, find them named
+`HISTORICAL_CHANGELOG.md` in their respective crate directories.
+
+## [Unreleased]
+
+Highlight of this release is the bump to
+[Buildpack API 0.8](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.8) which brings support for SBOM to
+libcnb.rs. This is also the first release where all libcnb.rs crates are released at the same time and with the same
+version number. See the changelog below for other changes.
+
+### Changes
+
+- libcnb.rs now targets [Buildpack API 0.8](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.8). Buildpacks need to upgrade the `api` key to `0.8` in their `buildpack.toml`. ([#489](https://github.com/heroku/libcnb.rs/pull/489))
+- In accordance to the CNB specification `>=0.7`, `BuildpackId` no longer permits `sbom` as a buildpack id. ([#489](https://github.com/heroku/libcnb.rs/pull/489))
+- Replace builder style functions from `Launch` with a dedicated `LaunchBuilder` to be more consistent with other builders in the library. Additionally, all fields of `Launch` can now be modified via the builder pattern. ([#487](https://github.com/heroku/libcnb.rs/pull/487))
+- Rename `paths` field in `launch::Slice` to `path_globs` and add docs to make it clearer that these strings are Go standard library globs. ([#487](https://github.com/heroku/libcnb.rs/pull/487))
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,37 @@
...
+`HISTORICAL_CHANGELOG.md` in their respective crate directories.
+
+## [Unreleased]
+
+Highlight of this release is the bump to
+[Buildpack API 0.8](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.8) which brings support for SBOM to
+libcnb.rs. This is also the first release where all libcnb.rs crates are released at the same time and with the same
+version number. See the changelog below for other changes.
+
+### Changes
```
edmorley, line 14:
```suggestion
### Changed
```

(To make tense consistent with the other sections, plus the suggested categories from https://keepachangelog.com/en/1.0.0/ )

edmorley (changes requested):
Thank you for doing this! I agree the benefits outweigh the downsides.

Could you also add some docs on releasing (perhaps a "Releasing" section of a `CONTRIBUTING.md` or `MAINTAINERS.md` or similar, based on the checklist in #335) so it's clearer how the new workflow will work? 

eg: One question I had was "are we now only going to create one tag/GitHub release entry?" and these docs would answer that/allow us to discuss/review those changes along with the code changes.

Malax:
I added a `RELEASING.md` in [48ff87c](https://github.com/heroku/libcnb.rs/pull/492/commits/48ff87c059674081f8b39ed7520e065f2d9447c0). I didn't go with `MAINTAINERS.md` since prior-art looks like it's most often used for a list of maintainers, not maintainer instructions.

edmorley (approved):
🎉

PR #495, opened 2022-09-21
Migrate libherokubuildpack into libcnb.rs

### LINKED ISSUE #10: add core helper libs

Malax:
This PR migrates the contents of `libherokubuildpack` into the libcnb.rs repository. The utilities contained in that crate are useful beyond Heroku buildpacks and moving them in this repository makes it easier to maintain (since it depends on libcnb itself). Eventually, we want to ship an `util` module or crate.

Beyond just moving the code here, I made some changes:

- Added cargo features for logical blocks of functionality. Users can pick and choose which helpers they need to avoid unnecessary binary size bloat. Especially the download utilities come with a bunch of dependencies that might not be needed in some cases. All features are enabled by default for now.
- `libherokubuildpack` already divided the provided functionality in logical modules but re-exported them at the crate level. I removed the re-export to make the crate features easier to understand as they now match the module names.
- Removed all mentions of Heroku in documentation and function names. In all cases it wasn't really Heroku specific, but just an opinionated implementation. (This has been done in preparation for eventual migration to libcnb.rs proper)
- Added a rudimentary `README.md` file.

Closes #10, GUS-W-11796008


edmorley:
I used GitHub code search to check which repositories are using `libherokubuildpack`, and it's just these:
https://github.com/heroku/buildpacks-go
https://github.com/heroku/buildpacks-jvm
https://github.com/heroku/buildpacks-nodejs
https://github.com/heroku/procfile-cnb

Plus these not-yet-released branches:
https://github.com/heroku/buildpacks-ruby/tree/schneems/fresh-start-rust
https://github.com/heroku/buildpacks-python/tree/wip (I'll update this one)

libherokubuildpack/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "libherokubuildpack"
+version = "0.10.0"
+edition = "2021"
+rust-version = "1.60"
+license = "BSD-3-Clause"
+description = "Opinionated common code for buildpacks implemented with libcnb.rs"
+keywords = ["buildpacks", "CNB"]
+repository = "https://github.com/heroku/libcnb.rs"
```
edmorley, line 9:
```suggestion
repository = "https://github.com/heroku/libcnb.rs/tree/main/libherokubuildpack"
```

libherokubuildpack/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+digest = ["dep:sha2"]
+error = ["log", "dep:libcnb"]
+log = ["dep:termcolor"]
+tar = ["dep:tar", "dep:flate2"]
+toml = ["dep:toml"]
+fs = ["dep:pathdiff"]
+
+[dependencies]
+libcnb = { path = "../libcnb", version = "0.10.0", optional = true }
+flate2 = { version = "1.0.24", optional = true }
```
edmorley, line 26:
```suggestion
flate2 = { version = "1.0.24", optional = true }
libcnb = { path = "../libcnb", version = "0.10.0", optional = true }
```

libherokubuildpack/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+license = "BSD-3-Clause"
+description = "Opinionated common code for buildpacks implemented with libcnb.rs"
+keywords = ["buildpacks", "CNB"]
+repository = "https://github.com/heroku/libcnb.rs"
+documentation = "https://docs.rs/libherokubuildpack"
+readme = "./README.md"
+include = ["src/**/*", "../LICENSE", "../README.md"]
+
+[features]
+default = ["digest", "error", "log", "tar", "toml"]
```
edmorley, line 15:
```suggestion
```

I would be inclined to have no features enabled by default. Virtually zero consumers will need all features, so otherwise everyone is going to have to do:

```toml
libherokubuildpack = { version = "0.10.0", default-features = false, features = ["foo", "bar"] }
```

Whereas if no features are enabled by default, this is simplified to:


```toml
libherokubuildpack = { version = "0.10.0", features = ["foo", "bar"] }
```

In addition to less boilerplate, it prevents people from forgetting to disable the features they are not using.

libherokubuildpack/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+This crate is optional and not required to write buildpacks with libcnb.rs. It provides helpers that buildpack authors 
+commonly need. Examples are digest generation, filesystem utilities, HTTP download helpers and tarball extraction. 
+
+## Crate Features
+
+It is common to not need all the helpers in this crate. To avoid including unnecessary code and dependencies, this crate
+uses Cargo features to allow opt-out of certain modules if they're not needed.
+
+The feature names line up with the modules in this crate. All features are enabled by default.
```
edmorley, line 15: This says all features are enabled by default. However even now, only some features are enabled by default. (And we may also want to disable all by default, pending the other comment.)

libherokubuildpack/Cargo.toml
```diff
@@ -0,0 +1,36 @@
...
+edition = "2021"
+rust-version = "1.60"
+license = "BSD-3-Clause"
+description = "Opinionated common code for buildpacks implemented with libcnb.rs"
+keywords = ["buildpacks", "CNB"]
+repository = "https://github.com/heroku/libcnb.rs"
+documentation = "https://docs.rs/libherokubuildpack"
+readme = "./README.md"
+include = ["src/**/*", "../LICENSE", "../README.md"]
+
```
edmorley, line 13: This will ensure the docs.rs build enables all features, so they are documented (we should probably do this for the `libcnb` crate too, so the anyhow error `From` implementation shows up):

```suggestion

[package.metadata.docs.rs]
all-features = true

```

edmorley (approved):

libherokubuildpack/Cargo.toml
Malax: That would mean in practice that a mindless `libherokubuildpack = "x.x.x"` would give them nothing since all modules are behind feature flags. Isn't that a bit confusing? Whereas reducing binary size/compilation time is a task where you usually go to feature flags anyway.

libherokubuildpack/README.md
Malax: Not all of them being enabled is an oversight on my part. :) Let's continue the all enabled as default discussion on the other comment. :)

libherokubuildpack/Cargo.toml
Malax: This is a good one, I wasn't even aware that this is a thing we need to do. Thanks! 👍🏻

libherokubuildpack/README.md
Malax: Enabled all of them as part of [bdebdbf](https://github.com/heroku/libcnb.rs/pull/495/commits/bdebdbfa79bf3f7fe479d498b1d78b119a9bd7fc)

libherokubuildpack/Cargo.toml
Malax: Sorted dependencies as part of [bdebdbf](https://github.com/heroku/libcnb.rs/pull/495/commits/bdebdbfa79bf3f7fe479d498b1d78b119a9bd7fc)

libherokubuildpack/Cargo.toml
Malax: Fixed in [bdebdbf](https://github.com/heroku/libcnb.rs/pull/495/commits/bdebdbfa79bf3f7fe479d498b1d78b119a9bd7fc)

libherokubuildpack/Cargo.toml
edmorley: I don't think it's too confusing, since they'd just get a compile error then enable the feature. ie: It's not possible to forget to enable the feature and introduce a bug for example.

The main thing here is that the default shouldn't be slow compile times. If you make something opt in, there will always be that person who doesn't read the docs or bother to look at the details, who will then never realise they are paying the additional compile time price.

edmorley: As you say, I think an "all" feature would result in the same mistakes. I also don't think that's how people develop in practice. You start with one part of a buildpack, and then add thinks as you need them. I don't go to crates.io, think "what would be useful?" and add half a dozen dependencies in advance for example. All of the features will still be visible in the docs (and eventually our libcnb examples, I hope) so people know what's there :-)

Malax: It a bit different than that though. It's a "helpful bit and pieces for buildpacks" library, not a collection of full libraries (as in your example). I still find it weird that you will include this library and get... nothing by default. This matter is not hill I wanna die on, but it doesn't feel right to me. 🤔

edmorley: If there are a subset of features we think 95% of buildpacks will always be using, then I'm fine with enabling just those by default. (Candidates to me seem to be: `log`, `error` maybe?)

However I'm against enabling all features by default, since that would effectively mean using `libherokubuildpack = "x.x.x"` is always a performance bug, at which point why do we provide such a footgun?

colincasey: @Malax @edmorley are we optimizing for convenience or performance?  

personally i'm leaning towards convenience.  i would not be pleased to add the `libherokubuildpack` utilities only to discover it gives you zero utilities out of the box 😅

i think it makes more sense from a consumer perspective to give them all the tools and they have the option to opt-out of ones that aren't needed later on when it's time to tune for performance.

edmorley: I don't think that's a fair comparison though. It's not exactly hard to add a feature to `features`, whereas it's very easy to not remember to turn unused features off. This is speaking as someone who has spent quite a bit of time on libcnb and buildpacks trying to improve compile times by looking through all of our crates and working out what on-by-default features can be switched off.

edmorley: Our current CNBs usage of libherokubuildpack functionality is as follows:
- https://github.com/heroku/buildpacks-go: `log`
- https://github.com/heroku/buildpacks-jvm: `log`, `error`, `download`, `tar`, `digest`, `toml`
- https://github.com/heroku/buildpacks-nodejs: `log`, `error`, `download`, `tar`, `fs`, `toml`
- https://github.com/heroku/procfile-cnb: `log`, `error`
- https://github.com/heroku/buildpacks-python/tree/wip: `log`, `error`
- https://github.com/heroku/buildpacks-ruby/tree/schneems/fresh-start-rust: Doesn't use `libherokubuildpack`

For Python I don't use `download` / `tar` / `digest` since I stream the download instead. If the libcnb utils version streamed then I would switch to that.

This would suggest to me that we:
- enable only `log` and `error` by default for now
- leave the other features disabled by default
- revisit as and when parts of libherokubuildpack get moved into libcnb proper (and as the utils improve we'll get a sense of how many buildpacks will use them, which will help with the decision)

colincasey: i hear what you're saying @edmorley but performance tuning is usually a step that happens at the tail end of the development process (if at all).  you can't know what's unnecessary until the code is done so it seems reasonable to give library authors a complete toolbox and have them figure out what they don't need afterwards.  

this is much like how you describe your process of improving compile times.  you work with a functioning buildpack and figure out what you can subtract from it.

edmorley: I still don't think that's a fair comparison. 

If we were talking about rewriting an algorithm to make it more verbose/complex in order to improve performance, then yes, that's something you absolutely don't want to do until (a) the end, (b) profiling has told you that you need to make that change (in most cases the simpler implementation will be fine).

In this case, we're talking about adding unused functionality/features, which is quite different.

To take this example to the extreme, should Cargo out of the box add the top 100 dependencies to my `Cargo.toml` so I can "work out what I might use", and then me have to remove them when I don't need them?

In terms of discoverability of the libherokubuildpack features, they will be listed in the docs already, so I don't feel that activating the features by default in Cargo really improves that?

edmorley: Let's just add the things we know people definitely use for now (`log`, `error`, per above) and then we can always expand the list over time if it turns out most people use more of these utils (which realistically won't happen until we improve the utils in the first place). Adding additional "enabled by default" features is not a breaking change, so something we can easily do later :-)

edmorley: Just to check - is the main objection that needing to add lesser-used features to `Cargo.toml` is too much friction, or something else?

Malax: Looking at the output of `cargo tree -l libherokubuildpack`, `ureq` seems to be the biggest offender in terms of added bloat that is potentially unused:

```
libherokubuildpack v0.10.0 (/Users/manuel.fuchs/projects/libcnb.rs/libherokubuildpack)
├── flate2 v1.0.24
│   ├── crc32fast v1.3.2
│   │   └── cfg-if v1.0.0
│   └── miniz_oxide v0.5.3
│       └── adler v1.0.2
├── libcnb v0.10.0 (/Users/manuel.fuchs/projects/libcnb.rs/libcnb)
│   ├── libcnb-data v0.10.0 (/Users/manuel.fuchs/projects/libcnb.rs/libcnb-data)
│   │   ├── fancy-regex v0.10.0
│   │   │   ├── bit-set v0.5.2
│   │   │   │   └── bit-vec v0.6.3
│   │   │   └── regex v1.6.0
│   │   │       └── regex-syntax v0.6.27
│   │   ├── libcnb-proc-macros v0.10.0 (proc-macro) (/Users/manuel.fuchs/projects/libcnb.rs/libcnb-proc-macros)
│   │   │   ├── cargo_metadata v0.15.0
│   │   │   │   ├── camino v1.0.9
│   │   │   │   │   └── serde v1.0.143
│   │   │   │   │       └── serde_derive v1.0.143 (proc-macro)
│   │   │   │   │           ├── proc-macro2 v1.0.40
│   │   │   │   │           │   └── unicode-ident v1.0.1
│   │   │   │   │           ├── quote v1.0.20
│   │   │   │   │           │   └── proc-macro2 v1.0.40 (*)
│   │   │   │   │           └── syn v1.0.98
│   │   │   │   │               ├── proc-macro2 v1.0.40 (*)
│   │   │   │   │               ├── quote v1.0.20 (*)
│   │   │   │   │               └── unicode-ident v1.0.1
│   │   │   │   ├── cargo-platform v0.1.2
│   │   │   │   │   └── serde v1.0.143 (*)
│   │   │   │   ├── semver v1.0.10
│   │   │   │   │   └── serde v1.0.143 (*)
│   │   │   │   ├── serde v1.0.143 (*)
│   │   │   │   └── serde_json v1.0.83
│   │   │   │       ├── itoa v1.0.2
│   │   │   │       ├── ryu v1.0.10
│   │   │   │       └── serde v1.0.143 (*)
│   │   │   ├── fancy-regex v0.10.0 (*)
│   │   │   ├── quote v1.0.20 (*)
│   │   │   └── syn v1.0.98 (*)
│   │   ├── serde v1.0.143 (*)
│   │   ├── thiserror v1.0.31
│   │   │   └── thiserror-impl v1.0.31 (proc-macro)
│   │   │       ├── proc-macro2 v1.0.40 (*)
│   │   │       ├── quote v1.0.20 (*)
│   │   │       └── syn v1.0.98 (*)
│   │   └── toml v0.5.9
│   │       └── serde v1.0.143 (*)
│   ├── libcnb-proc-macros v0.10.0 (proc-macro) (/Users/manuel.fuchs/projects/libcnb.rs/libcnb-proc-macros) (*)
│   ├── serde v1.0.143 (*)
│   ├── stacker v0.1.15
│   │   ├── cfg-if v1.0.0
│   │   ├── libc v0.2.126
│   │   └── psm v0.1.20
│   │       [build-dependencies]
│   │       └── cc v1.0.73
│   │   [build-dependencies]
│   │   └── cc v1.0.73
│   ├── thiserror v1.0.31 (*)
│   └── toml v0.5.9 (*)
├── pathdiff v0.2.1
├── sha2 v0.10.2
│   ├── cfg-if v1.0.0
│   ├── cpufeatures v0.2.2
│   │   └── libc v0.2.126
│   └── digest v0.10.3
│       ├── block-buffer v0.10.2
│       │   └── generic-array v0.14.5
│       │       └── typenum v1.15.0
│       │       [build-dependencies]
│       │       └── version_check v0.9.4
│       └── crypto-common v0.1.3
│           ├── generic-array v0.14.5 (*)
│           └── typenum v1.15.0
├── tar v0.4.38
│   ├── filetime v0.2.16
│   │   ├── cfg-if v1.0.0
│   │   └── libc v0.2.126
│   ├── libc v0.2.126
│   └── xattr v0.2.3
│       └── libc v0.2.126
├── termcolor v1.1.3
├── thiserror v1.0.31 (*)
├── toml v0.5.9 (*)
└── ureq v2.5.0
    ├── base64 v0.13.0
    ├── chunked_transfer v1.4.0
    ├── flate2 v1.0.24 (*)
    ├── log v0.4.17
    │   └── cfg-if v1.0.0
    ├── once_cell v1.13.0
    ├── rustls v0.20.6
    │   ├── log v0.4.17 (*)
    │   ├── ring v0.16.20
    │   │   └── untrusted v0.7.1
    │   │   [build-dependencies]
    │   │   └── cc v1.0.73
    │   ├── sct v0.7.0
    │   │   ├── ring v0.16.20 (*)
    │   │   └── untrusted v0.7.1
    │   └── webpki v0.22.0
    │       ├── ring v0.16.20 (*)
    │       └── untrusted v0.7.1
    ├── url v2.2.2
    │   ├── form_urlencoded v1.0.1
    │   │   ├── matches v0.1.9
    │   │   └── percent-encoding v2.1.0
    │   ├── idna v0.2.3
    │   │   ├── matches v0.1.9
    │   │   ├── unicode-bidi v0.3.8
    │   │   └── unicode-normalization v0.1.19
    │   │       └── tinyvec v1.6.0
    │   │           └── tinyvec_macros v0.1.0
    │   ├── matches v0.1.9
    │   └── percent-encoding v2.1.0
    ├── webpki v0.22.0 (*)
    └── webpki-roots v0.22.3
        └── webpki v0.22.0 (*)
[dev-dependencies]
└── tempfile v3.3.0
    ├── cfg-if v1.0.0
    ├── fastrand v1.7.0
    ├── libc v0.2.126
    └── remove_dir_all v0.5.3
```

Some other `libherokubuildpack` dependencies depend on `libc`, but since `libcnb` itself transitively depends on `libc` its not that bad IMO.

What do we think of just omitting the `download` feature from the default set of features for now then? I think the argument for not including it by default because it carries a bunch of dependencies with it makes sense, whereas disabling the other features are more of a mini-optimisation as they are much more lightweight.

edmorley: It seems if we're not enabling `download` by default, then it seems we shouldn't enable the related `digest` or `tar` by default either?

Also, no one has currently provided a counter-argument for why we shouldn't use current usage of features (https://github.com/heroku/libcnb.rs/pull/495#discussion_r977877448) to guide what we enable by default for now? 

This seems to be the most data-driven argument, so I don't think we should ignore it.

edmorley: That said, I think a lot of the decisions here are going to need to be revisited in the future, as features are moved into libcnb proper, since:
- as the utils improve, usage of them will change
- when we switch to streaming downloads, we'll potentially want to merge the `download` and `tar` features into one (since the majority of cases will need both), and possibly also `digest` too (particularly if we want to encourage people to "do the right think" in most cases)
- being in libcnb means the utils will always be included (whereas now people can just not use `libherokubuildpack` at all), so potentially this means we need to be more strict about what we enable by default
etc

So I'd be willing to accept all features enabled by default in `libherokubuildpack` for now, with the understanding that this is a temporary state and will not be the case necessarily in the future as things move into libcnb.

Malax: > So I'd be willing to accept all features enabled by default in libherokubuildpack for now, with the understanding that this is a temporary state

I think this is true for most of the stuff in this repository. :) Thanks for unblocking this for now!

PR #500, opened 2022-09-23
Use Cargo 1.64 features for workspace metadata

Malax:
Cargo `1.64` allows to specify some package metadata at workspace level (https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html#cargo-improvements-workspace-inheritance-and-multi-target-builds). This is something we wanted for some time now, especially when #492 was implemented. 

The new Cargo version also allows to specify dependencies at workspace level that packages can inherit. This is quite useful for repository-local dependencies that are now only specified once and not across all packages in the repository. This makes cutting a release even easier.

Downsides are that Rust 1.64 is a very new release at the time of writing and people using libcnb might need to bump their Rust version in both CI and locally to use it. However, it will only affect users that bump their libcnb version.

Closes GUS-W-11800636

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ edition = "2021"
 license = "BSD-3-Clause"
 
 [workspace.dependencies]
+libcnb = { version = "0.10.0", path = "libcnb" }
 libcnb-data = { version = "0.10.0", path = "libcnb-data" }
 libcnb-package = { version = "0.10.0", path = "libcnb-package" }
 libcnb-proc-macros = { version = "0.10.0", path = "libcnb-proc-macros" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,5 +24,3 @@ license = "BSD-3-Clause"
 libcnb-data = { version = "0.10.0", path = "libcnb-data" }
 libcnb-package = { version = "0.10.0", path = "libcnb-package" }
 libcnb-proc-macros = { version = "0.10.0", path = "libcnb-proc-macros" }
-thiserror = "1.0.31"
-toml = "0.5.9"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,3 +25,4 @@ libcnb = { version = "0.10.0", path = "libcnb" }
 libcnb-data = { version = "0.10.0", path = "libcnb-data" }
 libcnb-package = { version = "0.10.0", path = "libcnb-package" }
 libcnb-proc-macros = { version = "0.10.0", path = "libcnb-proc-macros" }
+libcnb-test = { version = "0.10.0", path = "libcnb-test" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,3 +13,16 @@ members = [
     "test-buildpacks/readonly-layer-files",
     "test-buildpacks/sbom"
 ]
+
+[workspace.package]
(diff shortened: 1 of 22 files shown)


test-buildpacks/readonly-layer-files/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
 name = "readonly-layer-files"
 version = "0.0.0"
-edition = "2021"
-rust-version = "1.60"
+edition.workspace = true
+rust-version.workspace = true
 publish = false
 
 [dependencies]
```
edmorley, line 8: This and the other test-buildpack depend on libcnb via explicit dependencies, but could be switched to using inherited dependencies too.

(The same applies to the example buildpacks, but for the examples I think we want them to be more easily copy-pasteable by end users; whereas these test-buildpacks presumably aren't intended for use as examples?)

Cargo.toml
```diff
@@ -12,3 +12,16 @@ members = [
...
     "test-buildpacks/sbom"
 ]
+
+[workspace.package]
+version = "0.10.0"
+rust-version = "1.64"
+edition = "2021"
+license = "BSD-3-Clause"
+
+[workspace.dependencies]
```
edmorley, line 22: Using `cat **/Cargo.toml | rg ' = "\d' | sort | uniq -d` I found the following additional dependencies that are shared across more than one crate:

```
cargo_metadata
fancy-regex
fastrand
serde
tempfile
```

A couple use features - which means they'd need both the workspace entry, and a more complex entry in the crate `Cargo.toml` than just `true`.

It feels like we should either include all duplicated third-party dependencies here, or else no third-party dependencies here.

In the past I've typically updated all dependencies using `cargo upgrade`, which means the versions don't get out of sync and there's no effort from having to update multiple locations (which potentially reduces one of the benefit of having inherited dependencies for these).

It also seems like over time we might adjust the dependencies of a crate and forget to re-adjust the workspace, and so end up with new duplicates, or alternatively dependencies listed in the root `Cargo.toml` that are now only used by one crate. This seems like it might end up being additional overhead / potential for inconsistency.

As such, I'm on the fence as to whether to include the third-party dependencies here - and am maybe very slightly leaning towards not including them? Though could be persuaded either way. (The libcnb* ones are a definite win however.)

Thoughts? :-)

edmorley (approved):

Cargo.toml
Malax: Yeah, honestly, I was very so-so on the third-party dependencies as well. Hard to draw a clear line when they should be in the workspace or not. The main benefit of this is that we don't have different versions of a dependency in packages. However, with dependabot, this is really a non-issue in practice.

Let remove the thrid-party dependencies for now. 👍🏻

test-buildpacks/readonly-layer-files/Cargo.toml
Malax: I mean, the examples aren't very copy & paste friendly anyway with the repository-local dependency. Since users wanting to c&p an example, with both the current solution and the workspace one, need to replace everything coming after `libcnb` in the `Cargo.toml`, I lean towards using the workspace dependency there as well.

edmorley: Yeah agree :-)

PR #517, opened 2022-10-17
libcnb: Remove use of the `stacker` crate

### LINKED ISSUE #516: Reconsider the use of `stacker` in `remove_recursively()`
In #488 the `stacker` crate was added as a dependency, since there was a concern about stack overflows, given the added `remove_recursively()` function is recursive.

The explanation for adding this can be seen in the code comment:
https://github.com/heroku/libcnb.rs/blob/109ba190c086176c9d393cfa201f71f2c2606655/libcnb/src/util.rs#L44-L51

However:
- There was no testing performed at the time to see if overflowing the stack was something that could occur (or even mention as to max depths given filesystem limitation etc).
- The stdlib `remove_dir_all` implementation (which this function replaces) is also recursive, and it doesn't protect against stack overflows.
- I was away at the time, so didn't get to provide the above feedback before the PR was merged.

Whilst the dependency is only small (so the impact on compile times likely not significant), we should still avoid unnecessary dependencies where possible, since in aggregate it can make a difference. In addition, it seems worth avoiding the cognitive burden of this uncommon dependency, if we don't actually need to use it.

edmorley:
Since:
- The stdlib `remove_dir_all` recursive implementation (which this function is replacing) doesn't have stack overflow protection.
- Testing has proven it's unnecessary (see #516).

In addition:
- The function has been refactored to avoid recursing for every file (now only directories), which reduces depth of recursion by one, and saves an additional stat on every file, since the file type is already retrieved by `read_dir` so it's redundant to call the more expensive `symlink_metadata()` again on the next recursion.
- `remove_recursively` has been renamed to `remove_dir_recursively` to more clearly reflect it's intended to operate on directories (and for closer parity to the name of `remove_dir_all`).

Closes #516.
GUS-W-11917787.


Malax (approved):
Generally speaking, this PR makes a trade-off I wouldn't usually make (no surprise as the changed code was originally written by me 😄). When dealing with recursion I'd rather have a larger(-ish) binary (and the additional compile time for that code) and be sure that we won't smash the stack. I find debugging stack overflows rather hard, when the stack size varies across operating systems, configurations, optimisation level, etc. Especially when the code relies on external state (the FS in this case) that might be required to reproduce the issue.

In practice, removing the code that prevents stack overflows is _probably fine_ in this context, as demonstrated by your testing. I therefore won't block this from merging.

edmorley:
@Malax I still don't agree that this is an issue in practice (particularly given the constrained environment in which this code runs; in a CNB), and so characterising this as "probably fine" seems unfair. If this genuinely was an issue, then the stdlib would have implemented such protections, or there would be reports of stack overflows when using `remove_dir_all` (there are no such reports I could find). We obviously want to ensure libcnb is as reliable as possible, however it's also important that we not regularly over-engineer things without proof, just because we can.

PR #520, opened 2022-10-18
libcnb-test: Check the exit code of `pack sbom download`

### LINKED ISSUE #519: `TestContext::download_sbom_files()` doesn't check the exit code of `pack sbom download`
`TestContext::download_sbom_files()` runs `pack sbom download` in order to download the SBOM files from the image locally.

It runs pack using the wrapper `run_pack_command()`, however that wrapper returns an `Output` without failing on an unsuccessful status code. This makes sense for the wrapper given that the other consumer of it (the pack build stage) needs to be able to support allowed failures (via `BuildConfig::expected_pack_result()`).

However this does mean:
- every user of `run_pack_command()` needs to be aware it's up to them to check `output.status.success()` (and this isn't clear from its rustdocs)
- consumers that want to display stdout/stderr in the case of an unexpected exit code need to all reinvent the message formatting/handling for that output

So perhaps `run_pack_command()` should do the exit code checking, and accept an additional argument for expected result (that `download_sbom_files()` would always pass as expects success etc)?

GUS-W-11924792.

edmorley:
Previously if the `pack sbom download` command exited non-zero, `TestContext::download_sbom_files` would ignore the error and return an `SbomFiles` referencing an empty directory.

Now, such errors will result in a test panic showing the stdout/stderr from pack - like already occurs for errors during `pack build`.

I tried writing an integration test for this, however wasn't able to find an easy way to make `pack sbom download` fail (it appears to succeed even for images that don't have any SBOMs for example, and I don't want to resort to having to make manual `docker rmi <image>` calls etc). However the implementation is already tested via the `pack build` consumer, so this could be worse.

Fixes #519.
GUS-W-11924792.

### DIFF
--- a/libcnb-test/src/pack.rs
+++ b/libcnb-test/src/pack.rs
@@ -1,6 +1,7 @@
+use crate::PackResult;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
-use std::process::{Command, Output, Stdio};
+use std::process::{Command, Output};
 use tempfile::TempDir;
 
 /// Represents a `pack build` command.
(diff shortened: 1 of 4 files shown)


libcnb-test/src/pack.rs
```diff
@@ -175,22 +176,41 @@ impl From<PackSbomDownloadCommand> for Command {
...
-        .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .spawn()
+/// Runs the given pack command, panicking with user-friendly error messages when errors occur.
+pub(crate) fn run_pack_command<C: Into<Command>>(
+    command: C,
+    expected_result: &PackResult,
+) -> Output {
+    let output = command.into()
+        .output()
```
edmorley, line 185: When using [`Command::Output`]() it's not necessary to set `Stdio::piped()` manually - they are already the default. Using `output()` also saves having to make the manual `wait_with_output()` call.

Malax (approved):
Nice, 👍🏻!

PR #522, opened 2022-10-19
Introduce CachedLayer and CachedLayerTypes

### LINKED ISSUE #371: Make it harder to forget to implement `existing_layer_strategy()` for a layer
Something that came up in our team huddle today was that if a Layer has a `types()` that includes `cache: true`, then if the Layer doesn't implement `existing_layer_strategy()` the layer will silently not be cached, which can cause confusion.

To improve this, we could:
- Add additional more specific traits for common scenarios (eg `CachedLayer`), which can enforce that `existing_layer_strategy()` is implemented
- Add a runtime check to the default `Layer::existing_layer_strategy` implementation to fail if `types()` returns `cache: true`

cc @jonnymacs

schneems:
The current default for `existing_layer_strategy` will recreate a layer on every subsequent build. Effectively that means that if `cached: true` but `existing_layer_strategy` is not set, then the layer is not cached. This is documented in #371.

This PR introduces a CachedLayer trait that implements `Layer` such that the user MUST define `existing_layer_strategy`. 

Sidenote: Is it possible to assert some code doesn't compile? It would be good to encode this desire into the framework so someone doesn't accidentally provide a default implementation in the future, for example.

Fixes #371

### DIFF
--- a/libcnb/src/layer/public_interface.rs
+++ b/libcnb/src/layer/public_interface.rs
@@ -10,6 +10,94 @@ use serde::Serialize;
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
+/// A specialized version of [`LayerTypes`](LayerTypes)
+/// where [`cached`](LayerTypes.cached) is always true.
+#[derive(Debug, Default, Eq, PartialEq)]
+pub struct CachedLayerTypes {
+    pub build: bool,
+    pub launch: bool,
+}
+/// Represents a buildpack layer that uses `cache: true` with the libcnb framework.
+///
+/// Use of this layer trait enforces that `existing_layer_strategy` is implemented.
+///
+/// Buildpack authors implement this trait to define how a layer is created/updated/removed
+/// depending on its state. To use a `Layer` implementation during build, use
+/// [`BuildContext::handle_layer`](crate::build::BuildContext::handle_layer).
+#[allow(unused_variables)]
+pub trait CachedLayer {
+    type Buildpack: Buildpack;
+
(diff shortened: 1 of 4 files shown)


libcnb/src/layer/public_interface.rs
```diff
@@ -10,6 +10,94 @@ use serde::Serialize;
...
+        layer_data: &LayerData<Self::Metadata>,
+    ) -> Result<ExistingLayerStrategy, <Self::Buildpack as Buildpack>::Error>;
+
+    fn update(
+        &self,
+        context: &BuildContext<Self::Buildpack>,
+        layer_data: &LayerData<Self::Metadata>,
+    ) -> Result<LayerResult<Self::Metadata>, <Self::Buildpack as Buildpack>::Error> {
+        // I couldn't figure out how to allow defining this method without making it required
+        // and still defer to the original default implementation in `Layer`.
```
schneems, line 56: I couldn't figure out how to default this to the default Layer implementation of `update`. I'm not sure it's possible. My one thought was to make some function like `default_update` and refactor `LayerType::update` to use it internally, and then use it here as well.

I'm open to other suggestions, or maybe unimplemented here is actually okay?

Same issue with `migrate_incompatible_metadata`

schneems:
We also talked about the possibility of other traits:

- // Cachable: Require existing_layer_strategy is defined
- // Updateable (must be cachable, provably): Require update is defined
- // MetadataMigratable (must be cachable, provably): Require migrate_incompatible_metadata

Malax:

PR #529, opened 2022-11-08
Update imports on `layer_name!`

schneems:
The existing comments are invalid.


schneems:
Assuming this test was running before and was passing, is there a way to add these docs so that they're correct externally to libcnb users?

PR #531, opened 2022-11-20
Initial code generating support `cargo libcnb init .`

schneems:
Why: Make more buildpacks, faster with generators. Before you say "this is unnecessary" 

A) The Ruby buildpack I've been working on for months is missing some of these files
B) Maybe it's unnecessary, but I did it anyway so  might as well use it

How: Use jinja inspired templates (support via the `tera` crate) and a cli (via the `clap` crate) to build dynamic projects that follow project best practices.

Demo:

```
$ cargo run --package libcnb-cargo -- libcnb init  --help
    Finished dev [unoptimized + debuginfo] target(s) in 0.25s
     Running `/Users/rschneeman/Documents/projects/work/libcnb.rs/target/debug/cargo-libcnb libcnb init --help`
Creates a new buildpack from template

Usage: cargo libcnb init [OPTIONS] <DESTINATION>

Arguments:
  <DESTINATION>  Buildpack path

Options:
      --name <NAME_NAMESPACE>      Buildpack name with namespace, must include a slash i.e. `heroku/ruby` [default: todo-namespace/todo-name]
      --detect-file <DETECT_FILE>  Filename in the project's root used to detect if the buildpack will execute or not [default: README.md]
      --license <LICENSE>          Generated license for the project [default: mit] [possible values: mit, bsd3]
      --copyright <COPYRIGHT>      Name of copyright holder for the license [default: "<TODO license holder name>"]
      --coc <CONDUCT>              Defines the code of conduct used in the generated project [default: contributor_covenant] [possible values: salesforce, contributor_covenant]
  -f, --force                      When enabled will skip all prompts
  -h, --help                       Print help information
```

Beyond a simple generator, this command also supports file collision resolution via STDIN prompts to the user [inspired by the thor library that Ruby on Rails uses](https://github.com/rails/thor/blob/e4907fdd663d0a1ff51d18eb1827552ee61300a3/lib/thor/shell/basic.rb#L287). Example:

```
$ cargo run --package libcnb-cargo -- libcnb init /tmp/lol
    Finished dev [unoptimized + debuginfo] target(s) in 0.46s
     Running `/Users/rschneeman/Documents/projects/work/libcnb.rs/target/debug/cargo-libcnb libcnb init /tmp/lol`
Overwrite /tmp/lol/.gitignore? (enter 'h' for help) [Ynaqdh]
> n
INFO - Identical /tmp/lol/cargo.toml
INFO - Identical /tmp/lol/CHANGELOG.md
INFO - Identical /tmp/lol/SECURITY.md
Overwrite /tmp/lol/src/main.rs? (enter 'h' for help) [Ynaqdh]
> y
INFO - Writing /tmp/lol/src/main.rs
INFO - Identical /tmp/lol/package.toml
INFO - Identical /tmp/lol/.github/workflows/check_changelog.yml
INFO - Identical /tmp/lol/.github/workflows/ci.yml
INFO - Identical /tmp/lol/CODE_OF_CONDUCT.md
INFO - Identical /tmp/lol/tests/integration_test.rs
INFO - Identical /tmp/lol/LICENSE.txt
INFO - Identical /tmp/lol/.github/dependabot.yml
Overwrite /tmp/lol/README.md? (enter 'h' for help) [Ynaqdh]
> n
INFO - Identical /tmp/lol/tests/fixtures/hello_world/README.md
Overwrite /tmp/lol/buildpack.toml? (enter 'h' for help) [Ynaqdh]
> d
--- /tmp/lol/buildpack.toml	2022-11-19 20:45:25.000000000 -0600
+++ /var/folders/l9/w5ggmcjd28d57p4rwb1m7kph0000gp/T/.tmpCKtkbR/new_contents.txt	2022-11-20 17:13:49.000000000 -0600
@@ -1,7 +1,7 @@
 api = "0.8"

 [buildpack]
-id = "lol/lol"
+id = "todo-namespace/todo-name"
 version = "0.1.0"
 name = "My Buildpack"


Retrying...
Overwrite /tmp/lol/buildpack.toml? (enter 'h' for help) [Ynaqdh]
```


.github/workflows/ci.yml
```diff
@@ -46,7 +46,7 @@ jobs:
       - name: Rust Cache
         uses: Swatinem/rust-cache@v2.2.0
       - name: Run unit tests
-        run: cargo test --all-features
+        run: RUST_BACKTRACE=1 cargo test --all-features
```
edmorley, line 49: If you'd like to adjust the CI config , please can you open a separate PR for this? We've intentionally not enabled `RUST_BACKTRACE` for the reasons described here:
https://github.com/heroku/procfile-cnb/pull/75

Also worth noting for integration tests it won't do what you think it does, since it won't be propagated inside the container that runs the build/test.

edmorley (commented):
Hi! 

I know this is a draft, but I was curious so took a look and thought I've leave these comments before I forget.

I do like the idea of having a command to (a) save having to create a buildpack skeleton manually, (b) steer people towards best practices (eg Dependabot, optimal CI setup etc), so I think this is a feature we want (at least at some point).

It seems this feature could even be used to simplify some of the tutorial in the README - ie: people could init the skeleton and then the README talk them through it, rather than having to manually create files.

I do have a bit of a hesitation about including any Salesforce specific files however. 

Salesforce open source projects have a specific onboarding process / tool (the OSS Request site), which covers what files need to be created in order to get the request approved. The templates for the files themselves live at:
https://github.com/salesforce/oss-template

The reason the Ruby buildpack repo is missing those files, is since it hasn't gone through that process, as (a) it predated them turning off the ability for us to create new open source repositories outside of the OSS tool, (b) it hasn't been submitted manually since then.

As such, I would lean towards keeping the template free of Salesforce-specific content.

I would also lean towards keeping the template as simple as possible - eg: not using `libherokubuildpack` and perhaps being closer to the content in the README tutorial? The examples folder in this repository might be a better place to put more advanced content?

Lastly, I wonder if there is a way to ensure the template `Cargo.toml` stays up to date with respects to dependency versions? (eg dynamically look them up and template them). It just seems suboptimal if a newly created template is already using out of date things :-)

schneems:
> Lastly, I wonder if there is a way to ensure the template Cargo.toml stays up to date with respects to dependency versions? (eg dynamically look them up and template them). It just seems suboptimal if a newly created template is already using out of date things :-)

100% this. I'm not sure the best way to do that, but I want it. I think we can grab it from the other packages locally somehow. I had a similar question about changing things like the version of the API we default to, currently it's hardcoded in there `api = "0.8"`.

> I do have a bit of a hesitation about including any Salesforce specific files however.

Maybe worth talking about this more in person. Ultimately the main consumers of this will be our team. At least at first I want to remove barriers to making new buildpacks when it's the best solution. Right now the only salesforce specific thing in there is a Salesforce CoC. I had previously considered some kind of a plugin system.

The `bundle gem <name>` command is one inspiration in addition to `rails new <name>`. The `bundle gem` command has long term global storage for config (https://bundler.io/man/bundle-config.1.html) so if you tell it to use rspec instead of minitest you only have to tell it once. We could combine the two, designating a local config folder like `~/.libcnb/config` where you could put templates like `~/.libcnb/config/templates/code_of_conduct/salesforce.md` then you could possibly configure your settings to make that the default.

Anyway, getting off the rails a bit. Probably better to speak about it in person. I think my stance is: I want to favor pragmatism and speed of development. I'm for the goal of making this company/salesforce agnostic, if it's easy to do now, we should do it now. Otherwise i'm not against iterating towards a better solution later.

.github/workflows/ci.yml
schneems: Sounds good. I can't figure out why current tests are failing. I'll move my debugging logic to it's own commit so it's easier to remove.

schneems: Tests were failing because my template was `cargo.toml.jinja` instead of `Cargo.toml.jinja`. It passes locally because mac has a case insensitive filesystem. The reason they were getting no output was because I was following a similar pattern to the other CLI method and calling process::exit() manually. Unlike a panic, it won't output all the stdout and stderr that was captured in the test.

I refactored it to instead return a Result, then I used a wrapper function to exit if an Err is returned. we call it with the wrapper function in the CLI and without it in the tests.

Malax:
@schneems Thanks for thinking about the scaffolding problem! Have you considered doing this via https://github.com/buildpacks/rfcs/pull/212? What is missing upstream to implement it via pack's scaffolding? I think adding it there has the additional benefit serving as advertisement for libcnb.rs, allowing buildpack authors to discover it when they want to write a buildpack.

In any case, adding some sort of scaffolding via `pack` seems worth doing. And I'd rather only maintain one template than two so I'm in favour of moving this upstream. :)

I've similar concerns around Salesforce specifics in the template, they're not required files for libcnb.rs, but required to meet company policies. If possible, I want libcnb.rs to be as generic as possible so it has greater chances for adoption outside of Salesforce.

schneems:
> Have you considered doing this via https://github.com/buildpacks/rfcs/pull/212?

@Malax  I've not. I'm but I just read through it. I've got some thoughts below. TLDR; I don't love the RFC. If I had to pick just supporting one I would pick moving forwards with this PR, though it's my code-baby so I'm clearly biased. An alternative way to move forward could be to create a template and use it and see how it turns out. 

## Thoughts on the RFC

- I don't love the syntax of calling --arg followed by a key/value pair:

```
$ pack buildpack create --arg ProjectDirectory=bash_buildpack --arg BuildpackApi=0.8 --arg BuildpackID=example/bash --arg BuildpakStacks=io.buildpacks.samples.stacks.bionic
```

- The templates seem to be very static. In the above example I'm using user inputs to do things like define structs so that we don't have hundreds of `HelloWorldBuildpack` buildpacks out there. 

- There's silent failures if there's a typo in the buildpack or template:

> [from the RFC] When a source file contains a Go `text/template` style expression and the variable name does not appear in `prompts.toml`, then the `text/template` style expression is not replaced.  For example, if a `README.md` file contains the expression `{{.Example}}` and `Example` is not a variable defined in `prompts.toml`, then the string `{{.Example}}` is not replaced in `README.md`.

- Templates are not advertised, the user has to discover them i.e. pass in the full URL `pack buildpack create --help --template=<url>` which I'm never going to remember.  Not to mention it would mean us getting another OSS approval for yet another repo like `github.com/heroku/rust-cnb-template`. 

If we wanted to make it more discoverable, we would have to also add docs/tutorial to the project docs. If docs are the only thing getting us noticed, I don't know why we shouldn't have docs that use our tooling directly instead of filtering through an intermediate templating language. 

- No template versioning. If templates are pulled from URLs then when it updates everyone who uses that template (say in CI) could break. We could hack on our own versioning via github releases, but shipping in a crate gives it to us for free.
- Error experience. No CLI hints. This could probably be added.
- Ties us to someone else's interface. If pack wants to deprecate or change something with the interface then we've got to update all our docs. That's also true on our side, but at least we have more agency and visibility into the impacts.
- Future features to templates in pack require RFCs to the project and/or the ability to code it in Golang (which our team maintains loose/fragmented competency for at best).

Malax:

Issue #532, opened 2022-11-21
Clippy spending a lot of time in CI checking dependencies? 


schneems:
Unit tests actually finish much faster than running clippy. It looks like it's checking dependencies by mistake somehow?

<img width="1278" alt="image" src="https://user-images.githubusercontent.com/59744/202951231-b7f95647-b644-45ba-9a70-e9c4ee568344.png">

edmorley:
The cache restore step of that `lint` job says:

```
... Restoring cache ...
No cache found.
```

...whereas the `unit-test` job found/restored a cache.

This explains the time difference.

The reason for this is likely that:
- the cache is only stored when the job succeeds, and this was the first time in four pushes to the branch where linting didn't fail, whereas the test job was passing earlier, so had already stored its cache
- GitHub Actions discards any caches not accessed for 7 days, and the last GitHub Actions run on this repo was 11 days ago

In terms of things to bear in mind with the caching:
- the cache is also invalidated every time the Rust version changes (which happens every 6 weeks)
- for this repo specifically, since it's a library not an application, we intentionally don't commit the `Cargo.lock`, and as such it means the Rust cache plugin can't as efficiently manage the cache (though it's still a net benefit)

Out of curiosity, do you have linting enabled in your IDE? And specifically it set to run `clippy` not `check` as the lint command? I would presume not given the four pushes to get linting to pass? (If using VSCode the default is `check` - and so why https://salesforce.quip.com/geXbAMwRn5Pb suggests changing it.)

schneems:
> Out of curiosity, do you have linting enabled in your IDE? And specifically it set to run clippy not check as the lint command? I would presume not given the four pushes to get linting to pass? (If using VSCode the default is check - and so why https://salesforce.quip.com/geXbAMwRn5Pb suggests changing it.)

My local rust version was older than CI so clippy was passing. It took me a few pushes to figure that out.  My workflow is generally running this https://github.com/schneems/dotfiles/blob/main/bash/aliases#L8 though I tend to push before remembering to run it. So I push, run it, then push again after I've fixed. 

Looking into updating the IDE again.

PR #533, opened 2022-11-21
Convert "rust learning" quip doc to an mdbook

schneems:
We reference the "rust learning doc" frequently, for example https://github.com/heroku/libcnb.rs/issues/532#issuecomment-1321608026. It's a good resource and I've long talked about formatting it for public consumption. 

This PR introduces a "rust quickstart" book. The goal is to have a public living document where we can record helpful resources, tools, and quality of life development tips (like IDE settings). The target is the existing development team or new developers coming to Rust for the first time wanting to build a buildpack with libcnb.

One spoken worry is that this will expand until it becomes a bloated or generic tutorial: I tried to outline the vision in the `introduction.md` page and make it clear that it's not a place to link to the blog post you just published. I think it should be a curated resource aimed at brevity and targeted at giving people a guiding star of sorts to learning enough Rust to work with libcnb.
 
On the other end of the spectrum there's a thought that it is simply unneeded since there are other resources. This guide is ours. There are many others like it, but this is ours.

The value is in curation. If it proves to be a drain on maintainer resources, we can always delete it. If it proves to be not useful, we can also remove it.


Malax:

PR #535, opened 2022-12-07
Add `CommandExt` for working with `std::process::Command` output streams

Malax:
While working on the Gradle buildpack, I had the need to prefix `gradlew` output so the user can clearly distinguish between `gradlew` and buildpack output. While implementing a solution, @dzuelke mentioned that he has the same need and also some additional ones. So I went the extra mile, made my solution generic and added docs so that it is easy to use. Implementing this is less trivial than it might sound so I thought it's a great fit for `libherokubuildpack` (or whatever it will be called in the future)

It all revolves around a new trait: `CommandExt` which extends `std::process:Command` with (currently) two helpers that allow writing the unbuffered output of `stdout` and `stderr` to a `std::io::Writer`.

This is the main hook to add more complex logic like `tee`ing output to two different locations, prefixing lines and live-modifying streams using standard `String` functions. These helpers are contained in the new `write` module which has the aforementioned helpers. These can be used without `CommentExt` and are standalone. Their only interface is `std::io::Write`.

Here is an example how this can be used:

```rust
use libherokubuildpack::command::CommandExt;
use libherokubuildpack::write::tee;
use std::fs;
use std::process::Command;

let logfile = fs::File::open("log.txt").unwrap();

let exit_status = Command::new("date")
    .spawn_and_write_streams(tee(std::io::stdout(), &logfile), std::io::stderr())
    .and_then(|mut child| child.wait())
    .unwrap();
```

The exposed API has proper rust docstrings. For detailed info and more usage examples refer to those.

Closes GUS-W-12188139

### DIFF
--- a/libherokubuildpack/Cargo.toml
+++ b/libherokubuildpack/Cargo.toml
@@ -23,10 +23,11 @@ log = ["dep:termcolor"]
 tar = ["dep:tar", "dep:flate2"]
 toml = ["dep:toml"]
 fs = ["dep:pathdiff"]
-command = ["write", "dep:crossbeam"]
+command = ["write", "dep:crossbeam-utils"]
 write = []
 
 [dependencies]
+crossbeam-utils = { version = "0.8.2", optional = true }
 flate2 = { version = "1.0.24", optional = true }
 libcnb = { workspace = true, optional = true }
 pathdiff = { version = "0.2.1", optional = true }
@@ -36,7 +37,6 @@ termcolor = { version = "1.1.3", optional = true }
 thiserror = { version = "1.0.35", optional = true }
 toml = { version = "0.5.9", optional = true }
 ureq = { version = "2.5.0", optional = true }
-crossbeam = { version = "0.8.2", optional = true }
 
 [dev-dependencies]
 tempfile = "3.3.0"
--- a/libherokubuildpack/Cargo.toml
+++ b/libherokubuildpack/Cargo.toml
@@ -15,7 +15,7 @@ include = ["src/**/*", "LICENSE", "README.md"]
 all-features = true
 
 [features]
-default = ["download", "digest", "error", "log", "tar", "toml", "fs"]
+default = ["command", "download", "digest", "error", "log", "tar", "toml", "fs", "write"]
 download = ["dep:ureq", "dep:thiserror"]
 digest = ["dep:sha2"]
 error = ["log", "dep:libcnb"]
@@ -23,6 +23,8 @@ log = ["dep:termcolor"]
 tar = ["dep:tar", "dep:flate2"]
 toml = ["dep:toml"]
 fs = ["dep:pathdiff"]
+command = ["write", "dep:crossbeam"]
+write = []
 
 [dependencies]
 flate2 = { version = "1.0.24", optional = true }
@@ -34,6 +36,7 @@ termcolor = { version = "1.1.3", optional = true }
 thiserror = { version = "1.0.35", optional = true }
 toml = { version = "0.5.9", optional = true }
 ureq = { version = "2.5.0", optional = true }
+crossbeam = { version = "0.8.2", optional = true }
 
 [dev-dependencies]
 tempfile = "3.3.0"
--- a/libherokubuildpack/src/command.rs
+++ b/libherokubuildpack/src/command.rs
@@ -127,7 +127,7 @@ fn write_child_process_output<OW: Write + Send, EW: Write + Send>(
         let stdout_copy_result = stdout_copy_thread.map_or_else(|| Ok(0), join_and_unwind_panic);
         let stderr_copy_result = stderr_copy_thread.map_or_else(|| Ok(0), join_and_unwind_panic);
 
-        // Return the first error from either Result or the child process value
+        // Return the first error from either Result, or the child process value
         stdout_copy_result.and(stderr_copy_result).map(|_| child)
     }))
 }
--- a/libherokubuildpack/src/command.rs
+++ b/libherokubuildpack/src/command.rs
@@ -1,5 +1,5 @@
 use crate::write::tee;
-use crossbeam::thread::ScopedJoinHandle;
+use crossbeam_utils::thread::ScopedJoinHandle;
 use std::io::Write;
 use std::{io, process, thread};
 use std::{mem, panic};
@@ -117,7 +117,7 @@ fn write_child_process_output<OW: Write + Send, EW: Write + Send>(
     // almost immediate join calls into account) and therefore requires that data used in a thread
     // lives forever. To avoid requiring 'static lifetimes for the writers, we use crossbeam's
     // scoped threads here. This enables writers that write, for example, to a mutable buffer.
-    unwind_panic(crossbeam::scope(|scope| {
+    unwind_panic(crossbeam_utils::thread::scope(|scope| {
         let stdout_copy_thread = mem::take(&mut child.stdout)
             .map(|mut stdout| scope.spawn(move |_| std::io::copy(&mut stdout, &mut stdout_writer)));
 
--- a/libherokubuildpack/src/command.rs
+++ b/libherokubuildpack/src/command.rs
@@ -1,0 +1,185 @@
+use crate::write::tee;
+use crossbeam::thread::ScopedJoinHandle;
+use std::io::Write;
+use std::{io, process, thread};
+use std::{mem, panic};
+
+/// Extension trait for [`process::Command`] that adds functions for use within buildpacks.
+pub trait CommandExt {
+    /// Spawns the command process and sends the output of stdout and stderr to the given writers.
+    ///
+    /// This allows for additional flexibility when dealing with these output streams compared to
+    /// functionality that the stock [`process::Command`] provides. See the [`write`](crate::write)
+    /// module for [`std::io::Write`] implementations designed for common buildpack tasks.
+    ///
+    /// This function will redirect the output unbuffered and in parallel for both streams. This
+    /// means that it can be used to output data from these streams while the command is running,
+    /// providing a live view into the process' output. This function will block until both streams
+    /// have been closed.
+    ///
+    /// # Example:
+    /// ```no_run
+    /// use libherokubuildpack::command::CommandExt;
+    /// use libherokubuildpack::write::tee;
+    /// use std::fs;
+    /// use std::process::Command;
+    ///
+    /// let logfile = fs::File::open("log.txt").unwrap();
+    /// let exit_status = Command::new("date")
+    ///     .spawn_and_write_streams(tee(std::io::stdout(), &logfile), std::io::stderr())
+    ///     .and_then(|mut child| child.wait())
+    ///     .unwrap();
+    /// ```
+    fn spawn_and_write_streams<OW: Write + Send, EW: Write + Send>(
+        &mut self,
+        stdout_write: OW,
+        stderr_write: EW,
+    ) -> io::Result<process::Child>;
+
+    /// Spawns the command process and sends the output of stdout and stderr to the given writers.
+    ///
+    /// In addition to what [`spawn_and_write_streams`](Self::spawn_and_write_streams) does, this
+    /// function captures stdout and stderr as `Vec<u8>` and returns them after waiting for the
+    /// process to finish. This function is meant as a drop-in replacement for existing
+    /// `Command:output` calls.
+    ///
+    /// # Example:
+    /// ```no_run
+    /// use libherokubuildpack::command::CommandExt;
+    /// use libherokubuildpack::write::tee;
+    /// use std::fs;
+    /// use std::process::Command;
+    ///
+    /// let logfile = fs::File::open("log.txt").unwrap();
+    /// let output = Command::new("date")
+    ///     .output_and_write_streams(tee(std::io::stdout(), &logfile), std::io::stderr())
+    ///     .unwrap();
+    ///
+    /// // Return value can be used as with Command::output, but the streams will also be written to
+    /// // the given writers.
+    /// println!(
+    ///     "Process exited with {}, stdout: {:?}, stderr: {:?}",
+    ///     output.status, output.stdout, output.stderr
+    /// );
+    /// ```
+    fn output_and_write_streams<OW: Write + Send, EW: Write + Send>(
+        &mut self,
+        stdout_write: OW,
+        stderr_write: EW,
+    ) -> io::Result<process::Output>;
+}
+
+impl CommandExt for process::Command {
+    fn spawn_and_write_streams<OW: Write + Send, EW: Write + Send>(
+        &mut self,
+        stdout_write: OW,
+        stderr_write: EW,
+    ) -> io::Result<process::Child> {
+        self.stdout(process::Stdio::piped())
+            .stderr(process::Stdio::piped())
+            .spawn()
+            .and_then(|child| write_child_process_output(child, stdout_write, stderr_write))
+    }
+
+    fn output_and_write_streams<OW: Write + Send, EW: Write + Send>(
+        &mut self,
+        stdout_write: OW,
+        stderr_write: EW,
+    ) -> io::Result<process::Output> {
+        let mut stdout_buffer = vec![];
+        let mut stderr_buffer = vec![];
+
+        self.spawn_and_write_streams(
+            tee(&mut stdout_buffer, stdout_write),
+            tee(&mut stderr_buffer, stderr_write),
+        )
+        .and_then(|mut child| child.wait())
+        .map(|status| process::Output {
+            status,
+            stdout: stdout_buffer,
+            stderr: stderr_buffer,
+        })
+    }
+}
+
+fn write_child_process_output<OW: Write + Send, EW: Write + Send>(
+    mut child: process::Child,
+    mut stdout_writer: OW,
+    mut stderr_writer: EW,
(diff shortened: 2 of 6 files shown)


libherokubuildpack/src/command.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// );
+    /// ```
+    fn output_and_write_streams<OW: Write + Send, EW: Write + Send>(
+        &mut self,
+        stdout_write: OW,
+        stderr_write: EW,
+    ) -> io::Result<process::Output>;
+}
+
+impl CommandExt for process::Command {
```
schneems, line 72: How does this not violate orphan rules? I thought we couldn't implement a trait for a struct we don't own (broad strokes).

libherokubuildpack/src/command.rs
```diff
@@ -0,0 +1,185 @@
...
+
+        let stderr_copy_thread = mem::take(&mut child.stderr)
+            .map(|mut stderr| scope.spawn(move |_| std::io::copy(&mut stderr, &mut stderr_writer)));
+
+        let stdout_copy_result = stdout_copy_thread.map_or_else(|| Ok(0), join_and_unwind_panic);
+        let stderr_copy_result = stderr_copy_thread.map_or_else(|| Ok(0), join_and_unwind_panic);
+
+        // Return the first error from either Result or the child process value
+        stdout_copy_result.and(stderr_copy_result).map(|_| child)
+    }))
```
schneems, line 132: I basically have no clue what's happening here. I'm not familiar with crossbeam. I'm aware it's a parallelism-ish library. 

To write this out it looks like we're using crossbeam scope which allows us to spawn scoped threads  https://docs.rs/crossbeam/0.7.3/crossbeam/fn.scope.html. We're moving the memory of stdou and stderr into those threads and essentially just copying whatever goes into stdout into whatever is passed in to the stdout_writer. I'm assuming this is blocking (inside of the thread).

Then we do some stuff which essentially boils down to joining the threads.

I'm not entirely following how this last line works, but it compiles so i'm sure it does:

```rust
        // Return the first error from either Result or the child process value
        stdout_copy_result.and(stderr_copy_result).map(|_| child)
```

schneems: Notes for self: My naieve implementation is here https://github.com/heroku/buildpacks-ruby/blob/f33fd189414507645d563ceea675157c78847fc5/commons/src/env_command.rs#L330-L376. I used channels to get data out of the threads without lifetime issues.

schneems (commented):
Still working on the review. Wanted to leave these questions/notes for you. I'll pick up the write.rs and finish up later.

Overall looks good. I like having some of this general purpose goodness in shared libraries.

libherokubuildpack/src/command.rs
Malax: Since this crate defines the `CommandExt` trait it's fine. The crate implementing a trait must either define the trait _or_ the type it's implemented for.

libherokubuildpack/src/command.rs
Malax:
```
// Return the first error from either Result or the child process value
stdout_copy_result.and(stderr_copy_result).map(|_| child)
```

The [`and`](https://doc.rust-lang.org/std/result/enum.Result.html#method.and) does the following:
- If `stdout_copy_result` is an `Err`, return it and do nothing with `stderr_copy_result`.
- If `stdout_copy_result` is an `Ok`, return `stderr_copy_result` as is (so either an `Err` or an `Ok`).
- This means, after the `and`, we will have either the `Err` of `stdout_copy_result` or the `Ok` or `Err` of `stderr_copy_result`.

The `map` is a regular map, but in this case we actually don't mind what's in the `Ok`. If the `Result` after the `and` is `Ok` (meaning that both succeeded), we put `child` into the `Ok`, dropping whatever was in the `Ok` previously.

schneems: > With crossbeam, we can convince the compiler to believe us that OW and EW are only referenced by the threads until the crossbeam scope ends.

That was my understanding as well. I think it's commented well enough. I'm just not familiar with the API so it was more of a talk-aloud. If my comments are generally correct there's nothing to update. Thanks for taking the time to add those annotations

libherokubuildpack/src/command.rs
```diff
@@ -0,0 +1,185 @@
...
+    ///     .unwrap();
+    ///
+    /// // Return value can be used as with Command::output, but the streams will also be written to
+    /// // the given writers.
+    /// println!(
+    ///     "Process exited with {}, stdout: {:?}, stderr: {:?}",
+    ///     output.status, output.stdout, output.stderr
+    /// );
+    /// ```
+    fn output_and_write_streams<OW: Write + Send, EW: Write + Send>(
```
schneems, line 65: I don't love these method names. I know they match Command's but I find them confusing. 

Looking at the code both `output` and  `spawn` seem like generic actions rather than behaviors.

Would you be open to `capture_output_and_write_streams` or `capture_and_write_streams`. Maybe just `capture_and_stream`.

schneems (approved):
Added some additional comments. 

- One question about whether we need Clone or why we might not want it. 
- Another comment related to naming. I would like to add `capture` to `output_and_write_streams` or otherwise indicate what it's doing as I feel the existing naming of `spawn` or `output` isn't very descriptive. I would prefer it but don't want to block on it if you feel strongly otherwise.

libherokubuildpack/src/write.rs
```diff
@@ -0,0 +1,212 @@
...
+/// encountered and then applies a given mapping function to the data before passing the result to
+/// the wrapped writer.
+///
+/// See the [`mappers`] module for a collection of commonly used mappers.
+pub fn line_mapped<W: io::Write, F: (Fn(Vec<u8>) -> Vec<u8>) + Sync + Send + 'static>(
+    w: W,
+    f: F,
+) -> MappedWrite<W> {
+    mapped(w, NEWLINE_ASCII_BYTE, f)
+}
```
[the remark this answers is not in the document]
Malax: Not sure what you mean by this comment. Yeah, it does with with both `\n` and `\r\n` so we're good when we think about Windows and *NIX. The Rustdoc goes a bit into detail, explaining what it does under the hood. If there is some other ecosystem where lines are separated by some other delimiter this won't work, correct. 

Do you want me to change the name of this function to smth. like `newline_delimited_map`?

libherokubuildpack/src/write.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+/// A mapped writer that was created with the [`mapped`] or [`line_mapped`] function.
+pub struct MappedWrite<W: io::Write> {
+    inner: W,
+    marker_byte: u8,
+    buffer: Vec<u8>,
+    mapping_fn: Arc<dyn (Fn(Vec<u8>) -> Vec<u8>) + Sync + Send>,
+}
```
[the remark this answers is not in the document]
Malax: I guess YAGNI is the way to go here - if there are no obvious benefits of implementing a specific trait (either manually or via `derive`) I would wait until we need it so we a) don't add things we don't need and b) know _how_ we should implement them. When thinking about `Debug` for example, what do we want, how should we represent the function? Do we need to?

libherokubuildpack/src/write.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// use libherokubuildpack::write::mappers::map_utf8_lossy;
+    /// use libherokubuildpack::command::CommandExt;
+    /// use std::process::Command;
+    ///
+    /// Command::new("date")
+    ///     .spawn_and_write_streams(
+    ///         line_mapped(
+    ///             std::io::stdout(),
+    ///             map_utf8_lossy(|string| string.replace("foo", "bar")),
+    ///         ),
```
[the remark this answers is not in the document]
Malax: `output_and_write_streams` would contain the _original_ output. The manipulations via the `Write`s happens separately and _in addition_. This ensures `output_and_write_streams` is a drop-in replacement for `output`. `CommandExt` shouldn't really be a concern of this module, it only contains `Write` implementations that can be used wherever. I use `CommandExt` in the Rustdoc as an example because that is what they are designed for, but the implementation itself shouldn't be coupled to it.

If you want to capture the modified stream, a `&mut [u8]` is `Write`. So you can easily "pipe" it into a variable of that type and inspect it after the command finished.

libherokubuildpack/src/command.rs
Malax: Hm, I'm quite happy with the naming of these - especially compared to earlier iterations I had. 😂 

I think it's beneficial to name them after the `Command` functions they emulate (they are drop-in replacements). I think of these functions and an additional "tap" for the output streams - in the form of the `Write` interface. 

Do you think that `Command::spawn` and `Command::output` should be renamed to `Command::capture` and `Command::capture_output` too? I think at least `capture` is a bit misleading. In any case, I don't see the benefit of inventing another name for the same thing that users then have to lean in addition - no matter how "bad" the original naming is.

libherokubuildpack/src/write.rs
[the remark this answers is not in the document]
schneems: I'm pushing up my glasses and squinting while thinking "well actually" in a nasally voice while writing the original comment. I think we shouldn't need to change the name `line_mapped` makes sense. If we wanted to make a change I think specifying "newline" in the rustdoc would be the way to go.

libherokubuildpack/src/command.rs
schneems: TLDR; I agree but it took awhile to get there

> Do you think that Command::spawn and Command::output should be renamed to Command::capture and Command::capture_output too? 

I care about behaviors `Command::capture_output` would be better than `Command::output` imho. I guess you could advocate for `Command::output` being named because it GIVES you the output. But if that's the case `Command::spawn` makes no sense because it does not GIVE you the `spawn` (what does that even mean?). It gives you a child which I guess is a spawned process https://doc.rust-lang.org/std/process/struct.Child.html. So, meh...maybe that makes sense. But I've been using `Command` for a long time and it took me until now to make that connection. 

I think that it __technically__ makes sense, but only when you squint. I have a comparison to these functions: 

https://github.com/heroku/buildpacks-ruby/blob/schneems/fresh-start-rust/commons/src/in_app_dir_cache.rs

While technically `copy_app_path_to_cache` and `move_app_path_to_cache` are correctly named but the subtle difference is easy to miss, so I renamed one of them `destructive_move_app_path_to_cache` so it is PAINFULLY obvious what you're getting it when reading it.

So i think to sum up my argument: `output` and `spawn` might be technically correct and an okay name, but I don't think they communicate behavior or intent to someone coming from a different language or context. I feel that command is actually too low level of a primitive (hence why I wrote my own wrapper struct that none of you like). 

> I think at least capture is a bit misleading.

Can you say more here?

> I don't see the benefit of inventing another name for the same thing that users then have to learn in addition - no matter how "bad" the original naming is.

I'm thinking about it and using my own arguments...I agree. The difference between capturing and not capturing would be detected at code time as the developer would realize they can't access the stdout/stderr. Versus in the stream/no-stream case the difference to customers is huge. However in both our methods we are streaming and it explicitly says so in the method name `*_and_write_streams`

I'm leaving this long comment, because it's how I derived this conclusion and want to leave it as a note to myself.

libherokubuildpack/src/write.rs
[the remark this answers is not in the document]
edmorley: > So in general I think we should at least consider implementing these derives by default if we can: Clone, Default, PartialEq, PartialOrd, Hash, Eq, and Ord. Less because we'll directly need those things on this struct, and more that any struct that contains this one is blocked from deriving these traits if it contains our struct.

Some thoughts:
- In the general case, yes we should at least consider whether some of these are useful when adding new code.
- `and more that any struct that contains this one is blocked from deriving` -> good point, though in this specific case, why would they need to put the mapper in their own struct?
- Adding derives isn't free - it has an impact on compile times (there's been a bunch of performance work, eg [here](https://nnethercote.github.io/2022/07/20/how-to-speed-up-the-rust-compiler-in-july-2022.html), trying to reduce the size of the generated code, since in some cases it's quite verbose and gives the compiler extra work to do). This shouldn't stop us using them if there is a genuine chance someone will need them, but we should probably not default to adding all 7 of those derive attributes to every struct without thinking about it at least briefly case by case.

libherokubuildpack/src/write.rs
```diff
@@ -0,0 +1,212 @@
+use std::io;
+use std::mem;
+use std::sync::Arc;
+
+/// Constructs a writer that buffers written data until given marker byte is encountered and
+/// then applies a given mapping function to the data before passing the result to the wrapped
```
edmorley, line 6:
```suggestion
/// Constructs a writer that buffers written data until the given marker byte is encountered and
/// then applies the given mapping function to the data before passing the result to the wrapped
```

libherokubuildpack/Cargo.toml
```diff
@@ -34,6 +36,7 @@ termcolor = { version = "1.1.3", optional = true }
 thiserror = { version = "1.0.35", optional = true }
 toml = { version = "0.5.9", optional = true }
 ureq = { version = "2.5.0", optional = true }
+crossbeam = { version = "0.8.2", optional = true }
```
edmorley, line 39: The `crossbeam` crate itself is just a re-export of half a dozen other crossbeam crates:
https://github.com/crossbeam-rs/crossbeam#crates

It looks like the `crossbeam` features used in this PR are from the `crossbeam-utils` crate only.

Using that crate standalone seems to be a supported/expected use-case, given the docs:
https://crates.io/crates/crossbeam-utils

Please could we switch to `crossbeam-utils` for now, to reduce the number of dependencies that are downloaded and compiled but never needed? We can always switch back to the `crossbeam` crate if we ever start using several of the `crossbeam-*` crates :-)

Also pedantic nit: The dependencies list was previously alphabetical, now it's not 😆

edmorley (commented):
Thank you for adding this!

libherokubuildpack/Cargo.toml
Malax: Changed in [49ac7ea](https://github.com/heroku/libcnb.rs/pull/535/commits/49ac7ea4752b57a81d908cef478f230d109834fe)

libherokubuildpack/src/write.rs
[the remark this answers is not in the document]
Malax: Isn't this already in the rustdocs? Or do you mean something else?
> Constructs a writer that buffers written data until an ASCII/UTF-8 newline byte (`0x0A`)

libherokubuildpack/src/command.rs
Malax: I'm marking this is "resolved", LMK if I misunderstood your comment and feel free to re-open! :)

edmorley (approved):

libherokubuildpack/src/write.rs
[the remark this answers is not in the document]
Malax: Looking at the standard library for similar types that serve as intermediate types (often named after the function that creates them: `FlatMap`, `Child`, etc.), I changed my mind and think `Clone` and `Debug` make sense here. Except for `Debug` in the case of `MappedWrite`, they can be automatically derived and the `Debug` implementation isn't too bad.

I agree with the "lost momentum" argument. :)

Thanks for raising this! Added in 86462bee1558713fc4f757d1ad2240b4a6ce4248.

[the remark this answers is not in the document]
schneems: Thank you for this @Malax ❤️ . 

> Adding derives isn't free

I meant in terms of programmer/maintainer time. But I hadn't really considered binary size or compile time. In my head I'm still basically laughing at how much faster even "bad" rust code is compared to carefully written Ruby code. It would be nice to get a sense of the cost of that tradoff (size and compile time). I'm not sure how we could do it.

Maybe we could remove deriving Debug from all structs in libcnb, clear the caches and time the compile time and check binary size? For me, i'm in a mode of usually wanting to optimize programmer productivity and experience first and then really only caring about perf and size second. I care a lot about performance in Ruby land mostly as a factor of it ultimately affecting productivity and experience. I'm curious to brainstorm ways to better tune intuition on these tradeoffs.

I think it's a good point, I want to know more. I always want to understand the impact of my (coding) actions.

> why would they need to put the mapper in their own struct?

More coming from a place of assuming people will build structs (and enums) from pretty much any other struct that exists. In this case, I've stored a bunch of stuff for command execution in a struct in my buildpack.

If there's a good reason to NOT allow someone to use or store a struct in their own code, we should probably annotate/document that case (or make the structs private all together). Otherwise as library implementers we need to write for what people might likely do, even if we've not got concrete ideas for why. In this case I don't think leaving out Clone or Debug would make someone second guess their code "hmm, maybe I shouldn't store this in a struct/enum" and instead they would likely jump straight to opening an issue or a PR. 

In general it's good to be opinionated and also have a good sense for what we want to encourage people to do. I think we also need to balance that with some general guidelines about reasonable expectations of what our users will and won't reasonably do.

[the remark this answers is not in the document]
edmorley: I'm not talking about not adding `Debug` (of all the traits, that seems like a no-brainer to add; my post was about the "7 traits" aspect, per: "we should probably not default to adding all 7 of those derive attributes to every struct without thinking about it at least briefly case by case"), nor am I talking about binary size or runtime performance (I only mention compile times). 

I know it's not intentional, but it feels like the response above is nearing straw man territory? :-)

PR #543, opened 2023-01-09
Expose `store.toml` contents in `BuildContext`

Malax:
Up until now, it wasn't possible to access to contents of `store.toml` via libcnb. This PR adds a minimal implementation and tests for accessing that data. I deliberately limited this PR to just that and not address issues with `store.toml` handling in general:

- Ideally, `store.toml` should be typed like layer metadata is. Working with raw TOML with Rust is tedious, especially when we need a specific TOML type (store.toml metadata needs to be a `toml::value::Table`. However, this is not just a matter of slapping a new associated type on `Buildpack` and use that to parse the metadata. We need a mechanism to handle incompatibilities (like we do with layer metadata). (I filed: #544)
- Right now, if the `BuildResult` does not contain `store.toml` data (i.e. that field is `None`), the file on disk is not modified. This is inconsistent with i.e. layers where the metadata needs to be written every time to keep it around. There is currently no way to remove all metadata either since `None` signals "don't change" instead of "no data". (I filed: #545)

Without the fixes for the issues above, this PR a non breaking change we can roll out immediately as a point release to provide the needed functionality to buildpacks that need it and tackle the harder breaking changes separately. 

Closes GUS-W-12299144

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ separate changelogs for each crate were used. If you need to refer to these old 
 
 ## [Unreleased]
 
+### Added
+
+- libcnb: Add `store` field to `BuildContext`, exposing the contents of `store.toml` if present. ([#543](https://github.com/heroku/libcnb.rs/pull/543))
+
 ## [0.11.2] 2022-12-15
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


edmorley (approved):
Thank you!

PR #547, opened 2023-01-10
Update Store derives

schneems:
## Store cannot be cloned 

I want to have write some code like this:

```rust
        let store = context.store.get_or_insert_with(|| {
            let mut rng = rand::thread_rng();
            let value = (0..64)
                .map(|_| rng.sample(rand::distributions::Alphanumeric) as char)
                .collect::<String>();

            let mut metadata = toml::value::Table::default();
            metadata.insert(String::from("SECRET_KEY_BASE"), value.into());

            Store { metadata }
        });
```

But I can't since the store isn't mutable and I cannot clone it:

```

   Compiling heroku-ruby-buildpack v0.0.0 (/Users/rschneeman/Documents/projects/work/buildpacks/buildpacks-ruby/buildpacks/ruby)
warning: unused variable: `store`
   --> buildpacks/ruby/src/main.rs:117:13
    |
117 |         let store = context.store.get_or_insert_with(|| {
    |             ^^^^^ help: if this is intentional, prefix it with an underscore: `_store`
    |
    = note: `#[warn(unused_variables)]` on by default

error[E0596]: cannot borrow `context.store` as mutable, as `context` is not declared as mutable
   --> buildpacks/ruby/src/main.rs:117:21
    |
65  |       fn build(&self, context: BuildContext<Self>) -> libcnb::Result<BuildResult, Self::Error> {
    |                       ------- help: consider changing this to be mutable: `mut context`
...
117 |           let store = context.store.get_or_insert_with(|| {
    |  _____________________^
118 | |             let mut rng = rand::thread_rng();
119 | |             let value = (0..64)
120 | |                 .map(|_| rng.sample(rand::distributions::Alphanumeric) as char)
...   |
126 | |             Store { metadata }
127 | |         });
    | |__________^ cannot borrow as mutable

For more information about this error, try `rustc --explain E0596`.
warning: `heroku-ruby-buildpack` (bin "heroku-ruby-buildpack") generated 1 warning (1 duplicate)
error: could not compile `heroku-ruby-buildpack` due to previous error; 1 warning emitted
warning: build failed, waiting for other jobs to finish...
warning: `heroku-ruby-buildpack` (bin "heroku-ruby-buildpack" test) generated 1 warning
error: could not compile `heroku-ruby-buildpack` due to previous error; 1 warning emitted
[Finished running. Exit status: 101]
```

I'm able to extract the metadata and work on that directly, but I would rather maintain and manipulate Store directly.

```rust
        let mut metadata = match context.store {
            Some(store) => store.metadata,
            None => toml::value::Table::default(),
        }
        .clone();

        metadata.entry("SECRET_KEY_BASE").or_insert_with(|| {
            let mut rng = rand::thread_rng();

            (0..64)
                .map(|_| rng.sample(rand::distributions::Alphanumeric) as char)
                .collect::<String>()
                .into()
        });
```

## Store has no default

Not a big deal if it's not there, but Option has a `get_or_default` in nightly and it would be neat to be able to `context.store.get_or_default` when that is stable.


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ separate changelogs for each crate were used. If you need to refer to these old
 
 ## [Unreleased]
 
+### Fixed
+
+### Added
+
+- libcnb: Store struct now supports `clone()` and `default()`. ([#547](https://github.com/heroku/libcnb.rs/pull/547))
```
Malax, line 13: Technically, this is `libcnb-data`. I'm unsure what's best to write here since most users (I assume) will not use `libcnb-data` directly and might not even be aware that it exists. So far, we don't have `libcnb-data` prefixed changelog entries and prefixes seem to have started just a couple of versions ago.

I'm fine keeping this as-is, this is as minor as it gets and there is not a clear best way to handle this. Just bringing this up in case someone from @heroku/languages has a better idea. :)

Malax (approved):

CHANGELOG.md
edmorley: I think perhaps we should use `libcnb-data: ` for these, since it's more precise. For example, if we were to make a change to `libcnb-package`, I don't think we would want to describe it as two entries, one for `libcnb-test` and one `libcnb-cargo`, but a single `libcnb-package` entry. I think the changelog being unified should prevent end users from being confused about whether they are affected or not, which was the reason we had the cross-linking thing before.

Malax:
> Option has a get_or_default in nightly

@schneems FWIW, there is also `unwrap_or_default` available today. :)

CHANGELOG.md
schneems: Updated the changelog. I can go either way. 

FWIW I basically didn't even notice that's what was happening with `libcnb:` I just copied the next line down and edited it. With a question of tags or no tags. I'm not sure what we want to get out of them. I prefer to keep my changelog format lightweight (I even prefer to not make my links to PRs proper markdown links so it's one less thing to change/edit and one less thing to mess up and and have a weird git blame). Though you all know me and my tendency to make mistakes on small stuff like that.

I would say the value of the tag is in letting people filter for only libraries they're using. In that case I would say that `libherokubuildpack` seems to be the only odd one out, though there could be more. I would expect anyone else to be using all of libcnb. 

Merge is blocked on resolution of this comment thread

schneems:
Thanks for the merge and release. Looks much better https://github.com/heroku/buildpacks-ruby/pull/93/commits/78a45b8b85405e930dd5a46f841e991dbfaf6d7f. (Also ready for review)

PR #549, opened 2023-01-12
Introduce a release script

schneems:
## What

Add a script for preparing the library for a release. I've got something similar (though less solid) on my preparing my legacy buildpack for a release: https://github.com/heroku/heroku-buildpack-ruby/blob/a1dc09a93fc81a73482d6db0ecb02298001d5af4/Rakefile#L70-L104. 

While this isn't a buildpack, the preparation and updating are similar experiences.

## Why

Releasing is a chore. Automation improves quality and reduces toil.

Worst case, this is a throwaway experiment:  I want to explore automating our processes using Rust to serve as an opportunity for me to learn and to inform future efforts in scripting and automation. 

Best case, it helps us move towards more automation: For example, we could move this logic into a GitHub action. That would allow one fewer review step for the release process. 

## Demo

```
$ cargo run -p prepare
    Finished dev [unoptimized + debuginfo] target(s) in 0.37s
     Running `target/debug/prepare`
# Preparing libcnb.rs for release

 - Found project root at /Users/rschneeman/Documents/projects/work/libcnb.rs

## Detecting semver from CHANGELOG.md

- Semver change is: Major

## Updating Cargo.toml /Users/rschneeman/Documents/projects/work/libcnb.rs/Cargo.toml

- Last release was version: '0.11.3'
- Updated version: '1.0.0'

## Updating CHANGELOG.md /Users/rschneeman/Documents/projects/work/libcnb.rs/CHANGELOG.md

- Done
```

### Detect SemVer value from CHANGELOG.md

The script parses the CHANGELOG.md as an AST and knows if the next feature should be a major/minor/patch bump based on the presence/absence of Changed/Added/Fixed sections.

### Update the Cargo.toml

Once we know the semver change, the script can read in the old version from `Cargo.toml`, apply the SemVer update to that version and write the new version back to `Cargo.toml`. This process uses `toml_edit` which preserves formatting including whitespace, ordering, and comments.

### Updating the CHANGELOG.md

Finally the changelog needs to move the elements under `## [Unreleased]` to a new section with the version and date. The version was determined previously and the date is grabbed via `crono`.

## Draft comments

I'm looking for feedback on the overall idea as well as implementation-specific bits. The intent here is to be explorative so, go wild. That being said, I would like to actually ship some automation, so while it will be valuable to think about the best possible ways forward, I'm also looking for guidance on where "good enough" lies. 

Ultimately I want scripting and automation to be high-quality and robust (both to the user and maintainer), but I'm mindful that it should be pragmatic as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ members = [
     "examples/basics",
     "examples/execd",
     "examples/ruby-sample",
+    "scripts/prepare",
     "libcnb",
     "libcnb-cargo",
     "libcnb-data",
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -6,19 +6,19 @@ easier to gauge cross-crate compatibility.
 ## Prepare Release
 
 1. Create a new branch for the upcoming release
-2. Update [Cargo.toml](./Cargo.toml) in the root of the repository:
-   1. In the `workspace.package` table, update `version` to the new version
-   2. In the `workspace.dependencies` table, update the `version` of each of the repository-local dependencies to the new version
-3. Update [CHANGELOG.md](./CHANGELOG.md)
-   1. Move all content under `## [Unreleased]` to a new section that follows this pattern: `## [VERSION] YYYY-MM-DD`
-   2. If appropriate, add a high-level summary of changes at the beginning of the new section
-4. Commit the changes, push them and open a PR targeting `main`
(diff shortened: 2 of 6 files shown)


Malax:

edmorley:
Btw I found https://github.com/crate-ci/cargo-release which would avoid the need to write our own custom tooling.

PR #567, opened 2023-04-28
Remove Process::direct, make Process::command a list

Malax:


README.md
```diff
@@ -79,7 +79,7 @@ Since we're writing a Cloud Native Buildpack, we also need a `buildpack.toml`. U
 file named `buildpack.toml` in the root of your project, right next to `Cargo.toml`.
 
 ```toml
-api = "0.8"
+api = "0.9"
```
edmorley, line 82: There are a few more instances of `api = "0.8"` in `libcnb-data/src/buildpack/mod.rs` (rustdocs + tests) that I think we should update to `api = "0.9"` too.

libcnb-data/src/launch.rs
```diff
@@ -184,29 +182,27 @@ pub struct ProcessBuilder {
 impl ProcessBuilder {
     /// Constructs a new `ProcessBuilder` with the following defaults:
```
edmorley, line 183: (Optional) I wonder if we need a quick mention in this rustdocs about why `command` is a vec, and how it differs from the args? As in, to avoid the scenario where a new user of libcnb doesn't see `.arg()` / `.args()` and (quite reasonably) presumes that `command` being an array means all args should just be added there?

CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ separate changelogs for each crate were used. If you need to refer to these old
 
 ### Changed
 
+- libcnb.rs now targets [Buildpack API 0.9](https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.9). Buildpacks need to upgrade the `api` key to `0.9` in their `buildpack.toml`. ([#567](https://github.com/heroku/libcnb.rs/pull/567))
+  - `Process` no longer supports the `direct` flag. All processes are now `direct`. ([#567](https://github.com/heroku/libcnb.rs/pull/567))
```
edmorley, line 12: I wonder if it might help people migrating, if we mentioned something like ~"Processes that need to use bash should now wrap any commands with it explicitly"

edmorley (approved):
Thank you!

README.md
Malax: Thanks! Fixed via d7e26ee.

CHANGELOG.md
Malax: Nice idea, added something in 2f4ffb5.

libcnb-data/src/launch.rs
Malax: I have a hard time to come up with something here. I agree that `args` is now confusing but that's on the upstream spec. It's not super well explained there either. I will mark this a resolved and merge, we can add something in a separate PR.

Issue #484, opened 2022-07-22
Add a cargo command for performing both packaging and a `pack build`


edmorley:
When developing a buildpack there are a few ways to iterate, that all have their place:
1. Via compiler + Clippy + unit test feedback
2. Via `libcnb-test` powered integration tests
3. Via `cargo libcnb package` followed by `pack build` with the correct arguments

Currently (3) requires quite a bit of boilerplate, for example:

```
cargo libcnb package && pack build --builder heroku/builder:22 --pull-policy if-not-present --path tests/fixtures/empty_procfile/ --buildpack target/buildpack/debug/heroku_procfile procfile-test
```

In addition, this is also error prone, since it's easy to accidentally point at the wrong compiled buildpack directory (for example release instead of debug), or a leftover previous build (as happened recently with @schneems). 

One way to make this workflow easier would be to have a `cargo libcnb run` (or similar) command, that effectively performed both the `cargo libcnb package` followed by the `pack build` with suitable `--buildpack` arguments. For bonus points this command could clean up the old image (or not, depending on args).

edmorley:
As an added bonus, if we were to add a `cargo libcnb run` command, we could then perhaps make `cargo libcnb package` default to `--release` (given the primary use-case of that command would be for production), thereby fixing #359.

PR #580, opened 2023-06-22
Add target triple to buildpack output path

### LINKED ISSUE #579: Multi-Arch Directory Structure
Currently when using `cargo libcnb package`, it outputs the resulting buildpack into `target/buildpack`. This works great in the default (single arch) use case, but if you want to support multiple archs they will clobber each other. What if we put `buildpack` inside an arch directory? For instance:

`target/buildpack/<mode>/<name>` -> `target/buildpack/<arch>/<mode>/<name>`

Another alternative but could just be a good addition would be an `--out-dir` like `cargo build` (though this is unstable atm).

Malax:
When building a buildpack for multiple target triples, the same output directory is currently used by `libcnb-cargo`, clobbering earlier builds for other target triples. This PR changes the output directory to include the target triple, just as cargo itself does.

Fixes #579


libcnb-package/src/lib.rs
```diff
@@ -259,9 +259,11 @@ pub fn get_buildpack_target_dir(
     buildpack_id: &BuildpackId,
     target_dir: &Path,
     is_release: bool,
+    target_triple: &str,
 ) -> PathBuf {
     target_dir
         .join("buildpack")
```
edmorley, line 265: I was wondering if we should revisit the `target/buildpack/...` prefix, given that:
(a) we're making a breaking change to the path already
(b) the GitHub Actions Rust cache action caches the whole `target/` directory (minus a few things it cleans up), when we don't need/really want the `target/buildpack/...` directory being cached
(c) it's more time consuming/error-prone to tab-complete a path under `target/...` (vs an otherwise empty directory) when wanting to run a buildpack locally and hand-typing `pack build --builpack target/buildpack/...` - since `target/` contains many other files/directories
(d) the path is now getting quite deep

eg: Using something like:
`dist/x86_64-unknown-linux-musl/debug/some-org_with-buildpack`

...instead of:
`target/buildpack/x86_64-unknown-linux-musl/debug/some-org_with-buildpack`

If not, then that's ok too. We could fix (b) via adding manual `target/` cleanup steps to our CI configs, and could improve the UX for (c) by implementing #484 instead.

hone: This makes sense to me. I think `target/` was only picked to just have a single output directory. I think not mixing it with the rust output directory makes sense, but it does mean that `dist` becomes part of the interface which I think is fine.

Malax: Using a separate directory would also mean adding configuration options for said directory since it no longer can be configured by the standard cargo config options. Implicitly writing `dist` to the same (parent) directory `target` is in doesn't work reliably since that directory might be, for example, read-only.

@edmorley would you be fine approving this PR in the current state and I'll add another PR that moves the output to a separate directory that defaults to `$WORKSPACE_ROOT/dist/`? This way, we keep the changes grouped together and if we merge both before release there would be only one breaking change event for users.

edmorley: Yeah that sounds great, thank you :-)

edmorley: Also, feel free to switch to something other than `dist/` if you can think of a default better name.

PR #583, opened 2023-06-29
Move packaged buildpack directory out of `target/`

Malax:
`cargo libcnb package` currently puts the packaged buildpacks in Cargo's `target` directory. This requires a deep path to avoid clashing with Cargo itself. Modifiying the `target` directory might also interfere with other Rust tooling that doesn't expect non-Cargo files in that directory. One case is caching via https://github.com/Swatinem/rust-cache.

This PR moves the packaged buildpacks into a dedicated directory. The default name and location for that directory is `$WORKSPACE_ROOT/libcnb-packaged`. This can be changed via the new `--package-dir` option of `cargo libcnb package`.

See: https://github.com/heroku/libcnb.rs/pull/580#discussion_r1238406927

Closes GUS-W-13813150

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 /target/
-/libcnb-packaged/
+/packaged/
 .DS_Store
 .idea
 Cargo.lock
 **/fixtures/*/target/
-**/fixtures/*/libcnb-packaged/
+**/fixtures/*/packaged/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 /target/
+/libcnb-packaged/
 .DS_Store
 .idea
 Cargo.lock
 **/fixtures/*/target/
+**/fixtures/*/libcnb-packaged/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@ separate changelogs for each crate were used. If you need to refer to these old 
   - `get_buildpack_target_dir` was renamed to `get_buildpack_package_dir` ([#583](https://github.com/heroku/libcnb.rs/pull/583))
 - `libherokubuildpack`: Switch the `flate2` decompression backend from `miniz_oxide` to `zlib`. ([#593](https://github.com/heroku/libcnb.rs/pull/593))
 - Bump minimum external dependency versions. ([#587](https://github.com/heroku/libcnb.rs/pull/587))
-- `libcnb-cargo`: Default location for packaged buildpacks moved from Cargo's `target` directory to `libcnb-packaged` in the Cargo workspace root. This simplifies the path and stops modification of the `target` directory which previously might have caching implications when other tools didn't expect non-Cargo output in that directory. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#583](https://github.com/heroku/libcnb.rs/pull/583))
+- `libcnb-cargo`: Default location for packaged buildpacks moved from Cargo's `target` directory to `packaged` in the Cargo workspace root. This simplifies the path and stops modification of the `target` directory which previously might have caching implications when other tools didn't expect non-Cargo output in that directory. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#583](https://github.com/heroku/libcnb.rs/pull/583))
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ separate changelogs for each crate were used. If you need to refer to these old 
 - `libcnb-package`: buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
 - `libherokubuildpack`: Switch the `flate2` decompression backend from `miniz_oxide` to `zlib`. ([#593](https://github.com/heroku/libcnb.rs/pull/593))
 - Bump minimum external dependency versions. ([#587](https://github.com/heroku/libcnb.rs/pull/587))
+- `libcnb-cargo`: Default location for packaged buildpacks moved from Cargo's `target` directory to `libcnb-packaged` in the Cargo workspace root. This simplifies the path and stops modification of the `target` directory which previously might have caching implications when other tools didn't expect non-Cargo output in that directory. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#583](https://github.com/heroku/libcnb.rs/pull/583))
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,9 @@ separate changelogs for each crate were used. If you need to refer to these old 
   - `TestRunner::new` has been removed, since its only purpose was for advanced configuration that's no longer applicable. Use `TestRunner::default` instead. ([#620](https://github.com/heroku/libcnb.rs/pull/620))
   - `LogOutput` no longer exposes `stdout_raw` and `stderr_raw`. ([#607](https://github.com/heroku/libcnb.rs/pull/607))
   - Improved wording of panic error messages. ([#619](https://github.com/heroku/libcnb.rs/pull/619) and [#620](https://github.com/heroku/libcnb.rs/pull/620))
-- `libcnb-package`: buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
+- `libcnb-package`:
+  - buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
+  - `get_buildpack_target_dir` was renamed to `get_buildpack_package_dir` ([#583](https://github.com/heroku/libcnb.rs/pull/583))
 - `libherokubuildpack`: Switch the `flate2` decompression backend from `miniz_oxide` to `zlib`. ([#593](https://github.com/heroku/libcnb.rs/pull/593))
 - Bump minimum external dependency versions. ([#587](https://github.com/heroku/libcnb.rs/pull/587))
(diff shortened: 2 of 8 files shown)


libcnb-cargo/src/package/command.rs
```diff
@@ -25,21 +25,30 @@ pub(crate) fn execute(args: &PackageArgs) -> Result<()> {
...
+
+    std::fs::create_dir_all(&package_dir)
+        .map_err(|e| Error::CreatePackageDirectory(package_dir.clone(), e))?;
+
     let buildpack_packages = create_dependency_graph(
-        find_buildpack_dirs(&workspace, &[workspace_target_dir.clone()])
+        find_buildpack_dirs(&workspace_root_path, &[package_dir.clone()])
             .map_err(|e| Error::FindBuildpackDirs {
-                path: workspace_target_dir.clone(),
+                path: workspace_root_path,
```
Malax, line 51: This was a bug, the error message referred to this `path` as the location, when it really was the ignored directory.

libcnb-cargo/src/cli.rs
```diff
@@ -25,6 +26,9 @@ pub(crate) struct PackageArgs {
     /// Build for the target triple
     #[arg(long, default_value = "x86_64-unknown-linux-musl")]
     pub target: String,
+    /// Directory for packaged buildpacks, defaults to 'libcnb-packaged' in Cargo workspace root
+    #[arg(long)]
+    pub package_dir: Option<PathBuf>,
```
Malax, line 31: I'm not 💯 happy with the way the default is handled here. But since the actual path depends on the cargo metadata, we cannot determine it here. Essentially, the default path is dynamic. From my research, this is the best we can do with `clap` in such cases. Happy to get corrected :)

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ separate changelogs for each crate were used. If you need to refer to these old
 - `libcnb-package`: buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
 - `libherokubuildpack`: Switch the `flate2` decompression backend from `miniz_oxide` to `zlib`. ([#593](https://github.com/heroku/libcnb.rs/pull/593))
 - Bump minimum external dependency versions. ([#587](https://github.com/heroku/libcnb.rs/pull/587))
+- `libcnb-cargo`: Default location for packaged buildpacks moved from Cargo's `target` directory to `libcnb-packaged` in the Cargo workspace root. This simplifies the path and stops modification of the `target` directory which previously might have caching implications when other tools didn't expect non-Cargo output in that directory. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#583](https://github.com/heroku/libcnb.rs/pull/583))
```
edmorley, line 28: There's a `libcnb-package` changelog entry a couple of lines above this one that perhaps needs to be combined with this new entry? (Since the path has been changed twice)

libcnb-cargo/src/cli.rs
colincasey: would using custom [value parser](https://docs.rs/clap/latest/clap/builder/struct.ValueParser.html) be possible?

colincasey:
@Malax there is some overlap with #632 here. can we resolve this PR after that one or would you prefer this one first?

Malax:
@colincasey I understand. The thing is that we want to cut a release and get this breaking change in since the new release will have breaking changes already. If we postpone we have two releases with breaking changes. Happy to help out getting #632 back on track after this PR is merged - is that good enough for you?

libcnb-cargo/src/cli.rs
edmorley, line 31: I think this new `--package-dir` option needs a changelog entry added?

libcnb-package/src/lib.rs
```diff
@@ -255,42 +255,37 @@ pub fn find_buildpack_dirs(start_dir: &Path, ignore: &[PathBuf]) -> std::io::Res
 
 /// Provides a standard path to use for storing a compiled buildpack's artifacts.
 #[must_use]
-pub fn get_buildpack_target_dir(
+pub fn get_buildpack_package_dir(
```
edmorley, line 258: `get_buildpack_target_dir` was public so this needs a changelog entry

libcnb-cargo/src/cli.rs
Malax: From what I've seen, the logic around mandatory arguments and defaults is implemented outside of `ValueParser`s. Even if we could handle optional values in `ValueParsers`, we cannot specify the default before parsing so that it will show up when using `--help` (which is my main concern).

colincasey:
maybe we should also prioritize #631, #632, #633 now then?  all those changes to support composite buildpack testing will mean we'll still have two releases with breaking changes, right?

libcnb-cargo/tests/test.rs
```diff
@@ -330,5 +329,5 @@ fn copy_dir_recursively(source: &Path, destination: &Path) -> std::io::Result<()
 }
 
 const X86_64_UNKNOWN_LINUX_MUSL: &str = "x86_64-unknown-linux-musl";
-const TARGET_DIR_NAME: &str = "target";
 const CARGO_LIBCNB_BINARY_UNDER_TEST: &str = env!("CARGO_BIN_EXE_cargo-libcnb");
+const DEFAULT_PACKAGE_DIR_NAME: &str = "libcnb-packaged";
```
edmorley, line 333: I think it would be worth adding `libcnb-packaged/` to this repo's `.gitignore`.

edmorley:
> all those changes to support composite buildpack testing will mean we'll still have two releases with breaking changes, right?

Manuel was referring to breaking changes to the output directory of `cargo libcnb package`.

There is already one breaking change to that path (as of #580), which will mean needing the adjust the hardcoded paths in:
https://github.com/heroku/languages-github-actions

If we release libcnb 0.14.0 without this PR, then once libcnb 0.15.0 is released, we'll have to update the hardcoded paths in languages-github-actions a second time.

libcnb-cargo/tests/test.rs
Malax: Good call, [268a193](https://github.com/heroku/libcnb.rs/pull/583/commits/268a193f96861684cf23c4af9ebbc1f0d9507fa3)

.gitignore
```diff
@@ -1,5 +1,7 @@
 /target/
+/libcnb-packaged/
 .DS_Store
 .idea
 Cargo.lock
 **/fixtures/*/target/
+**/fixtures/*/libcnb-packaged/
```
edmorley, line 7: Does the packaging happen directly in this repo, or inside the tempdir into which the fixtures are copied? If the latter, then it seems that `**/fixtures/*/libcnb-packaged/` doesn't need to be here?

CHANGELOG.md
Malax: This is a weird one. The target triple in the output path is controlled by `libcnb-package`. The new output directory root however is a `libcnb-cargo` change only. `libcnb-package` always got an explicit root path and created the distribution under that one.

So technically, these are two separate changes. Is that easy to understand for users `cargo libcnb package`? 💯  not. I'm not really sure which solution I like the best here. In the majority of cases, `libcnb-package` is only used as an internal dependency so that changes to it could be considered changes of `cargo-libcnb`.

We have a similar situation with `libcnb-data` where it's not always clear where the changelog should go.

libcnb-cargo/src/cli.rs
Malax: Added in [292155e](https://github.com/heroku/libcnb.rs/pull/583/commits/292155edaf930f754e40429c4716bbdaffdf840a)

libcnb-package/src/lib.rs
Malax: Added in [a3690ba](https://github.com/heroku/libcnb.rs/pull/583/commits/a3690ba54199288f6bc4f3ec44a96ec0f7e528ce)

libcnb-cargo/tests/test.rs
edmorley, line 333: So one downside to the default name of `libcnb-packaged` is that when `cargo libcnb package` is run in the libcnb repo (eg when testing), one ends up with an output directory that's right next to the very similarly named directory for `libcnb-package`:

Screenshot of repo directories

In addition, I think the name `libcnb-packaged` is also possibly a bit redundant, given that in general we'll only ever have repos that are libcnb.rs-only (once the temporary transition from bash to libcnb.rs for our CNBs is over), so the directory name perhaps doesn't need to include the words "libcnb"?

Some alternatives that spring to mind:

```
build/
dist/
output/
packaged/
```

I'd probably lean away from `build/` since:
(a) it would never tab complete cleanly due to buildpack repo roots sometimes containing the file `buildpack.toml` (depending on repo layout),
(b) conceptually the name overlaps too much with compiling etc, so it might not be clear how it differs from the output of the compiler etc.

(Whilst the same issue about tab completion could be said for `packaged/` and `package.toml`, I think we should just delete the `package.toml` of file in the repo roots that have them, since they are currently rewritten to be correct during packging anyway.)

As such, I think my favourites would be `dist/` or `packaged/`.

libcnb-cargo/src/package/command.rs
```diff
@@ -25,21 +25,30 @@ pub(crate) fn execute(args: &PackageArgs) -> Result<()> {
...
-            path: workspace.clone(),
+            path: workspace_root_path.clone(),
             source: e,
         })?;
 
+    let package_dir = args.package_dir.clone().unwrap_or_else(|| {
+        cargo_metadata
+            .workspace_root
+            .into_std_path_buf()
+            .join("libcnb-packaged")
```
edmorley, line 42: Ah we probably need to update the example `cargo libcnb package` log output here at some point:
https://github.com/heroku/libcnb.rs/blob/e3a3da8fecfd08b4815a782ba2dc449d71e8f6e9/README.md?plain=1#L188-L190
https://github.com/heroku/libcnb.rs/blob/e3a3da8fecfd08b4815a782ba2dc449d71e8f6e9/libcnb-cargo/README.md?plain=1#L26-L28

Malax: We should do that in this PR I think. I'm on it!

libcnb-cargo/tests/test.rs
Malax: I have zero passion about the name of the directory. Happy to change. :) I changed it to `packaged` in [2919416](https://github.com/heroku/libcnb.rs/pull/583/commits/29194163b2c0239f611ca8fc89a85777bdb1d2c1).

edmorley (approved):

PR #588, opened 2023-07-05
Clean up `TestContext` when pack build unexpectedly succeeds

### LINKED ISSUE #586: libcnb-test: Images not cleaned up when `pack build` unexpectedly suceeeds
When libcnb-test invokes `pack build`, it passes an explicit Docker image name that is then `docker rmi`'d once the associated `TestContext` struct is dropped:
https://github.com/heroku/libcnb.rs/blob/d9f0bc11993ace2da8f867d8862deccae6ce02c6/libcnb-test/src/test_context.rs#L248-L264

However, there is one scenario where this clean-up is not taking place: When `.expected_pack_result(PackResult::Failure)` has been used, and then the `pack build` unexpectedly succeeds. 

For example, during this the `unexpected_pack_success` testcase:
https://github.com/heroku/libcnb.rs/blob/d9f0bc11993ace2da8f867d8862deccae6ce02c6/libcnb-test/tests/integration_test.rs#L161-L174

In this scenario, the panic occurs in `run_pack_command` before the `TestContext` has been created:
https://github.com/heroku/libcnb.rs/blob/d9f0bc11993ace2da8f867d8862deccae6ce02c6/libcnb-test/src/pack.rs#L179-L214

This means there is no `TestContext` to be dropped, so no remove image command is invoked.

This causes a leak of one image each time the `libcnb.rs` repo's integration tests are run, eg:

```
$ docker images
REPOSITORY                                  TAG       IMAGE ID       CREATED        SIZE
...
libcnbtest_bwgvfgtzszywyveaqjfnhhknfycclp   latest    d35f71755225   43 years ago   650MB
```

GUS-W-13903181.

colincasey:
Moves the creation of the `TestContext` to before the invocation of `run_pack_command` so that the context will be cleaned up regardless of the outcome.

@fixes #586

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ separate changelogs for each crate were used. If you need to refer to these old 
 
 ### Changed
 
+- `libcnb-test`: ensures that an image created during a test run will be cleaned up when `.expected_pack_result(PackResult::Failure)` is used but the result is  `PackResult::Success`. ([#588](https://github.com/heroku/libcnb.rs/pull/588))
 - `libcnb-package`: buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo 
 libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
 - Bump minimum external dependency versions. ([#587](https://github.com/heroku/libcnb.rs/pull/587))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ separate changelogs for each crate were used. If you need to refer to these old 
 
 ### Changed
 
-- `libcnb-test`: ensures that an image created during a test run will be cleaned up when `.expected_pack_result(PackResult::Failure)` is used but the result is  `PackResult::Success`. ([#588](https://github.com/heroku/libcnb.rs/pull/588))
+- `libcnb-test`: ensures that an image created during a test run will be cleaned up when `.expected_pack_result(PackResult::Failure)` is used but the result is `PackResult::Success`. ([#588](https://github.com/heroku/libcnb.rs/pull/588))
 - `libcnb-package`: buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo 
 libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
 - Bump minimum external dependency versions. ([#587](https://github.com/heroku/libcnb.rs/pull/587))
--- a/libcnb-test/src/test_runner.rs
+++ b/libcnb-test/src/test_runner.rs
@@ -156,15 +156,18 @@ impl TestRunner {
             };
         }
 
-        let output = run_pack_command(pack_command, &config.expected_pack_result);
-
-        let test_context = TestContext {
-            pack_stdout: String::from_utf8_lossy(&output.stdout).into_owned(),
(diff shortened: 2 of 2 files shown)


schneems (changes requested):
Your PR looks good in terms of I think it does what it says it does. I'm wanting to make it so that it doesn't regress in the future i.e. if someone were to re-order the commands in a future PR tests wouldn't fail.

That's the "changes requested": To encode the fix in the build/test logic (through tests, types, or both).

## We could (brainstorm):

- Add a test for this explicit behavior (probably hard and/or  expensive)
- Couple `run_pack_command` with needing a `TestContext` (even if it's not used) to force it's creation before it's run or possibly introduce another function
- Move the creation of the TestContext to the same location that allocates the resource (RAII)
- Couple the creation of the image name with the creation of something that tears it down.

Ideally I want to move the logic closest to where the resource is allocated. In this case we decouple creating an image name from allocating/generating that image (via pack). I'm proposing to couple the teardown logic of the image name along with image name creation so then there's never a scenario where one would be created and not torn down.

## Sketch

- Introduce a `ImageName` struct that holds, well, an image name.
- Have `util::random_docker_identifier()` return this new struct instead of a string
- Move the drop logic to this struct `ImageName` struct
- Add a new field to TestContext to hold an ImageName

Now when `TestContext` is dropped it will drop it's contents and call drop on `ImageName` triggering the cleanup (existing behavior is preserved). And since `ImageName` is allocated before `run_pack_command` is ever called, if it fails then `ImageName` drop behavior will be called (new behavior). 

This assumes that the drop logic is a no-op if an image name has not been created in docker.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ separate changelogs for each crate were used. If you need to refer to these old
 
 ### Changed
 
+- `libcnb-test`: ensures that an image created during a test run will be cleaned up when `.expected_pack_result(PackResult::Failure)` is used but the result is  `PackResult::Success`. ([#588](https://github.com/heroku/libcnb.rs/pull/588))
```
Malax, line 15:
```suggestion
- `libcnb-test`: ensures that an image created during a test run will be cleaned up when `.expected_pack_result(PackResult::Failure)` is used but the result is `PackResult::Success`. ([#588](https://github.com/heroku/libcnb.rs/pull/588))
```

Malax:
@edmorley does #619 also resolve this issue? In any case, this PR here won't apply cleanly anymore after `run_pack_command` changed so much.:/

edmorley:
@Malax No #619 does not fix this directly, though it does give more options here as to how to fix this issue.

It seems the options here now are to:
1. (This is now only possible after #619) Make an explicit call to remove the Docker image on the "successful but panicing" branch here: https://github.com/heroku/libcnb.rs/blob/346a186b35a5941a593228b25d68a196c7e75116/libcnb-test/src/test_runner.rs#L170-L171
2. (As Richard suggested above) Add a new `Image` (or similar) struct, which has the `Drop` on it itself, which can be created sooner than the `TestContext` - and will handle the image removal itself, rather than on `TestContext` drop.
3. Move the creation of `TestContext` earlier (the current PR's implementation), but add a comment explaining why.

I think my preference is for option (1) or (2). 

Thoughts?

Malax:
> Thoughts?

I would vote for option 1 here. I find `Drop` complicates the code and makes it hard to reason about. If we can avoid using it I think we should.

edmorley:
> I would vote for option 1 here. I find `Drop` complicates the code and makes it hard to reason about. If we can avoid using it I think we should.

Works for me! I've opened #625.

PR #589, opened 2023-07-07
Initial stubs for docs

schneems:
## Values:

- Make the invisible visible
- Make the implicit explicit

## Goal(s):

- Provide a location for documenting things we think are worth documenting that currently don't have a better location.
- Provide a centralized location for discussing best practices.
- Serve as onboarding docs for future Heroku language members

## Design

Two guides are proposed:

- Owners
- Users

The owners guide is for developers contributing to libcnb (and friends). The users guide is for buildpack developers using libcnb (and friends).

Articles are marked with a status to denote their maturity following the model from the rust stdlib developer guide https://github.com/rust-lang/std-dev-guide/blob/addb3ba3580db84cb327015e817fbeec2713654c/src/getting-started.md.

The structure is merely proposed and not something we're committing to preserving.

## Stakeholders

Eventually I would like others to use libcnb but in the short term "owners" means "Heroku languages team contributing to libcnb" and "users" means "Heroku language team members using libcnb".


edmorley (changes requested):
Thank you for thinking more about the kind of docs we need for libcnb.rs and wider team things!

I like the idea of trying to come up with an overall structure / thinking of the top level concepts we should try and cover with the docs, however:
- I'm not keen on having placeholder stubs for things that don't yet exist (people reading the docs will struggle to find the 10% of completed content amongst the dozens of "TODO" pages etc)
- it's hard for me to understand exactly what's intended for each section when all we have to go by is the title - for some sections my initial reaction is "we don't need this" or "this doesn't belong in this repo" etc
- the mdbook approach presumably means we will need a build step? (vs some simpler flat markdown files with less nesting, that can be rendered be GitHub's UI)
- I'd like to make use of GitHub's native features for contributor onboarding if possible - eg using a file named `CONTRIBUTING.md` which GitHub automatically shows in the UI.

When it comes to adding initial content, my suggestion would be for us to start with a single `CONTRIBUTING.md` file with a handful of headings in it (which would work with GitHub's UI as above). Then over time as we add more content, we can refactor the body of that file out to separate markdown files (with `CONTRIBUTING.md` becoming more of an index). It feels like it will be much easier to figure out how fine-grained the markdown file structure should be once the content actually exists, and that trying to split it up into lots of separate files and sections right now would be a case of premature optimisation?

That said, we clearly do need an iteration/idea generation phase - what do you think about moving the stubs/titles to a handful of GitHub issues (maybe one GitHub issue per rough theme/high level section), where we can discuss section/content ideas first? A GitHub issue would mean we have somewhere ideas can be stored, people can assign themselves as they start working on that docs section etc. Also, GitHub issues can be moved between repos (thanks to GitHub's newish feature), so we can always move any issues to another repo for something we later decide belongs elsewhere.

Malax:
I like the suggestion to grow the documentation organically at first as @edmorley suggested. Empty scaffolding is confusing and the chances of getting filled are quite low given other priorities. If we have a place to put things (such as CONTRIBUTING) we can collect documentation we actually need and re-structure when we need more structure.

Should we close this PR for now then?

schneems:
It turns out that sometimes I need a place to put things before I can make them. I'm going back through our meeting notes and this PR was opened as a result of talking about buildpack output and someone pushing back on it.

Someone made the comment  “some languages do some things differently" and I desired a place to clarify where the team sits on where we can diverge and where we should align.

In recent weeks we've lamented more than a few times that we wish we had somewhere more centralized and official to document decisions. I think this PR was the smallest possible attempt I could think of at doing that. I was imagining that over the months and years we could all collaboratively edit docs as a way of doing that. I was looking for an artifact to align on, but never found it.

PR #590, opened 2023-07-07
Support local buildpacks and meta-buildpacks in libcnb-test

colincasey:
The following changes have made to support local testing of buildpacks and meta-buildpacks:

- refactored shared functionality between `libcnb-cargo` and `libcnb-test` into `libcnb-package`
- modified the test runner to support a `BuildpackReference::Local` variant which is similar to `BuildpackReference::Crate` because both represent buildpacks on the file-system which need to be compiled
- after local and crate references are compiled, they are packaged into Docker images to get around an issue with pack not handling meta-buildpacks correctly (https://github.com/buildpacks/pack/issues/1320)
- all the images created during the test are then cleaned up afterwards

> Note:
> This is still rough around the edges.  A lot of the result error handling in `libcnb-test` is being skipped, documentation is missing, and some tests are still commented out.  Still, I'm putting this up now in case anyone wants to make comments or modifications
>
> If you want to try this out locally in a buildpack project:
> - check out this branch
> - use a path references for the `libcnb-test` dependency
> ```toml
>  libcnb-test = { path = "/Users/ccasey/Projects/heroku/libcnb.rs/libcnb-test" }
> ```
> - use the `BuildpackReference::Local` variant to specify a path to a buildpack or meta-buildpack
> ```rust
> TestRunner::default().build(
>   BuildConfig::new("heroku/builder:22", "/path/to/app")
>     .buildpacks(vec![
>       BuildpackReference::Local(PathBuf::from("../../meta-buildpacks/nodejs"))
>   ]), 
>   |ctx| {
>     // test body
>   }
> )
> ```

### DIFF
--- a/libcnb-package/src/output.rs
+++ b/libcnb-package/src/output.rs
@@ -1,0 +1,243 @@
+use crate::build::BuildpackBinaries;
+use crate::buildpack_dependency::{
+    rewrite_buildpackage_local_dependencies,
+    rewrite_buildpackage_relative_path_dependencies_to_absolute,
+    RewriteBuildpackageLocalDependenciesError,
+    RewriteBuildpackageRelativePathDependenciesToAbsoluteError,
+};
+use crate::CargoProfile;
+use libcnb_data::buildpack::BuildpackId;
+use libcnb_data::buildpackage::Buildpackage;
+use std::fs;
+use std::path::{Path, PathBuf};
+
+pub struct BuildpackOutputDirectoryLocator {
+    root_dir: PathBuf,
+    cargo_profile: CargoProfile,
+    target_triple: String,
+}
+
+impl BuildpackOutputDirectoryLocator {
+    #[must_use]
+    pub fn new(root_dir: PathBuf, cargo_profile: CargoProfile, target_triple: String) -> Self {
+        Self {
+            root_dir,
+            cargo_profile,
+            target_triple,
+        }
+    }
+
+    #[must_use]
+    pub fn get(&self, buildpack_id: &BuildpackId) -> PathBuf {
+        self.root_dir
+            .join("buildpack")
+            .join(&self.target_triple)
+            .join(match self.cargo_profile {
+                CargoProfile::Dev => "debug",
+                CargoProfile::Release => "release",
+            })
+            .join(default_buildpack_directory_name(buildpack_id))
+    }
+}
+
+/// Construct a good default filename for a buildpack directory.
+///
+/// This function ensures the resulting name is valid and does not contain problematic characters
+/// such as `/`.
+#[must_use]
+pub fn default_buildpack_directory_name(buildpack_id: &BuildpackId) -> String {
+    buildpack_id.replace('/', "_")
+}
+
+#[derive(Debug)]
+pub enum AssembleBuildpackDirectoryError {
+    CreateBuildpackDestinationDirectory(PathBuf, std::io::Error),
+    WriteBuildpack(PathBuf, std::io::Error),
+    SerializeBuildpackage(toml::ser::Error),
+    WriteBuildpackage(PathBuf, std::io::Error),
+    CreateBinDirectory(PathBuf, std::io::Error),
+    WriteBuildBinary(PathBuf, std::io::Error),
+    WriteDetectBinary(PathBuf, std::io::Error),
+    CreateAdditionalBinariesDirectory(PathBuf, std::io::Error),
+    WriteAdditionalBinary(PathBuf, std::io::Error),
+    RewriteLocalDependencies(RewriteBuildpackageLocalDependenciesError),
+    RewriteRelativePathDependencies(RewriteBuildpackageRelativePathDependenciesToAbsoluteError),
+}
+
+/// Creates a buildpack directory and copies all buildpack assets to it.
+///
+/// Assembly of the directory follows the constraints set by the libcnb framework. For example,
+/// the buildpack binary is only copied once and symlinks are used to refer to it when the CNB
+/// spec requires different file(name)s.
+///
+/// This function will not validate if the buildpack descriptor at the given path is valid and will
+/// use it as-is.
+///
+/// # Errors
+///
+/// Will return `Err` if the buildpack directory could not be assembled.
+pub fn assemble_single_buildpack_directory(
+    destination_path: impl AsRef<Path>,
+    buildpack_path: impl AsRef<Path>,
+    buildpackage: Option<&Buildpackage>,
+    buildpack_binaries: &BuildpackBinaries,
+) -> Result<(), AssembleBuildpackDirectoryError> {
+    fs::create_dir_all(destination_path.as_ref()).map_err(|e| {
+        AssembleBuildpackDirectoryError::CreateBuildpackDestinationDirectory(
+            destination_path.as_ref().to_path_buf(),
+            e,
+        )
+    })?;
+
+    let destination_path = destination_path.as_ref();
+
+    fs::copy(
+        buildpack_path.as_ref(),
+        destination_path.join("buildpack.toml"),
+    )
+    .map_err(|e| {
+        AssembleBuildpackDirectoryError::WriteBuildpack(destination_path.join("buildpack.toml"), e)
+    })?;
+
+    let default_buildpackage = Buildpackage::default();
+    let buildpackage_content = toml::to_string(buildpackage.unwrap_or(&default_buildpackage))
+        .map_err(AssembleBuildpackDirectoryError::SerializeBuildpackage)?;
+
+    fs::write(destination_path.join("package.toml"), buildpackage_content).map_err(|e| {
+        AssembleBuildpackDirectoryError::WriteBuildpackage(destination_path.join("package.toml"), e)
+    })?;
+
+    let bin_path = destination_path.join("bin");
+    fs::create_dir_all(&bin_path)
+        .map_err(|e| AssembleBuildpackDirectoryError::CreateBinDirectory(bin_path.clone(), e))?;
+
+    fs::copy(
+        &buildpack_binaries.buildpack_target_binary_path,
+        bin_path.join("build"),
+    )
+    .map_err(|e| AssembleBuildpackDirectoryError::WriteBuildBinary(bin_path.join("build"), e))?;
+
+    create_file_symlink("build", bin_path.join("detect")).map_err(|e| {
+        AssembleBuildpackDirectoryError::WriteDetectBinary(bin_path.join("detect"), e)
+    })?;
+
+    if !buildpack_binaries.additional_target_binary_paths.is_empty() {
+        let additional_binaries_dir = destination_path
+            .join(".libcnb-cargo")
+            .join("additional-bin");
+
+        fs::create_dir_all(&additional_binaries_dir).map_err(|e| {
+            AssembleBuildpackDirectoryError::CreateAdditionalBinariesDirectory(
+                additional_binaries_dir.clone(),
+                e,
+            )
+        })?;
+
+        for (binary_target_name, binary_path) in &buildpack_binaries.additional_target_binary_paths
+        {
+            fs::copy(
+                binary_path,
+                additional_binaries_dir.join(binary_target_name),
+            )
+            .map_err(|e| {
+                AssembleBuildpackDirectoryError::WriteAdditionalBinary(
+                    additional_binaries_dir.join(binary_target_name),
+                    e,
+                )
+            })?;
+        }
+    }
+
+    Ok(())
+}
+
+/// TODO
+/// ## Errors
+pub fn assemble_meta_buildpack_directory(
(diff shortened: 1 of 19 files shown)


libcnb-test/src/test_runner.rs
```diff
@@ -101,6 +101,15 @@ impl TestRunner {
     ) {
         let config = config.borrow();
 
+        let mut test_context = TestContext {
```
schneems, line 104: [Question, no action] Does this also close #588 ? My read is that yes, it does. 

I know you talked about wanting to wait on any large refactors in that PR before this was through (as this one is large as is).

libcnb-test/tests/integration_test.rs
```diff
@@ -132,14 +133,29 @@ fn starting_containers() {
...
-#[should_panic(
-    expected = "Could not package current crate as buildpack: BuildBinariesError(ConfigError(NoBinTargetsFound))"
-)]
 fn buildpack_packaging_failure() {
-    TestRunner::default().build(
-        BuildConfig::new("libcnb/invalid-builder", "test-fixtures/empty"),
-        |_| {},
-    );
+    let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let result = std::panic::catch_unwind(|| {
```
schneems, line 138: [Comment, no action] TIL `std::panic::catch_unwind` it seems like this gives us more control in the test. Neat.

schneems (changes requested):
Ed and Manuel: I have some questions and comments related to libcnb-package in here. You can respond out of band or I can open an issue where it would be better to have a discussion. Raising them here since it's top of mind, and one or both of you will need to review before we can merge.

Colin: Thanks again for undertaking this. I also appreciate the walkthrough. I marked some things as change requested, we mostly covered those on the call. We also talked about adding an example usage to the existing `TestRunner` examples to show that it's an option.

I left a bunch of comments, they're marked as such. Writing helps me process. Also I marked other changes as optional.

In addition I also want to mention that we talked about the possibility of a performance optimization. After discussion I don't think we should do it now, but it's worth exploring later. This section is purely for future reference about what we talked about

## Compile once

The current implementation will compile the required local buildpack for every test invocation. We talked about moving to a compile once. This is the model used in Cutlass. To do it I used:

- A class similar to the `Local` variant https://github.com/heroku/cutlass/blob/ea2e67a3f3ccd1a3f14bb3e67cfa37b36537297c/lib/cutlass/local_buildpack.rb
- Teardown/destroy was provided by the test execution framework: https://github.com/heroku/cutlass#initial-config-localbuildpack-for-packagetoml

One challenge of the current implementation is that the buildpacks are compiled and then put into containers. This container packaging is done to support multibuildpacks. If we wanted to optimize to only compile once, then we would end up with a resource that could possibly leak. Also Rust makes sharing across threads/tests significantly harder. 

Originally I proposed using something like `lazy_static` to create and retain a single temp dir where all buildpacks could be compiled and referenced by directory namespace. The OCI part is harder. Also if it's worth optimizing, it's probably worth profiling a bit to gauge how expensive the container creation and stuffing is versus the compilation (my hunch is that it's much quicker, but I would like to verify).

I think it's worth validating the existing logic and implementation and getting this working (I want to use it soon) and then work to refactor for performance later.

libcnb-test/src/test_runner.rs
colincasey: yes and no.  this would close #586. i think refactoring how the test context is tracked/cleaned up (e.g.; attaching docker cleanup logic to an `ImageName` struct) should be addressed in #588 where that discussion is happening.

libcnb-test/src/build_config.rs
```diff
@@ -251,6 +251,8 @@ pub enum BuildpackReference {
     Crate,
     /// References another buildpack by id, local directory or tarball.
     Other(String),
+
+    Local(PathBuf),
```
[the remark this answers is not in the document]
colincasey: @schneems we also discussed that it could be a buildpack id within your project space.

@Malax should we just rename `Other` to `Uri` and do detection in the runner for if it's a local buildpack id or path?  that way we could have it align with how values in `package.toml` are treated during packaging.

libcnb-package/src/output.rs
```diff
@@ -0,0 +1,261 @@
...
+/// It will also rewrite all package.toml references that use the `libcnb:{buildpack_id}` format as
+/// well as relative file references to use absolute paths.
+///
+/// # Errors
+///
+/// Will return `Err` if the meta-buildpack directory could not be assembled.
+pub fn assemble_meta_buildpack_directory(
+    destination_path: impl AsRef<Path>,
+    buildpack_source_dir: impl AsRef<Path>,
+    buildpack_path: impl AsRef<Path>,
```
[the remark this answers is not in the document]
colincasey: * `buildpack_path` is the `buildpack.toml` file 
* `buildpack_source_dir` which is the directory that contains the `buildpack.toml` file.  
 
the `buildpack_source_dir` could be derived from `buildpack_path` though since it's the parent directory. alternatively, the `buildpack_path` could be replaced with `BuildpackData` which might add some clarity to the signature as well.

libcnb-test/src/build_config.rs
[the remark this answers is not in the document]
schneems: > we also discussed that it could be a buildpack id within your project space.

Right. I like the ergonomics of that more, if it's not too difficult to implement. One thought is that it might get confusing to see:

```
Other(String::from("heroku/yarn-engine")), 
CompileLocal(String::from("heroku/nodejs-engine"))
```

Which is related to your question to Manuel.

IDK what he thinks, but my preference is generally to try to separate out refactorings from new features unless they're in conflict. It makes it easier to follow the changes and review, though it's more work for implementers. I'm suggesting we push refactoring/changing `Other()` until after this PR is merged. Also, I'm willing to do some of that work if needed.

To the end of should it be changed. I think the interface is already confusing. It was unclear to me that specifying `heroku/jvm-internal` in the java buildpack test would pull in the one from the builder and not the local one. To that end I think it's worth considering what it does already it handles:

- Path to already compiled buildpack (zipped or not)
- Buildpack Id
- Something else?

Maybe we separate these out into their own variants. For the Buildpack Id case it will work if a buildpack is specified via the builder or via the CLI flags i.e. `--buildpack <path/to/heroku/node>` I'm not sure what to name or call that. Maybe "ExternalBuildpack" or "ExternalBuildpackID"?

schneems:
Linking some in-progress use cases that we can come back to after merge:

- https://github.com/heroku/buildpacks-jvm/pull/545
- https://github.com/heroku/buildpacks-ruby/tree/schneems/metrics-agent-download-rebase

@colincasey I see some activity since our conversation are you ready for a review from Manuel/Ed or still working on updates?

colincasey:
this PR needs to sync up with recent changes to `libcnb-test` but i'm not quite ready for that yet @schneems.  in the meantime, i wouldn't mind if @Malax or @edmorley have time to provide feedback on the refactoring between `libcnb-cargo` and `libcnb-package` that were done to share logic between these three crates.

edmorley:
> * after local and crate references are compiled, they are packaged into Docker images to get around an issue with pack not handling meta-buildpacks correctly (https://github.com/buildpacks/pack/issues/1320)

Thanks to Colin's PR buildpacks/pack#1841, that issue is now fixed in Pack CLI v0.30.0-rc1:
https://github.com/buildpacks/pack/releases/tag/v0.30.0-rc1

As such, I think the workaround should be removed from this PR, since:
(a) the final Pack CLI 0.3.0 release will be out shortly (and we can force the release candidate in CI in the meantime via the `pack-version` option to the `setup-pack` Action)
(b) we don't really need to worry about users with old Pack CLI, since this is a libcnb-test only issue, so only something buildpack maintainers will run into (vs an issue that prevents end-users performing a normal `pack build` with our published buildpacks)

CHANGELOG.md
```diff
@@ -6,6 +6,36 @@ separate changelogs for each crate were used. If you need to refer to these old
...
+### Added
+
+- `libcnb-package`
+  - Added `find_cargo_workspace` which provides a convenient starting point for locating buildpacks for packaging and testing purposes. ([#590](https://github.com/heroku/libcnb.rs/pull/590))
+  - Added the `BuildpackOutputDirectoryLocator` which contains information on how compiled buildpack directories are structured and provides a `.get(buildpack_id)` method which produces the output path for a buildpack. ([#590](https://github.com/heroku/libcnb.rs/pull/590))
+  - Added `output::assemble_single_buildpack_directory` and `output::assemble_meta_buildpack_directory` which construct buildpack output directories with all their required files during packaging. ([#590](https://github.com/heroku/libcnb.rs/pull/590))
+
+### Changed
+
+- `libcnb-package`
```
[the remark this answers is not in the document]
edmorley: > Are we treating `libcnb-package` as a "true" cargo project with version guarantees, or is it more like this internal serde crate

Good question. I'd thought the former, but I'm open to either really, as long as we're clear with potential external consumers of the create. Also at the moment with libcnb being pre-v1, we're bumping the major version most releases, so this mostly won't be an issue - and later once we go v1 I suspect `libcnb-package` will have settled down enough that we don't expect too much breakage.

One thing to watch out for though - if we were to switch to being more relaxed about the major version for `libcnb-package` - we'd really have to switch to using hard version pins instead of ranges for intra-crate dependencies. ie: `libcnb-cargo` would need to depend on `libcnb-package `==1.2.3` not `^1.2.3`. (We should perhaps make that switch regardless given our unified release versions approach.)

libcnb-test/src/pack.rs
```diff
@@ -213,6 +213,64 @@ pub(crate) fn run_pack_command<C: Into<Command>>(
     output
 }
 
+#[derive(Clone, Debug)]
+pub(crate) enum BuildpackPackageFormat {
+    Image,
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct PackBuildpackPackageCommand {
```
edmorley, line 222: This change can be dropped now that the Pack CLI bug has been fixed

libcnb-test/src/pack.rs
```diff
@@ -213,6 +213,64 @@ pub(crate) fn run_pack_command<C: Into<Command>>(
...
+            ),
+        ];
+
+        command.args(args);
+
+        command
+    }
+}
+
+pub(crate) fn run_buildpack_package_command<C: Into<Command>>(command: C) {
```
edmorley, line 262: (there's the new `run_command` util after #619 btw; though moot point given this can be removed due to the Pack CLI bug being fixed)

libcnb-test/src/test_context.rs
```diff
@@ -247,19 +248,23 @@ impl<'a> TestContext<'a> {
 
 impl<'a> Drop for TestContext<'a> {
     fn drop(&mut self) {
-        // We do not care if image removal succeeded or not. Panicking here would result in
-        // SIGILL since this function might be called in a Tokio runtime.
-        let _image_delete_result =
-            self.runner
+        let mut images = vec![&self.image_name];
+        images.extend(&self.buildpack_images);
```
edmorley, line 252: This change can be dropped now that the Pack CLI bug has been fixed

libcnb-test/src/test_runner.rs
edmorley, line 104: This change can be dropped now that the Pack CLI bug has been fixed

libcnb-test/tests/integration_test.rs
```diff
@@ -132,14 +133,29 @@ fn starting_containers() {
 
 #[test]
 #[ignore = "integration test"]
-#[should_panic(
-    expected = "Could not package current crate as buildpack: BuildBinariesError(ConfigError(NoBinTargetsFound))"
```
edmorley, line 136: This test refactoring might be best as a separate PR that lands before this one

edmorley (changes requested):

I've had a quick look through the PR, but am finding it quite hard to review quickly given the size of the PR. 

Now that the Pack CLI bug has been fixed, it seems a fair bit of the complexity/size of this PR might go away, since the solution no longer needs to create an intermediate Docker image for the meta-buildpacks (and keep track of cleaning them up etc).

Another thing that might make the PR easier to review is to split out any changes that can be made separately. For example, unrelated refactorings/changes, or even changes that are related to this work, but are preparatory in nature, and could be made in a prior PR. For example, for the bollard migration work I ended up opening half a dozen PRs, only the last of which was actually the migration itself - which made it much easier for the PR reviewers.

colincasey:
fair point, @edmorley.  i've split out the bulk of the refactoring work into https://github.com/heroku/libcnb.rs/pull/628.

CHANGELOG.md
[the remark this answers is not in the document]
edmorley: @schneems Just to follow up here - we've now (a) switched to hard version pins in #644, (b) marked `libcnb-package` (and the new `libcnb-common`) as internal in their READMEs in #662.

colincasey:
Closing this for #666

PR #606, opened 2023-07-21
Clean up formatting of Rust code in markdown

edmorley:
Sadly by default rustfmt doesn't format Rust code inside markdown in docs, and instead requires that the Nightly-only `format_code_in_doc_comments` feature be enabled:
https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#format_code_in_doc_comments 

We don't want to run Nightly in CI so cannot enable this permanently, however, I ran it as a one-off locally to clean up our docs slightly.

The `.rustfmt.toml` used was:

```
unstable_features = true
format_code_in_doc_comments = true
```

### DIFF
--- a/libcnb-data/src/buildpack/id.rs
+++ b/libcnb-data/src/buildpack/id.rs
@@ -8,8 +8,8 @@ libcnb_newtype!(
     ///
     /// # Examples:
     /// ```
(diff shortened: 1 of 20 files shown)


Malax (approved):
Nice!

PR #607, opened 2023-07-21
libcnb-test: Stop exposing raw output in `LogOutput`

edmorley:
Since:
- There are no common use-cases that require raw output, and buildpacks in the wild aren't using these fields (checked using GitHub code search).
- It saves beginners from having to think about which `LogOutput` field they should be using.

This cleanup has been split out of the future Bollard migration PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,9 @@ separate changelogs for each crate were used. If you need to refer to these old 
 
 ### Changed
 
-- `libcnb-test`: `ContainerContext::address_for_port` now returns `SocketAddr` directly instead of `Option<SocketAddr>`. ([#605](https://github.com/heroku/libcnb.rs/pull/605))
+- `libcnb-test`:
+  - `ContainerContext::address_for_port` now returns `SocketAddr` directly instead of `Option<SocketAddr>`. ([#605](https://github.com/heroku/libcnb.rs/pull/605))
+  - `LogOutput` no longer exposes `stdout_raw` and `stderr_raw`. ([#607](https://github.com/heroku/libcnb.rs/pull/607))
 - `libcnb-package`: buildpack target directory now contains the target triple. Users that implicitly rely on the output directory need to adapt. The output of `cargo libcnb package` will refer to the new locations. ([#580](https://github.com/heroku/libcnb.rs/pull/580))
 - `libherokubuildpack`: Switch the `flate2` decompression backend from `miniz_oxide` to `zlib`. ([#593](https://github.com/heroku/libcnb.rs/pull/593))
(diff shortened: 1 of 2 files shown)


Malax (commented):
I'm on the fence with this one. We're doing a lossy conversion for the other ways to access the output. If it would ever cause problems (even tho it's currently fine), it would be impossible to get the correct data for a workaround and/or debugging. Given that the cost of having the raw fields seems pretty low to me, I'd rather keep them around.

edmorley:
@Malax I'd rather not add (or continue to have) features "just in case", and instead wait until we have an actual use-case, since:
- It's trivial to re-add
- if there is ever a feature request for this/use-case, it might mean there is an even better way to support the use-case than just offering these raw fields, but without someone opening a feature request issue, we'll never know
- there's a cognitive burden to having multiple fields

In general, I suspect that nine times out of ten, when we remove a feature like this (or don't add it in the first place), we'll never need it in the future. If we never have to add extra features, or add features back, then IMO it shows we're constantly over-scoping the design.

Malax:
I'm with you in general wrt to feature creep and not having features "just in case". However, in this case the "feature" is exposing two fields with data we have to acquire regardless of exposing them or not. We're not adding complexity (and therefore maintenance cost) to the implementation by keeping it around. At the same time, if we remove these two fields we remove all ability to get the actual canonical data. If this were just anther convenience helper no-one uses I would remove it too, but this is about the canonical data and for me tips the scale for me towards keeping it.

edmorley:
@Malax And the other points I raised? eg cognitive burden.

edmorley:
> We're not adding complexity (and therefore maintenance cost)

Well in my Bollard rewrite (where the `consume_container_log_output` function gets removed), having these fields meant having to write another two lines of the new implementation. Whilst doing so is trivial, it is still another thing to think about, so it's not correct to say it's zero maintenance cost.

edmorley:
I feel like adding these fields was the right thing to do initially, since we weren't sure whether they'd be needed in the early days of `libcnb-test`. However, we've proven now that nobody actually uses them permanently in their tests - and I've heard of zero instances where they've been used for temporary debugging.

One downside to us being hesitant to removing unused features/APIs/... is that I think it will mean we end up having to be more strict on adding features/APIs in the first place. ie: It's more likely I'll leave "changes required" on a PR in the future, if I think there's no chance of having an API simplified in the future.

Malax:
After discussing this with @edmorley, I am fine with this PR at this point.

I still think having a way to get the canonical, non modified, data is useful to have. However, we have other bits in libcnb that deal with output streams that do not expose the raw output anyways. Ideally, we would have the ability to access it everywhere.

To unblock progress, I will approve this now.

Malax (approved):

PR #608, opened 2023-07-21
libcnb-test: Clean up Pack Command args construction

edmorley:
`Command` already has an `args` accumulator, so there's no need to construct a vector of args prior to passing to `Command`.

This cleanup is ported from the new Docker CLI module added in the upcoming Bollard PR, but has been split out of that PR to keep it smaller.


Malax (approved):
Nice :)

PR #610, opened 2023-07-24
libcnb-test: Only expose ports to localhost

### LINKED ISSUE #609: libcnb-test: `ContainerConfig::expose_port` exposes ports to the internet
`ContainerConfig::expose_port` exposes ports on the started container to the host machine.

Currently libcnb-test does not set a `host_ip` in the corresponding `PortBinding`:
https://github.com/heroku/libcnb.rs/blob/95c2e8164e1382f6da0be72dd99777c12f39f091/libcnb-test/src/container_port_mapping.rs#L82
https://docs.rs/bollard/latest/bollard/models/struct.PortBinding.html

The Bollard docs don't say what this implies, however, the Docker run docs say:

> Note that ports which are not bound to the host (i.e., -p 80:80 instead of -p 127.0.0.1:80:80) are externally accessible.

(on https://docs.docker.com/engine/reference/commandline/run/#publish)

And on the networking page:

> Publishing container ports is insecure by default. Meaning, when you publish a container’s ports it becomes available not only to the Docker host, but to the outside world as well.
> 
> If you include the localhost IP address (127.0.0.1) with the publish flag, only the Docker host can access the published container port.

(on https://docs.docker.com/network/#published-ports)

We should set the `host_ip` to 127.0.0.1, so that ports are only exposed to localhost, and not the internet.

IMO this is another example of how using Bollard makes it harder for us to understand what's going on, since Bollard's docs don't mention this pitfall, and it's not necessarily obvious at first glance how `PortBinding` correlates to the `docker run` options.

edmorley:
Previously ports exposed using `ContainerConfig::expose_port` would be exposed to the internet, rather than just localhost.

Sadly the Bollard docs don't mention this pitfall:
https://docs.rs/bollard/latest/bollard/models/struct.PortBinding.html

See:
https://docs.docker.com/engine/reference/commandline/run/#publish
https://docs.docker.com/network/#published-ports

Fixes #609.
GUS-W-13812521.

### DIFF
--- a/libcnb-test/src/container_port_mapping.rs
+++ b/libcnb-test/src/container_port_mapping.rs
@@ -79,7 +79,7 @@ pub(crate) fn port_mapped_container_config(
                         (
                             format!("{port}/tcp"),
                             Some(vec![PortBinding {
-                                host_ip: None,
+                                host_ip: Some(String::from("127.0.0.1")),
                                 host_port: None,
                             }]),
                         )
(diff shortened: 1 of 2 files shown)


libcnb-test/src/container_port_mapping.rs
```diff
@@ -79,7 +79,7 @@ pub(crate) fn port_mapped_container_config(
                         (
                             format!("{port}/tcp"),
                             Some(vec![PortBinding {
-                                host_ip: None,
+                                host_ip: Some(String::from("127.0.0.1")),
```
Malax, line 82: This will probably cause issues when the user has an IPv6 only system, which is unlikely but not impossible. It's certainly better than exposing to every interface though.

Malax (approved):

PR #621, opened 2023-07-31
libcnb-test: Migrate from Bollard to Docker CLI

### LINKED ISSUE #446: libcnb-test: Exit status of commands run via `run_shell_command` and `shell_exec` are not checked
The various `PrepareContainerContext` methods for starting containers silently ignore non-zero exit statuses of those containers.

This made debugging something recently harder, since I was getting an empty string `stderr` from `logs_now()`, and was thinking the server hadn't had long enough to start. But instead the issue was that the server start had failed, and the content was instead on `stderr`. 

If non-zero container exit status automatically failed the test (dumping both stdout and stderr at the same time), this would have been easier to debug.

GUS-W-11382688.

### LINKED ISSUE #620: libcnb-test: Use Docker CLI instead of Bollard and the Docker daemon API
Currently libcnb-test runs various Docker related actions/commands using the [bollard](https://github.com/fussybeaver/bollard) crate.

Whilst at first glance using a Rust client for Docker seems preferable, it actually has a number of disadvantages:
- Bollard's APIs mostly map to lower-level Docker daemon API primitives, meaning that what might otherwise be a single action using the Docker CLI instead requires the correct combination of multiple daemon API calls and associated edge case/error handling. This means we effectively have to reimplement parts of the Docker CLI, increasing effort of implementing any new feature and the chance of implementation error - which has already resulted in bugs: #446.
- The whole point of libcnb-test is that it's as representative as possible of how end users use CNBs and run the resultant images. End users don't run their containers using direct Docker daemon API calls, they use the Docker CLI. Plus, for maintainers of libcnb-test (or anyone debugging an issue with it), it's much harder to understand the codebase in terms of both Bollard+the Docker daemon API (particularly given their docs aren't great), given that we ourselves are really only used to the Docker CLI.
- Bollard has a very heavy dependency tree (partly due to requiring the tokio runtime) - for example, it increases the number of the Python CNB's transitive dependencies by 50%.

GUS-W-13853580.

edmorley:
Previously libcnb-test ran its various Docker related actions/commands (excluding those handled by Pack CLI) using the [bollard](https://github.com/fussybeaver/bollard) crate.

Whilst at first glance using a Rust client for Docker seems preferable, it turns out that using Bollard/the Docker daemon API actually has a number of disadvantages, which are explained in more detail in #620.

Now:
- The Docker CLI is used instead of Bollard+Tokio+the Docker daemon API, which is simpler to understand/maintain and also avoids ~55 additional transitive dependencies.
- The `run_shell_command` and `shell_exec` commands are now run as attached `docker run` invocations, making it trivial to check their exit code, fixing #446.
- Whenever errors occur, the error output is now easier for end users to understand, since it's presented as Docker CLI output with which they will be more familiar (vs Docker daemon/Bollard API errors).

As part of this change, `ContainerConfig::entrypoint` can no longer accept a vector of strings, since the Docker CLI's `--entrypoint` arg only accepts a single value, unlike the Docker daemon API. However, since the purpose of `libcnb-test` is to replicate typical end-user usage of the buildpacks and resultant images, this actually improves alignment of the framework with the use-cases we want to test.

Fixes #446.
Closes #620.
GUS-W-11382688.
GUS-W-13853580.


joshwlewis (approved):
This changes the nature of the `docker` dependency a bit - bollard was tightly constrained, and the docker daemon API is fairly stable. Now we're loosely dependent on a docker CLI, of which users may have varying versions. That might mean subtle usage changes that are incompatible with this implementation. Should we document the version (or range) we're targeting? Or maybe codify the installation of docker so we catch docker CLI changes when they happen?

Malax (approved):

PR #625, opened 2023-08-08
libcnb-test: Fix image cleanup when a pack build unexpectedly succeeds

### LINKED ISSUE #586: libcnb-test: Images not cleaned up when `pack build` unexpectedly suceeeds

edmorley:
Previously, if a test with an expected result of `PackResult::Failure` unexpectedly succeeded, the built app image was left behind since cleanup was only performed in `TestContext::Drop` (and in this scenario, a `TestContext` is never created in the first place).

Now the app image is correctly cleaned up.

This scenario is hard to write a test for unless we resort to `std::panic::catch_unwind` and manual Docker inspect subprocess calls, which is not really worth it - so I've just tested manually instead.

Fixes #586.
Closes #588.
GUS-W-13903181.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ separate changelogs for each crate were used. If you need to refer to these old 
   - `TestRunner::new` has been removed, since its only purpose was for advanced configuration that's no longer applicable. Use `TestRunner::default` instead. ([#620](https://github.com/heroku/libcnb.rs/pull/620))
   - `LogOutput` no longer exposes `stdout_raw` and `stderr_raw`. ([#607](https://github.com/heroku/libcnb.rs/pull/607))
   - Improved wording of panic error messages. ([#619](https://github.com/heroku/libcnb.rs/pull/619) and [#620](https://github.com/heroku/libcnb.rs/pull/620))
+  - If a test with an expected result of `PackResult::Failure` unexpectedly succeeds, the built app image is now correctly cleaned up. ([#625](https://github.com/heroku/libcnb.rs/pull/625))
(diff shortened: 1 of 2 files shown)


Malax (approved):
Neat impl, agree wrt tests. 👍🏻

PR #628, opened 2023-08-08
Refactor libcnb-package and libcnb-cargo in preparation for test support

colincasey:
The following changes have been made to maximize the sharing of code between `libcnb-cargo`, `libcnb-test`, and `libcnb-package`. This is to prepare for upcoming changes that will allow for testing of local and meta-buildpacks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,11 @@ separate changelogs for each crate were used. If you need to refer to these old 
 
 ### Added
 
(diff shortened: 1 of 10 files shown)


edmorley (commented):
Thank you for splitting out the refactoring from #590 - unfortunately I'm still finding this PR too hard to review as-is, since it's still making a lot of separate changes at once. 

Reading the changelog gives a clue that there are actually quite a few parts of this that can be split out, which we can quickly review/merge individually. To speed things up, you could also stack the PRs (ie: set the base branch to the second PR to that of the first, and similar for each PR after that) for any changes that would otherwise result in merge conflicts etc.

Malax:
Closing in favour of https://github.com/heroku/libcnb.rs/pull/657

Issue #600, opened 2023-07-17
libcnb-cargo: Prevent stdout spam when not packaging for automation


edmorley:
After #575, the `cargo libcnb package` command now outputs the paths of all compiled buildpacks to stdout (with the rest of the output on stderr).

This is intended so that automation can fish out the buildpack names, and is used by eg:
https://github.com/heroku/languages-github-actions

However, it gives a cluttered output for human users of `cargo libcnb package`, eg:

```
$ cargo libcnb package
🔍 Locating buildpacks...
📦 [1/1] Building heroku/python
Determining automatic cross-compile settings...
Building binaries (x86_64-unknown-linux-musl)...
    Finished dev [unoptimized] target(s) in 0.05s
Writing buildpack directory...
Successfully wrote buildpack directory: target/buildpack/x86_64-unknown-linux-musl/debug/heroku_python (7.66 MiB)
✨ Packaging successfully finished!

💡 To test your buildpack locally with pack, run:
pack build my-image-name \
  --buildpack /Users/emorley/src/buildpacks-python/target/buildpack/x86_64-unknown-linux-musl/debug/heroku_python \
  --path /path/to/application

/Users/emorley/src/buildpacks-python/target/buildpack/x86_64-unknown-linux-musl/debug/heroku_python
```

(ie: the very final line of the log output)

Perhaps we need a separate `--machine-readable` or some other output format control?

Malax:
Minor: Ideally we'd use `--message-format` like Cargo itself does. Otherwise I think it's the right direction - you either need output for humans or machines - usually not both.

Issue #599, opened 2023-07-17
libcnb-cargo: Example `pack build` command no longer uses relative paths after #575


edmorley:
When `cargo libcnb package` is run, at the end of the packaging process an example `pack build` command is displayed, with the aim of making it (a) easy for people to understand how to use the compiled buildpack, (b) to save them having to type all of the CLI arguments/paths by hand.

Prior to #575, this example `pack build` command used relative paths to the buildpack, however, it now uses absolute paths, eg:

```
💡 To test your buildpack locally with pack, run:
pack build my-image-name \
  --buildpack /Users/emorley/src/buildpacks-python/target/buildpack/x86_64-unknown-linux-musl/debug/heroku_python \
  --path /path/to/application
```

(ie: the `/Users/emorley/src/buildpacks-python/target/buildpack/x86_64-unknown-linux-musl/debug/heroku_python`)

We should switch back to relative paths to make the shown command shorter and easier to see that the files have actually been saved inside the source dir at a quick glance.

edmorley:
Fixed by #657 (this wasn't auto closed since the PR used the phrase `Fixes: #638, #599`, whereas GitHub requires that "{fixes, closes, ...}" precedes every issue number, rather than just the first in the list)

PR #640, opened 2023-08-18
Update Quick Start Guide

Malax:
As pointed out in https://github.com/heroku/libcnb.rs/pull/583#discussion_r1297831464, the output of `cargo libcnb package` and `pack build` has changed since we last updated the guide. This PR brings the output and output paths up-to-date.

Ref: GUS-W-13972958

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,16 +178,22 @@ In your project directory, run `cargo libcnb package` to start packaging:
 
 ```shell
 $ cargo libcnb package
-INFO - Reading buildpack metadata...
-INFO - Found buildpack libcnb-examples/my-buildpack with version 0.1.0.
-INFO - Determining automatic cross-compile settings...
-INFO - Building binaries (x86_64-unknown-linux-musl)...
+🔍 Locating buildpacks...
+📦 [1/1] Building libcnb-examples/my-buildpack
+Determining automatic cross-compile settings...
+Building binaries (x86_64-unknown-linux-musl)...
 # Omitting compilation output...
-    Finished dev [unoptimized] target(s) in 8.51s
-INFO - Writing buildpack directory...
-INFO - Successfully wrote buildpack directory: target/buildpack/debug/libcnb-examples_my-buildpack (3.03 MiB)
-INFO - Packaging successfully finished!
-INFO - Hint: To test your buildpack locally with pack, run: pack build my-image --buildpack target/buildpack/debug/libcnb-examples_my-buildpack --path /path/to/application
+    Finished dev [unoptimized] target(s) in 8.92s
+Writing buildpack directory...
+Successfully wrote buildpack directory: packaged/x86_64-unknown-linux-musl/debug/libcnb-examples_my-buildpack (4.06 MiB)
+✨ Packaging successfully finished!
+
+💡 To test your buildpack locally with pack, run:
+pack build my-image-name \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -178,16 +178,22 @@ In your project directory, run `cargo libcnb package` to start packaging:
...
+Writing buildpack directory...
+Successfully wrote buildpack directory: packaged/x86_64-unknown-linux-musl/debug/libcnb-examples_my-buildpack (4.06 MiB)
+✨ Packaging successfully finished!
+
+💡 To test your buildpack locally with pack, run:
+pack build my-image-name \
+  --buildpack /home/ponda.baba/my-buildpack/packaged/x86_64-unknown-linux-musl/debug/libcnb-examples_my-buildpack \
+  --path /path/to/application
+
+/home/ponda.baba/my-buildpack/packaged/x86_64-unknown-linux-musl/debug/libcnb-examples_my-buildpack
```
edmorley, line 196: (Fixing these is covered by #599 and #600 btw)

README.md
```diff
@@ -178,16 +178,22 @@ In your project directory, run `cargo libcnb package` to start packaging:
...
+Building binaries (x86_64-unknown-linux-musl)...
 # Omitting compilation output...
-    Finished dev [unoptimized] target(s) in 8.51s
-INFO - Writing buildpack directory...
-INFO - Successfully wrote buildpack directory: target/buildpack/debug/libcnb-examples_my-buildpack (3.03 MiB)
-INFO - Packaging successfully finished!
-INFO - Hint: To test your buildpack locally with pack, run: pack build my-image --buildpack target/buildpack/debug/libcnb-examples_my-buildpack --path /path/to/application
+    Finished dev [unoptimized] target(s) in 8.92s
+Writing buildpack directory...
+Successfully wrote buildpack directory: packaged/x86_64-unknown-linux-musl/debug/libcnb-examples_my-buildpack (4.06 MiB)
```
edmorley, line 188: I'm curious why the buildpack size has increased by 30% since the last time this was refreshed?

edmorley (approved):

README.md
Malax: I saw that too and knew you would latch onto it immediately :) It could really be anything since this is a debug build - rustc might have added more debug symbols or, more likely, we added more dependencies since we wrote the guide. I think I remember writing it before most of us even had CNBs for their languages.

FWIW: I re-ran this with `--release` and the size dropped to 1.78 MiB.

README.md
Malax: Ah nice, I was annoyed when it had my name in the output. OTOH, adding Star Wars references is also great. ;-) We need to make sure to do another update of the guide when #599 and #600 get merged then.

README.md
edmorley: The guide was last refreshed for libcnb 0.11.2, so not that long ago. Agree it might be debug-only inflation, since some new transitive but not really used dep might not get pruned in debug mode. Looking at the diff there were also some `toml` and `fancy-regex` major version upgrades - I know both projects have had significant overhauls (eg toml had a refactor due to wanting to support editing too etc).

edmorley: I saw increases in Procfile CNB release builds too - which I looked into here:
https://github.com/heroku/procfile-cnb/pull/175#issuecomment-1686312268

tl;dr: Mostly regex + toml dependency changes. We should also stop using regex when we don't need to use regex.

PR #652, opened 2023-08-22
First pass refactoring `libcnb-cargo`/`libcnb-package` error handling

### LINKED ISSUE #634: libcnb-cargo Simplify the error mappings
In `libcnb-cargo` all the packaging-related errors are defined on the `package::error::Error` enum which uses `thiserror` annotations to provide user-facing messages.  There are also several `impl From<T> for Error` mappings for internal errors that come from `libcnb-package` to flatten those specific errors into application errors.  This accomplishes the goal of having a specific user message for each possible error but comes with the side-effect that the `package::error::Error` enum is quickly becoming a large, unwieldy list.

A suggestion from @Malax was to provide a new trait that could be implemented for the `libcnb-package` errors that need to be mapped to application errors in `libcnb-cargo`.  E.g.;

```rust
#[derive(Debug, thiserror::Error)]
pub(crate) enum Error {
  //...
  #[error("Error finding workspace:\n{}", .0.to_error_message())]
  FindCargoWorkspace(FindCargoWorkspaceError),
}

trait ApplicationError {
    fn to_error_message(&self) -> String;
}

impl ApplicationError for FindCargoWorkspaceError {
    fn to_error_message(&self) -> String {
        match self {
            FindCargoWorkspaceError::GetCargoEnv(_) => {
                "No environment variable named `CARGO` is set".to_string()
            }
            FindCargoWorkspaceError::SpawnCommand(error) => {
                format!("Failed to spawn Cargo command\nError: {error}")
            }
            FindCargoWorkspaceError::CommandFailure(exit_status) => {
                format!(
                    "Unexpected Cargo exit status while attempting to read workspace root\nExit Status: {}\nExamine Cargo output for details and potential compilation errors.", 
                    exit_code_or_unknown(exit_status)
                )
            }
            FindCargoWorkspaceError::GetParentDirectory(path) => {
                format!(
                    "Could not locate a Cargo workspace within `{}` or it's parent directories",
                    path.display()
                )
            }
        }
    }
}
```

The benefit of this approach is:
* `package::error::Error` would not grow by `N` variants for each `libcnb-package` error
* `libcnb-cargo` still has control over the error message provided each `libcnb-package` error

Malax:
This PR is the first one in a series that cleans up the error handling in `libcnb-cargo` and `libcnb-package`. It is mostly about error structure.

- Removes unecesasry flattening (and duplication) of errors in `libcnb-cargo`
  - Closes #634
- Implement `std::error::Error` for all errors (via `thiserror`) in both crates.
  - Correctly annotate parent errors with thiserror's `#[source]` attribute.
- Rename some errors when the name wasn't good enough to get an idea what went wrong.
- Replaced some implicit error converions with explicit `.map_err()` calls.
- Replaced some `.map_err(std::convert::Into::into)` with explicit mappings.

Issues that this PR doesn't attempt to fix and will be fixed later:
- Error naming (there are many misleading or outright wrong names)
- Improving `Display`/`Debug` implementations for some errors
- Removal of new error values when "bubbling up" would've been good enough.
- Splitting errors that are re-used between (unrelated) functions

Ref: GUS-W-13999008

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `libcnb-package`:
   - Added the `output::create_packaged_buildpack_dir_resolver` helper which contains all the information on how compiled buildpack directories are structured returns a function that can be invoked with `BuildpackId` to produce the output path for a buildpack. ([#632](https://github.com/heroku/libcnb.rs/pull/632))
+  - `std::fmt::Display` and `std::error::Error` implementations for all error values. ([#652](https://github.com/heroku/libcnb.rs/pull/652))
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


edmorley (approved):
Thank you!

PR #656, opened 2023-08-24
Rename `Buildpackage` to `PackageDescriptor`

Malax:
While https://buildpacks.io refers to `package.toml` as a "buildpackage" once, it is usually referred to as `package.toml` which is what we also use in our day-to-day conversations. I think `Buildpackage` is a poor name to use within libcnb for the following reasons:

- It starts with "buildpack", so it turns up in searches for "buildpack".
- We also currently have `BuildpackPackage`, it's very easy to confuse the two.
- The name in general is quite similar to other names we have since it starts with "build".

I wanted to avoid renaming it to `PackageToml`. Even though it might be the first thing that comes to mind as better name, I don't think its a good one. The `Toml` suffix would imply it is TOML data which is not the case for the parsed representation. It can be created via code and never be (de)serialized from or to TOML and the serialised representation is irrelevant when we work with the data in code. We have a similar situation with `buildpack.toml` which we named `BuildpackDescriptor` for similar reasons.

Thus, I went with `PackageDescriptor` to align with `BuildpackDescriptor`. I think readability of the code already improved after this rename.

I have other changes in mind with regard to `PackageDescriptor` that should simplify the code a bunch, but these are out of scope for this PR. This is strictly just a rename.

Ref: GUS-W-14007685

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Renamed `libcnb-data`'s `Buildpackage` to `PackageDescriptor`. This required changes in many other names across multiple libcnb.rs crates for consistency. See the PR diff for details, but renames should be straightforward and unsurprising. ([#656](https://github.com/heroku/libcnb.rs/pull/656))
 - `libcnb-package`:
   - Added `find_cargo_workspace_root_dir` which provides a convenient starting point for locating buildpacks for packaging and testing purposes. ([#629](https://github.com/heroku/libcnb.rs/pull/629))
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Renamed `libcnb-data`'s `Buildpackage` to `PackageDescriptor`. This required changes in many other names across multiple libcnb.rs crates for consistency. See the PR diff for details, but renames should be straightforward and unsurprising. ([#656](https://github.com/heroku/libcnb.rs/pull/656))
```
Malax, line 20: I kept it brief instead of iterating all the changed names. To me, there isn't much value in listing all renames when the renaming schema should be obvious. I can see an argument for listing all, so happy to discuss here.

edmorley (approved):

PR #659, opened 2023-08-28
Add `libcnb-common`, move `toml_file` to it

Malax:
While working on #657, I wanted to use some internal helpers that are available in `libcnb`. To have a place for such common helpers, I added a new crate to the project: `libcnb-common`. I can see many more helpers moving over, especially file system related ones.

This PR also moves the `toml_file` module as the first shared code to this new crate. Since it's now in a library and `pub`, Clippy asks us to add documentation for possible error cases which I added as well.

Initially, I wanted this crate to not be published to crates.io to keep the maintainance and release burden low. However, we can't use a dependency without a version and only a path when we want to publish the crate that depends on it. This is the error message we would get at publishing time:

```
error: all dependencies must have a version specified when publishing.
dependency `libcnb-common` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

We therefore need another published `libcnb-*` crate. I adjusted `RELEASING.md`. No changelog entries have been made since this PR has no impact on users.

Ref: GUS-W-14028877


libcnb-common/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
edmorley, line 1: Could you turn this into a symlink, for parity with the other crates? :-)

libcnb-common/src/toml_file.rs
```diff
@@ -23,6 +28,11 @@ pub fn write_toml_file(
     Ok(())
 }
 
+/// Reads the file at the given paths and parses it as `A`.
```
edmorley, line 31:
```suggestion
/// Reads the file at the given path and parses it as `A`.
```

libcnb-common/src/toml_file.rs
```diff
@@ -14,6 +14,11 @@ pub enum TomlFileError {
     TomlSerializationError(#[from] toml::ser::Error),
 }
 
+/// Serializes the given value as TOML and writes it to the given file path.
+///
+/// # Errors
+///
+/// Will return `Err` if the file written or the value could not be serialized as a TOML string.
```
edmorley, line 21: I think there is a word missing after "written"?

edmorley (approved):

libcnb-common/LICENSE
Malax: Ooops, sure. :)

libcnb-common/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#![warn(clippy::pedantic)]
```
edmorley, line 1: (Sorry I missed this before) Should we add the `#![doc = include_str!("../README.md")]` that we use for all the other crates?

PR #664, opened 2023-08-30
Move GenericMetadata into libcnb-data and make it the default

### LINKED ISSUE #663: Move `GenericMetadata` into `libcnb-data` and make it the default
We expose a `GenericMetadata` type alias in `libcnb-package` because the packaging and testing processes don't need to deal with specific metadata in the buildpack descriptors.  We should move `GenericMetadata` into `libcnb-data` instead where it could be useful in other situations.

Furthermore, to support common use cases where dealing with metadata is not required, the `GenericMetadata` should be added as a default type.  E.g.;

```rust
type GenericMetadata = Option<toml::value::Table>;

pub enum BuildpackDescriptor<BM = GenericMetadata> {
    Single(SingleBuildpackDescriptor<BM>),
    Meta(MetaBuildpackDescriptor<BM>),
}

pub struct SingleBuildpackDescriptor<BM = GenericMetadata> { 
  // ...
}

pub struct MetaBuildpackDescriptor<BM = GenericMetadata> { 
  // ...
}

// no need to specify the metadata type here if we don't need to
let buildpack_descriptor = toml::from_str::<BuildpackDescriptor>(toml_str)?;
```

Malax:
Fixes #663, GUS-W


libcnb-data/src/generic.rs
```diff
@@ -0,0 +1,2 @@
+/// Generic TOML metadata.
+pub type GenericMetadata = Option<toml::value::Table>;
```
edmorley, line 2: Could `GenericMetadata` also be used here? (ie made the default and the tests simplified)
https://github.com/heroku/libcnb.rs/blob/7e800211d87a42f6ad60cfb626ca570f47539e7c/libcnb-data/src/layer_content_metadata.rs#L8
https://github.com/heroku/libcnb.rs/blob/7e800211d87a42f6ad60cfb626ca570f47539e7c/libcnb-data/src/layer_content_metadata.rs#L43

libcnb-data/src/buildpack/mod.rs
```diff
@@ -58,7 +59,7 @@ pub use version::*;
 /// ```
```
edmorley, line 59: The rustdocs for `BuildpackDescriptor`, `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` still use `<Option<toml::value::Table>>>`.

Now there is a default, that could be removed too. Also, whilst perhaps out of scope for this PR, I wonder if we need an example showing how to use them with a custom metadata type?

edmorley (approved):

libcnb-data/src/generic.rs
edmorley: Oh also here:
https://github.com/heroku/libcnb.rs/blob/785602d2a09dcf03c6c3be25354591ecbdec8639/libcnb-package/src/lib.rs#L39

Malax: Ah, that's cool! Changed it to use `GenericMetadata` (implicitly) as well. Thanks! 👍🏻

libcnb-data/src/buildpack/mod.rs
Malax: Changed to use the default.

PR #666, opened 2023-08-30
Support local buildpacks and meta-buildpacks in libcnb-test

colincasey:
The test runner has been modified to support a `BuildpackReference::LibCnbRs(BuildpackId)` variant which represents a buildpack (and any of it's dependencies) within the workspace that needs to be compiled for testing. This is similar to `BuildpackReference::Crate` but also supports meta-buildpacks.

This PR replaces #590 and should be easier to review now that similar refactorings from that PR have been incorporated into the codebase.


libcnb-test/src/test_runner.rs
```diff
@@ -56,6 +56,8 @@ impl TestRunner {
         config: C,
         f: F,
     ) {
+        check_required_pack_version();
```
edmorley, line 59: I think we should drop the Pack CLI version check, since:
- there's no way to only run it once before all tests (due to the way Cargo integration tests work), so we have to run it before every test
- our CI is already using Pack CLI 0.30 (thanks to https://github.com/buildpacks/github-actions already having updated the default version to it)
- this just leaves buildpack maintainers who will need to update - and realistically just the languages team at the moment (in the future libcnb-test might be used elsewhere, but by then this 0.30+ version check will be redundant and removed regardless)
- since the bug only affects this new feature (meta-buildpack support), we could just mention the pack 0.30+ requirement in the changelog entry for the new feature (+ also the rustdocs for `BuildpackReference::LibCnbRs`)

libcnb-test/src/build_config.rs
```diff
@@ -251,6 +252,8 @@ impl BuildConfig {
 pub enum BuildpackReference {
     /// References the buildpack in the Rust Crate currently being tested.
     Crate,
+    /// References a libcnb.rs buildpack within the Rust Workspace that needs to be packaged into a buildpack
+    LibCnbRs(BuildpackId),
     /// References another buildpack by id, local directory or tarball.
     Other(String),
```
edmorley, line 258: I find the current names a bit confusing - is there a way we can improve these?

Crate -> CurrentCrate?
LibCnbRs -> LocalCrate?


...or something else?

.gitignore
```diff
@@ -5,3 +5,4 @@
 Cargo.lock
 **/fixtures/*/target/
 **/fixtures/*/packaged/
+**/test-fixtures/buildpacks/*/target/
```
edmorley, line 8: We have quite a few different style of naming for the fixture directories - we should fix this at some point later (in a separate PR)

libcnb-test/src/build_config.rs
```diff
@@ -251,6 +252,8 @@ impl BuildConfig {
 pub enum BuildpackReference {
```
edmorley, line 252: The new feature (plus enum variant) needs a changelog entry

libcnb-test/src/build_config.rs
colincasey: i'm open to suggestions. if we want to keep the language close to Rust-isms (e.g.; `Crate`, `WorkspaceMember`) that kind of makes sense. it is a little weird when those combine with Buildpack-isms though because not everything is a crate. 

For example:
* using `WorkspaceMember("heroku/nodejs-engine")` or `LocalCrate("heroku/nodejs-engine")` for a component buildpack where `heroku/nodejs-engine` is the Buildpack Id & `buildpacks/nodejs-engine` is the Workspace path.
* using `WorkspaceMember("heroku/nodejs")` or `LocalCrate("heroku/nodejs")` for a composite buildpack where `heroku/nodejs` is the Buildpack Id & `meta-buildpacks/nodejs` is the Workspace path but it isn't a crate.

to avoid that mismatch what do you think of `WorkspaceBuildpack(BuildpackId)` @edmorley?

.gitignore
colincasey: i think this could be fixed if the buildpacks for the tests were copied into a temp folder and then used but, for that to work completely, there would need to be some way of setting the workspace directory also.  right now it defaults to the `libcnb.rs` workspace because that is where the integration tests run.

edmorley: I think you might have misread my comment - that wasn't what I was referring to? :-)

libcnb-test/src/build_config.rs
edmorley: Oh true, of course the meta-buildpacks aren't a crate.

`WorkspaceBuildpack` sounds good to me.

What do you think about also renaming `Crate` to `CurrentCrate`?

.gitignore
colincasey: yeah, my reply was more around the context of why i needed to add this line (which i was hoping to avoid), not the inconsistencies with fixture directory naming 😅

libcnb-test/src/build_config.rs
colincasey: Added

edmorley (commented):
Thank you for making those changes. I'll defer to Manuel for the main part of the review, since he's more familiar with the libcnb-package APIs than I am.

Malax:
Update: This is on my short list for this week for a review. I should get to this tomorrow.

libcnb-test/src/build.rs
```diff
@@ -1,62 +1,138 @@
...
+    let buildpack_dir_resolver = create_packaged_buildpack_dir_resolver(
+        package_dir.as_ref(),
         cargo_profile,
-        &cargo_env,
         target_triple.as_ref(),
-    )
-    .map_err(PackageCrateBuildpackError::BuildBinariesError)?;
+    );
+
+    // TODO: this could accidentally detect a packaged meta-buildpack twice. how should we ignore directories
```
colincasey, line 74: @Malax curious about your thoughts here.

libcnb-test/src/build.rs
```diff
@@ -1,62 +1,138 @@
...
+    Ok(PackagedBuildpackDir {
+        _root: package_dir,
+        path: buildpack_dir_resolver(buildpack_id),
+    })
+}
 
-    Ok(buildpack_dir)
+pub(crate) struct PackagedBuildpackDir {
+    _root: TempDir,
+    pub(crate) path: PathBuf,
```
Malax, line 125: It would be good to not have to have this additional struct, I removed most of these grouping structs from `libcnb-package` because they're confusing. We can do so easily by not creating a temporary directory in `package_buildpack` but instead pass in a target directory explicitly. The responsibility to create the temporary directory would automatically "bubble up" and end up in the place where the context of a single test lives.

All in all, we gain:
- A better isolated `package_buildpack` that does less implicit things.
  - Which also makes it easier to test (if we want to) or use in other contexts.
- Removal of the struct and, more importantly, the congnitive burden of having to remember what `PackagedBuildpackDir` actually means. When I saw the name for the first time, I immediately jumped to "it can't be just a path, otherwise if would be a `PathBuf`". But the name doesn't give a hint what that extra bit is.


Note: This will also effect `package_crate_buildpack`.

libcnb-test/src/build.rs
```diff
@@ -1,62 +1,138 @@
...
 
 #[derive(Debug)]
-pub(crate) enum PackageCrateBuildpackError {
-    BuildBinariesError(BuildBinariesError),
-    CannotAssembleBuildpackDirectory(std::io::Error),
+pub(crate) enum PackageTestBuildpackError {
     CannotCreateBuildpackTempDirectory(std::io::Error),
     CannotDetermineCrateDirectory(std::env::VarError),
-    CargoMetadataError(cargo_metadata::Error),
+    CannotReadBuildpack(TomlFileError),
```
Malax, line 132: `CannotReadBuildpack` what exactly? Given that it wraps a `TomlFileError` is probably the buildpack descriptor? We should use a more concrete name here I think.

libcnb-test/src/build.rs
```diff
@@ -1,62 +1,138 @@
...
+    CreateBuildpackDependencyGraph(BuildBuildpackDependencyGraphError),
     CrossCompileConfigurationError(String),
+    FindCargoWorkspace(FindCargoWorkspaceRootError),
```
Malax, line 135: When I see lines like this when I wrote them, I rename either one. If `FindCargoWorkspaceRoot` isn't a good name we should fix that instead of using a slightly different name for the (in this case) enum variant. If we think `FindCargoWorkspaceRoot` is fine, there is no reason why the enum variant should be named differnetly.

libcnb-test/src/build.rs
```diff
@@ -1,62 +1,138 @@
...
 #[derive(Debug)]
-pub(crate) enum PackageCrateBuildpackError {
-    BuildBinariesError(BuildBinariesError),
-    CannotAssembleBuildpackDirectory(std::io::Error),
+pub(crate) enum PackageTestBuildpackError {
     CannotCreateBuildpackTempDirectory(std::io::Error),
     CannotDetermineCrateDirectory(std::env::VarError),
-    CargoMetadataError(cargo_metadata::Error),
+    CannotReadBuildpack(TomlFileError),
+    CreateBuildpackDependencyGraph(BuildBuildpackDependencyGraphError),
```
Malax, line 133: Same as below, if there is a good reason to name this "create" instead of "build" we should rename `BuildBuildpackDependencyGraphError`. If there is no reason, the enum variant should follow the existing naming.

libcnb-test/src/build.rs
```diff
@@ -1,62 +1,138 @@
...
+use std::fs;
 use std::path::PathBuf;
 use tempfile::{tempdir, TempDir};
 
 /// Packages the current crate as a buildpack into a temporary directory.
 pub(crate) fn package_crate_buildpack(
     cargo_profile: CargoProfile,
     target_triple: impl AsRef<str>,
-) -> Result<TempDir, PackageCrateBuildpackError> {
+) -> Result<PackagedBuildpackDir, PackageTestBuildpackError> {
```
Malax, line 20: What up with this rename from `PackageCrateBuildpackError` to `PackageTestBuildpackError`? The previous name matches the error name. With this renaming its no longer the case. Should we rename the function too?

libcnb-test/tests/integration_test.rs
```diff
@@ -63,13 +64,21 @@ fn rebuild() {
 
 #[test]
 #[ignore = "integration test"]
-#[should_panic(
-    expected = "Could not package current crate as buildpack: BuildBinariesError(CannotDetermineBuildpackCargoTargetName(NoBinTargets))"
-)]
 fn buildpack_packaging_failure() {
-    TestRunner::default().build(BuildConfig::new("invalid!", "test-fixtures/empty"), |_| {
-        unreachable!("The test should panic prior to the TestContext being invoked.");
+    let result = std::panic::catch_unwind(|| {
```
Malax, line 68: Why are we not continuing to use `should_panic` here (and in the other tests)?

Malax (changes requested):
Overall good to go I think - some minor changes requested but nothing crazy I think.

libcnb-test/src/build.rs
colincasey: Addressed in https://github.com/heroku/libcnb.rs/pull/673

libcnb-test/src/build.rs
colincasey: I consider `package_crate_buildpack` to be a specialized version of `package_buildpack`.  All the errors are the same between both functions. Also, not all buildpacks are crates (i.e.; meta-buildpacks) which is why I chose to rename to `PackageTestBuildpackError` and don't see the need for the error name to align with the function names here.

libcnb-test/src/build.rs
colincasey: 👍 Good point.  The only reason I introduced this wrapper struct was so that the temporary directory wouldn't immediately go out of scope which would cause the buildpacks to be cleaned up before the actual test executes.  But it's not needed if I just move the temp dir creation to the runner.

libcnb-test/src/build.rs
colincasey: renamed to `CannotReadBuildpackDescriptor`

libcnb-test/src/build.rs
colincasey: renamed to `BuildBuildpackDependencyGraph`

libcnb-test/src/build.rs
colincasey: Most of my error names were carried over from my initial PR for this which had slightly different names for these same functions before refactoring. Renaming to `FindCargoWorkspaceRoot`

libcnb-test/src/build.rs
colincasey: @Malax given the above, would you be okay with `PackageTestBuildpackError` or `PackageBuildpackError`?

libcnb-test/tests/integration_test.rs
colincasey: Because this specific error now happens later on in the packaging process.  The error that happens now that we're reusing packaging logic is:
```
Could not package directory as buildpack! No `buildpack.toml` file exists 
at /some/unknown/directory/where/this/test/runs
```

To capture that error I can't use the `should_panic` macro, right?

libcnb-test/README.md
```diff
@@ -191,13 +191,15 @@ fn dynamic_fixture() {
 Building with multiple buildpacks, using [`BuildConfig::buildpacks`]:
 
 ```rust,no_run
+use libcnb_data::buildpack_id;
```
edmorley, line 194: (End users won't have libcnb_data in their direct dependencies list, but instead libcnb)

```suggestion
use libcnb::data::buildpack_id;
```

libcnb-test/tests/integration_test.rs
```diff
@@ -522,3 +541,42 @@ fn address_for_port_when_container_crashed() {
         },
     );
 }
+
+#[test]
+#[ignore = "integration test"]
+fn basic_build_with_libcnb_reference_to_single_buildpack() {
+    TestRunner::default().build(
+        BuildConfig::new("heroku/builder:22", "test-fixtures/procfile").buildpacks(vec![
```
edmorley, line 549: Since the test buildpacks always pass detection and don't use the procfile, they can use the empty fixture (using the procfile fixture but not the procfile buildpack made me wonder if there was something special about the procfile that meant it was still needed).

```suggestion
        BuildConfig::new("heroku/builder:22", "test-fixtures/empty").buildpacks(vec![
```

libcnb-test/tests/integration_test.rs
```diff
@@ -522,3 +541,42 @@ fn address_for_port_when_container_crashed() {
...
+            );
+        },
+    );
+}
+
+#[test]
+#[ignore = "integration test"]
+fn basic_build_with_libcnb_reference_to_meta_buildpack() {
+    TestRunner::default().build(
+        BuildConfig::new("heroku/builder:22", "test-fixtures/procfile").buildpacks(vec![
```
edmorley, line 568:
```suggestion
        BuildConfig::new("heroku/builder:22", "test-fixtures/empty").buildpacks(vec![
```

libcnb-test/README.md
colincasey: fyi - to make this change work i had to add `libcnb` as a dev dependency to `libcnb-test/Cargo.toml`

libcnb-test/src/build.rs
Malax: `PackageBuildpackError` seems good to me. Thanks for the explanation! :)

libcnb-test/tests/integration_test.rs
Malax: Hm, I'm not sure. This is not blocking for me, let's revisit this later after merging. :)

Malax (approved):
Approve to stop blocking, feel free to merge after fixing the two comments I made. :) 👍🏻

libcnb-test/Cargo.toml
```diff
@@ -12,13 +12,14 @@ readme = "README.md"
...
 fs_extra = "1.3.0"
+libcnb-common.workspace = true
 libcnb-data.workspace = true
 libcnb-package.workspace = true
 tempfile = "3.7.1"
 
 [dev-dependencies]
 indoc = "2.0.3"
 ureq = { version = "2.7.1", default-features = false }
+libcnb.workspace = true
```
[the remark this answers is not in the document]
colincasey: newline added

libcnb-test/tests/integration_test.rs
edmorley: > Why are we not continuing to use `should_panic` here (and in the other tests)?

So it turns out the reason is that the top level error handling had regressed, which is why the test only passed when using the lower level error: #704.

PR #668, opened 2023-09-07
Switch back to the example alpine builder in CI

edmorley:
In CI a `pack build` is performed using one of libcnb's example CNBs and the upstream `cnb/sample-builder:alpine` builder, in order to test that our static MUSL compile works with non-libc base images.

However, in #594 the version of that builder had to be pinned to an older release, to work around a temporary incompatibility between the latest version of the builder and the latest stable Pack CLI:
https://github.com/buildpacks/samples/issues/170

In addition, in #775 the image was switched to our builder image since the example images didn't use a CNB lifecycle that was compatible with Buildpack API 0.13.

Both issues have since been resolved, so we can switch back to the latest version of the upstream example alpine builder, giving us musl coverage again.

GUS-W-14086879.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,8 +89,5 @@ jobs:
         run: cargo run --package libcnb-cargo -- libcnb package
         working-directory: ./examples/basics
       - name: Pack build using examples/basics
-        # TODO: This test previously used a non-libc builder to validate the static musl cross-compilation.
-        # This image used the experimental image extensions feature which has to be explicitly enabled and doesn't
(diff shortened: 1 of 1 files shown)


edmorley:
Depends on https://github.com/buildpacks/samples/issues/170

Malax:
@edmorley should we close this as it won't apply cleanly anymore after #775. Or do you want to keep this around as a loud reminder?

edmorley:
I'm happy to close this for now; we have the TODO in the workflow file, plus I'm still subscribed to the upstream issue in their examples repo.

PR #669, opened 2023-09-07
Change Layer interface from &self to &mut self

schneems:
The primary driver for wanting to mutate a layer is a desire for stateful logging. I hit this in the consuming state-machine interface, and previously, Josh hit it in a logging interface designed to track indentation level.

Upside:

- Developers are now allowed to do whatever they want within a layer
- It's more technically correct to require a single reference for these functions. With the prior interface, it would be theoretically possible to execute the same Layer in parallel (though unlikely)
- We can now change things between function calls.

Downside:

- Mutable self means there's no guarantee something didn't change between function calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `libcnb`:
-  - Changed Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
+  - Changed `Layer` interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
 
 
 ## [0.17.0] - 2023-12-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `libcnb`:
+  - Change Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
+
 
 ## [0.17.0] - 2023-12-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - `libcnb`:
-  - Change Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
+  - Changed Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
 
 
 ## [0.17.0] - 2023-12-06
--- a/examples/execd/src/layer.rs
+++ b/examples/execd/src/layer.rs
@@ -21,7 +21,7 @@ impl Layer for ExecDLayer {
     }
 
     fn create(
-        &self,
+        &mut self,
         _context: &BuildContext<Self::Buildpack>,
         _layer_path: &Path,
     ) -> Result<LayerResult<Self::Metadata>, <Self::Buildpack as Buildpack>::Error> {
--- a/libcnb/src/build.rs
+++ b/libcnb/src/build.rs
@@ -85,7 +85,7 @@ impl<B: Buildpack + ?Sized> BuildContext<B> {
     /// #    }
     /// #
     ///     fn create(
-    ///         &self,
+    ///         &mut self,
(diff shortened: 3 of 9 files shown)


schneems:
For the "exploded layer" idea here's a sketch of what an API could look like. We could use the existing `LayerData` and `LayerResult` to introduce a get and set (read and write) API that gives us a lot of flexibility.

For example, instead of:


```
            let ruby_layer = context //
                .handle_layer(
                    layer_name!("ruby"),
                    RubyInstallLayer {
                        _in_section: section.as_ref(),
                        metadata: RubyInstallLayerMetadata {
                            stack: context.stack_id.clone(),
                            version: ruby_version.clone(),
                        },
                    },
                )?;
```

With a get/set combo we could have logging inside of `main.rs` like this:

```
let layer_data = context //
    .handle_layer(get_layer_name, GetLayerData::new())?;

let metadata = RubyInstallLayerMetadata {
    stack: context.stack_id.clone(),
    version: ruby_version.clone(),
};

let layer_result: LayerResult<RubyInstallLayerMetadata> = match metadata_diff(&layer_data, metadata) {
    MetadataDiff::Same(metadata) => {
        logger = logger.step("Using cache");
    }
    MetadataDiff::Different { old, now } => {
        let timer = logger.step("Clearing cache (metadata changed)").step_timed("Installing"); 
        let result = install_ruby(layer_data.path(), now);
        logger = timer.finish_timed_step();
        result
    }
    MetadataDiff::CannotDeserialize { old, now } => {
        let timer = logger.step("Clearing cache (cannot read metadata)").step_timed("Installing"); 
        let result = install_ruby(layer_data.path(), now);
        logger = timer.finish_timed_step();
        result
    }
    MetadataDiff::NoCache(metadata) => {
        let timer = logger.step_timed("Installing"); 
        let result = install_ruby(layer_data.path(), metadata);
        logger = timer.finish_timed_step();
        result
    }
}
let layer_data = context //
    .handle_layer(write_layer_name, WriteLayerData::new(layer_result))?;
```

Or the programmer could choose to push that logic to a function like

```
fn ruby_install(
	logger: Box<dyn SectionLogger>,
	env: &Env,
	context: BuildContext<RubyBuildpack>, 
	metadata: RubyInstallLayerMetadata,
) -> Result<(Box<dyn SectionLogger>, Env), RubyBuildpack::Error>;
```

I'm still working on it. There's some rough edges but I think I can get a prototype like that working.

schneems:
Here's a working PoC of a theoretical get/set based layer API https://github.com/heroku/buildpacks-ruby/pull/203

Malax (approved):
Heya, thanks for re-suggesting this! Good to see that the change would just work without a more complex refactoring. Bottom line, I think we should make this change.

What I would like to see is a way to get the value passed to `handle_layer` back out, or allowing to pass a mutable reference to `handle_layer`. With the change in this PR, we can modify, but code outside of the layer will never be able to see what changed. When we think about the buildpack output, it would be cool to be able to get the current output state back out. This could be a separate PR or part of this one - I think the changes are closely related.

> With the prior interface, it would be theoretically possible to execute the same Layer in parallel (though unlikely)

Since you needed to pass `handle_layer` an owned value, this would've not been possible. I also don't think that that would be a problem. The value implementing `Layer` is a description of a layer state and how to get there - it's not the layer itself. This is why the layer name is not part of the `Layer` trait. However, this is not really relevant to the change in this PR.

> Mutable self means there's no guarantee something didn't change between function calls.

True, but all modifications would come from either the `handle_layer` implementation or the functions defined in `Layer`. This softens up the issue a lot for me. If you want to modify in your `Layer` implementation you have to work with that in the same layer - the issue is contained.

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `libcnb`:
+  - Change Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
```
edmorley, line 15: (For consistency with the tense used in previous entries, and the tense used in the keep a changelog heading names)

```suggestion
  - Changed Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
```

edmorley (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- `libcnb`:
+  - Changed Layer interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
```
edmorley, line 15:
```suggestion
  - Changed `Layer` interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
```

PR #670, opened 2023-09-08
Derive Copy and Clone for LayerTypes

schneems:
I want to be able to inject `LayerType` into a struct instead of needing to hard code it. Currently this can be done but isn't ergonomic because `LayerTypes` is not clone, but it is serializable and deserializable and all values inside are copy. That allows us to hack around the issue currently:

```
struct InjectLayerTypes {
    visibility: LayerTypes,
	// ...
}

Imply Layer for InjectLayerTypes {
    // ...

    fn types(&self) -> LayerTypes {
        LayerTypes {
            launch: self.visibility.launch,
            build: self.visibility.build,
            cache: self.visibility.cache,
        }
    }

	// Or

    fn types(&self) -> LayerTypes {
        // Hack around LayerType not implementing `Clone`
        let toml_string =
            toml::to_string(&self.visibility).expect("LayerTypes can be serialized to toml");
        let layer_types: LayerTypes =
            toml::from_str(&toml_string).expect("Serialized data can be deserialized");

        layer_types
    }
```

This could be used to allow building of more generic interfaces. For example, in https://github.com/heroku/libcnb.rs/pull/598 I am trying to introduce a generic struct for use in setting environment variables. One gap in this implementation is that it assumes build, launch, cache should be true. It (or another struct) could instead make no assumptions and accept LayerTypes as an injectable value.


Malax (changes requested):
👍🏻 on the change itself, we should also implement `Copy`. After the changelog is fixed this is 🟢 from my side.

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `libcnb-data`:
   - `ExecDProgramOutputKey`, `ProcessType`, `LayerName`, `BuildpackId` and `StackId` now implement `Ord` and `PartialOrd`. ([#658](https://github.com/heroku/libcnb.rs/pull/658))
   - Add `generic::GenericMetadata` as a generic metadata type. Also makes it the default for `BuildpackDescriptor`, `SingleBuildpackDescriptor`, `MetaBuildpackDescriptor` and `LayerContentMetadata`. ([#664](https://github.com/heroku/libcnb.rs/pull/664))
+- `libcnb`:
+  - Struct `LayerTypes` now implements `Copy` and `Clone`. This allows injection of layer types into the struct implementing the `Layer` trait ([#670](https://github.com/heroku/libcnb.rs/pull/670)).
```
Malax, line 18: For consistency: 
```suggestion
  - Struct `LayerTypes` now implements `Copy` and `Clone`. This allows injection of layer types into the struct implementing the `Layer` trait. ([#670](https://github.com/heroku/libcnb.rs/pull/670))
```

Malax (approved):

CHANGELOG.md
edmorley: I think the `This allows injection of layer types into the struct implementing the `Layer` trait.` mention should be removed. It's a pretty obscure use-case that only covers a small subset of what the possible use-cases of making something `Copy` or `Clone` implies. I think it's sufficient (and more consistent with what we've used elsewhere in the changelog) to just list the added derived traits.

edmorley: I've resolved this as part of resolving the merge conflicts.

PR #673, opened 2023-09-13
Ignore file support when finding buildpacks

colincasey:
Previously we could rely on knowing that packaged buildpacks would be written to a location within the `target` directory configured in Crate. This was easy to determine through program code and could be supplied to `find_buildpack_dirs` so that packaged buildpacks weren't accidentally included when searching for buildpacks.

With the changes introduced in [#583](https://github.com/heroku/libcnb.rs/pull/583) it is no longer possible to easily determine the output location for packaged buildpacks. This PR adds the [ignore](https://crates.io/crates/ignore) crate to provide directory iteration that respects standard ignore files. This will give the user the ability to configure where we search within a project for buildpacks.

Related to #666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `libcnb-cargo`:
   - No longer outputs paths for non-libcnb.rs and non-meta buildpacks. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
   - Build output for humans changed slightly, output intended for machines/scripting didn't change. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
-- `libcnb-package`
-  - dropped `ignore` argument from `find_buildpack_dirs` that allowed for paths to be filtered out of the directory traversal. This function will now respect standard ignore files (like `.gitignore`) while traversing. 
+  - When performing buildpack detection, standard ignore files (`.ignore` and `.gitignore`) will be respected 
 
 ## [0.14.0] - 2023-08-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `libcnb-cargo`:
   - No longer outputs paths for non-libcnb.rs and non-meta buildpacks. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
   - Build output for humans changed slightly, output intended for machines/scripting didn't change. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
+- `libcnb-package`
+  - dropped `ignore` argument from `find_buildpack_dirs` that allowed for paths to be filtered out of the directory traversal. This function will now respect standard ignore files (like `.gitignore`) while traversing. 
 
 ## [0.14.0] - 2023-08-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,7 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `libcnb-cargo`:
   - No longer outputs paths for non-libcnb.rs and non-meta buildpacks. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
   - Build output for humans changed slightly, output intended for machines/scripting didn't change. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
-  - When performing buildpack detection, standard ignore files (`.ignore` and `.gitignore`) will be respected 
+  - When performing buildpack detection, standard ignore files (`.ignore` and `.gitignore`) will be respected. ([#673](https://github.com/heroku/libcnb.rs/pull/673))
 
 ## [0.14.0] - 2023-08-18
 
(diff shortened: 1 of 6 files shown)


libcnb-package/src/lib.rs
```diff
@@ -106,38 +106,20 @@ fn create_file_symlink<P: AsRef<Path>, Q: AsRef<Path>>(
...
-                    find_buildpack_dirs_recursive(&entry.path(), ignore, accumulator)?;
-                } else if let Some(file_name) = entry.path().file_name() {
-                    if file_name.to_string_lossy() == "buildpack.toml" {
-                        accumulator.push(path.to_path_buf());
-                    }
+/// Will return an `Err` if any I/O errors happen while walking the file system or any parsing errors
+/// from reading a gitignore file.
+pub fn find_buildpack_dirs(start_dir: &Path) -> Result<Vec<PathBuf>, ignore::Error> {
+    let mut buildpack_dirs = vec![];
+    for result in ignore::Walk::new(start_dir) {
```
edmorley, line 113: Would writing this using [Iterator::filter_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) rather than a loop be cleaner? (Iterators can be collected into a `Result`)

libcnb-package/src/buildpack_dependency_graph.rs
```diff
@@ -86,7 +85,7 @@ fn build_libcnb_buildpack_dependency_graph_node(
 #[derive(thiserror::Error, Debug)]
 pub enum BuildBuildpackDependencyGraphError {
     #[error("IO error while finding buildpack directories: {0}")]
```
edmorley, line 87: I think this error message needs updating, since it's not only IO errors that can occur in this enum variant now.

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `libcnb-cargo`:
   - No longer outputs paths for non-libcnb.rs and non-meta buildpacks. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
   - Build output for humans changed slightly, output intended for machines/scripting didn't change. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
+- `libcnb-package`
```
edmorley, line 24: As of #662 we're treating `libcnb-package` as an internal crate and so not including changes to it in the changelog, but instead only mentioning things that affect end-users of `libcnb-cargo`.

So this changelog entry should perhaps (a) go under `libcnb-cargo` not `libcnb-package`, (b) only mention that `.gitignore` is now respected, and not mention the signature change of `find_buildpack_dirs`at all.

edmorley (commented):

libcnb-package/src/lib.rs
Malax: > (Iterators can be collected into a Result)

and will also fail fast on the first `Err` since iterators are lazy. :tada:

Malax (changes requested):
My requested changes are the comments @edmorley made. ;-)

libcnb-cargo/tests/integration_test.rs
```diff
@@ -219,6 +219,47 @@ fn package_command_error_when_run_in_project_with_no_buildpacks() {
...
+
+    assert_ne!(output.status.code(), Some(0));
+    assert_eq!(
+        String::from_utf8_lossy(&output.stderr),
+        "🚚 Preparing package directory...\n🖥\u{fe0f} Gathering Cargo configuration (for x86_64-unknown-linux-musl)\n🏗\u{fe0f} Building buildpack dependency graph...\n🔀 Determining build order...\n❌ No buildpacks found!\n"
+    );
+
+    fs::remove_file(ignore_file).unwrap();
+
+    // when a git folder is not present, .gitignore can be used
```
Malax, line 245: Should this read `.gitignore` _cannot_ be used? Otherwise I can't tell what this comment means.

Malax (approved):
Two small comments, otherwise: ship it!

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `libcnb-cargo`:
   - No longer outputs paths for non-libcnb.rs and non-meta buildpacks. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
   - Build output for humans changed slightly, output intended for machines/scripting didn't change. ([#657](https://github.com/heroku/libcnb.rs/pull/657))
+  - When performing buildpack detection, standard ignore files (`.ignore` and `.gitignore`) will be respected
```
edmorley, line 24: Plus PR link:

```suggestion
  - When performing buildpack detection, standard ignore files (`.ignore` and `.gitignore`) will be respected.  ([#673](https://github.com/heroku/libcnb.rs/pull/673))
```

edmorley (approved):

libcnb-cargo/tests/integration_test.rs
colincasey: sorry, should have read "when a git folder is present...".  actually, i'll expand on both comments in this test with more detail around why.

CHANGELOG.md
colincasey: ugh, the post-create PR update changelog entries with link task.  i still forget this 50% of the time.

Malax: I also didn't see it when I commented on the missing period at the end. 🙄

edmorley (approved):

PR #675, opened 2023-09-14
Update example `cargo libcnb package` log output in READMEs

edmorley:
The logs output by `cargo libcnb package` have changed over recent PRs - this updates the example output to match.

There's still some improvements required to the output (eg #600), however, for now it at least reflects reality.

GUS-W-14121433.


Malax (approved):
Poor Ponda Baba.

PR #681, opened 2023-09-21
Use a consistent approach for organising test fixture files

edmorley:
Previously test fixtures were located under a mixture of:
* `<crate>/fixtures/`
* `<crate>/test-fixtures/`

Now  for consistency they are all located under:
`<crate>/tests/fixtures/`

This matches the pattern already used by the Python CNB, Procfile CNB etc.


.gitignore
```diff
@@ -1,8 +1,10 @@
-/target/
-/packaged/
+target/
```
edmorley, line 1: This style of entry (with no leading `/`) matches both items at the repository root, and also items within subdirectories - making the wildcard variants below redundant.

dzuelke (approved):

Malax (commented):
+1 for consistency, I'm not sure about the location though. Cargo will try to compile code it finds in `tests`, even when in subdirectories below `tests`. If we ever had a test fixture that Cargo would detect as test code (imagine a test fixture that is an app written in Rust) this location will give us trouble.

edmorley:
> Edit: it seems only `main.rs` directly in `test/fixtures` causes issues.

Yeah anything nested in a further subdirectory isn't picked up. There are already some `main.rs` files nested further (eg in the test buildpack fixtures).

Since all of our fixture will be in their own directory, and not loose in the `fixtures/` directory (plus the failure mode is pretty clear; it's not hard to debug undefined behaviour) I think this is fine :-)

Malax (approved):

PR #682, opened 2023-09-21
Switch to the new "component" vs "composite" buildpack terminology

### LINKED ISSUE #639: Buildpack descriptor naming differs from spec
The buildpack structs in `libcnb-data` are named `SingleBuildpackDescriptor` and `MetaBuildpackDescriptor` but the [terminology used is the spec](https://github.com/buildpacks/spec/blob/main/buildpack.md#cnb-terminology) is "component buildpack" and "composite buildpack". These structs should be renamed to `ComponentBuildpackDescriptor` and `CompositeBuildpackDescriptor`.

GUS-W-14161652.

edmorley:
Upstream has now officially defined names for what we had previously been calling meta-buildpacks vs single buildpacks: https://github.com/buildpacks/spec/blob/main/buildpack.md#cnb-terminology

All types, docs and internal functions have been renamed to match.

Fixes #639.
GUS-W-14161652.


CHANGELOG.md
```diff
@@ -13,13 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `libcnb-data`:
   - `ExecDProgramOutputKey`, `ProcessType`, `LayerName`, `BuildpackId` and `StackId` now implement `Ord` and `PartialOrd`. ([#658](https://github.com/heroku/libcnb.rs/pull/658))
-  - Add `generic::GenericMetadata` as a generic metadata type. Also makes it the default for `BuildpackDescriptor`, `SingleBuildpackDescriptor`, `MetaBuildpackDescriptor` and `LayerContentMetadata`. ([#664](https://github.com/heroku/libcnb.rs/pull/664))
+  - Added `generic::GenericMetadata` as a generic metadata type. Also makes it the default for `BuildpackDescriptor`, `SingleBuildpackDescriptor`, `MetaBuildpackDescriptor` and `LayerContentMetadata`. ([#664](https://github.com/heroku/libcnb.rs/pull/664))
```
colincasey, line 16: shouldn't this line should reflect the new names?

libcnb-package/src/buildpack_dependency_graph.rs
```diff
@@ -36,7 +36,7 @@ pub fn build_libcnb_buildpacks_dependency_graph(
                 .filter(|buildpack_directory| {
                     matches!(
                         determine_buildpack_kind(buildpack_directory),
-                        Some(BuildpackKind::LibCnbRs | BuildpackKind::Meta)
+                        Some(BuildpackKind::LibCnbRs | BuildpackKind::Composite)
```
colincasey, line 39: since we're renaming, i wonder if `LibCnbRs` should be renamed to something like `CompilableComponent` or `LibCnbComponent`.  not necessary but the current name looks like an unpronounceable string of consonants 😅

CHANGELOG.md
edmorley: Updated

libcnb-package/src/buildpack_dependency_graph.rs
edmorley: Yeah I'm not too keen on the current name. I've filed #683 to track this separately, since it's separate from this rename, and we'll need to figure out a better name (I like your ideas, but since there are several viable options including in this PR might hold it up whilst we pick one).

Malax: +1 on what Ed said. I remember having my issue with that name when I did the refactor recently, but couldn't come up with a better name on the spot.

Issue #595, opened 2023-07-12
Use GitHub Actions to perform new releases


edmorley:
Currently new libcnb.rs releases are performed via a sequence of locally run manual steps:
https://github.com/heroku/libcnb.rs/blob/main/RELEASING.md

We should (a) automate these, (b) perform them on GitHub Actions, in order to:
1. Reduce the chance of error
2. Reduce the toil for performing new releases
3. Improve security over using local creds
4. Allow for running tools to lint things like the semver bump before release (#596)

We might be able to make use of one or more of these tools:
https://github.com/killercup/cargo-edit#cargo-set-version
https://github.com/crate-ci/cargo-release
https://github.com/Byron/cargo-smart-release

PR #689, opened 2023-09-24
Add a GitHub Actions workflow for preparing a libcnb.rs release

edmorley:
Libcnb.rs uses a two stage release process. Until now both of those stages required a maintainer to perform the steps manually:
https://github.com/heroku/libcnb.rs/blob/main/RELEASING.md

This new GitHub Actions workflow replaces the manual steps in the first half of the release process, leading up to the creation of the "Prepare release" PR.

In the future, the second stage of the release process will be automated too, but that will require sorting out a service account (which will also have to be backed by a GitHub account, since currently crates.io only supports login via GitHub oauth).

See also:
https://github.com/killercup/cargo-edit#cargo-set-version
https://github.com/heroku/languages-github-actions/blob/main/.github/workflows/_buildpacks-release.yml

First part of #595.
GUS-W-13977983.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,99 @@
+name: Prepare release
+
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: "Which component of the version to increment"
+        required: true
+        default: "patch"
+        type: choice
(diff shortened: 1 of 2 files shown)


Malax (approved):
Neat!

PR #695, opened 2023-10-06
Add `GH_TOKEN` env to `gh` CLI invocation

Malax:
`gh` does not work properly without the `GH_TOKEN` environment variable. For the double release check, we fail the check if the invocation succeeds. Since the token is missing it will never succeed, making the check a no-op. This PR adds the missing token.

Ref: https://github.com/heroku/libcnb.rs/actions/runs/6299767340/job/17101136605#step:8:13


edmorley (approved):
Good spot, thank you!

PR #705, opened 2023-10-27
Add tracing to detect and build phases

joshwlewis:
This wraps both `build` and `detect` with tracing, which results in`.jsonl` tracing files compatible with [OpenTelemetry File Export](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/). This will grant tracing information for all libcnb.rs based buildpacks.

It's setup such that any `build` or `detect` errors are tracked as such with events and statuses.

`detect` is also instrumented with `detect-pass` or `detect-fail` events, so it should be clear if a detect phase errored, failed, or passed.

Additionally, this adds metadata from the buildpack descriptor (buildpack name, version, api version) to each span.


joshwlewis:
Closing in favor of #723.

Issue #704, opened 2023-10-26
Incorrect error message shown when cross-compile dependencies not installed


jose-fully-ported:
I was trying to run tests for the nodejs buildpack (to ensure the behavior around alternative package manager selection is tested, since it appears to have changed in the past two days) but am encountering an error:

```shell
# install rust
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

# clone the buildpack repo
git clone https://github.com/heroku/buildpacks-nodejs.git heroku/buildpacks-nodejs

# setup rust for the buildpack
cd heroku/buildpacks-nodejs
rustup update
rustup target add x86_64-unknown-linux-musl

# change to the buildpack in question
cd buildpacks/nodejs-npm-install

# attempt to run tests
INTEGRATION_TEST_CNB_BUILDER=heroku/builder:22 cargo test --locked -- --ignored --test-threads 16
```

<details>
  <summary>Output</summary>

```
warning: some crates are on edition 2021 which defaults to `resolver = "2"`, but virtual workspaces default to `resolver = "1"`
note: to keep the current resolver, specify `workspace.resolver = "1"` in the workspace root's manifest
note: to use the edition 2021 resolver, specify `workspace.resolver = "2"` in the workspace root's manifest
    Finished test [unoptimized + debuginfo] target(s) in 0.36s
     Running unittests src/main.rs (/Users/josediazgonzalez/go/src/github.com/heroku/buildpacks-nodejs/target/debug/deps/heroku_npm_install_buildpack-887ff546bb0ac9e4)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/integration_test.rs (/Users/josediazgonzalez/go/src/github.com/heroku/buildpacks-nodejs/target/debug/deps/integration_test-ecd4a784e3c3adea)

running 6 tests
test test_npm_install_with_lockfile ... FAILED
test test_npm_install_caching ... FAILED
test test_npm_install_new_package ... FAILED
test test_npm_start_script_creates_a_web_process_launcher ... FAILED
test test_npm_build_scripts ... FAILED
test test_npm_build_scripts_prefers_heroku_build_over_build ... FAILED

failures:

---- test_npm_install_with_lockfile stdout ----
[...]

PR #717, opened 2023-11-02
Add more integration tests to libcnb-test

edmorley:
Improves the integration test coverage in preparation for fixing #704, so as to (a) reduce the size of that PR, (b) make it easier to see the before/after of the error output.

The one testing scenario not added by this PR is the testing of the "cross-compile tools are missing" case, since it's going to be more complex to test (as the tools have to be installed for all of the other tests to pass) - so I'm saving it for another PR so we can discuss the possible approaches separately.

GUS-W-14416222.


colincasey (approved):
👍 lots of good tests added here

Issue #712, opened 2023-11-02
`log_error()`, `log_warning()`, `log_info()` and `log_header()` should use `.expect()` instead of `.unwrap()`


edmorley:
Currently there are lots of usages of `unwrap()` in production code, in the logging helpers in:
https://github.com/heroku/libcnb.rs/blob/2698a713edf19034e382aa702e12acc1516da8fa/libherokubuildpack/src/log.rs

Whilst the nature of these failure modes means that we can't use a Result (and so panicing is appropriate) - we should use `.expect()` instead of an `.unwrap()` so at least the panic message is more descriptive.

edmorley:
Note: The `log` module will be removed in the future, since it's been superseded by the `build_output` module.

Issue #555, opened 2023-01-31
Pack output eats libherokubuildpack colors


schneems:
Running this command:

```
$ pack build yolo --builder schneems-22 --path /tmp/rails61 --verbose
```

We should see header colors, but they're eaten by pack. Screenshot:

<img width="2153" alt="Screenshot 2023-01-31 at 3 23 10 PM" src="https://user-images.githubusercontent.com/59744/215891024-c1bc014a-9806-4b14-ad7c-c991b6e7644c.png">

Notice that the color of `[builder]` right next to `[Installing Ruby]` is different. I think what's happening is the color code is somehow being sent before `[builder]` and it sets it's own color, and then clears it. We get no color because of the clear.

I'm not sure if this is something we can fix here or need to go upstream with a bug report.

edmorley:
This is something that should be fixed improved (either here or upstream), but as a workaround in the meantime, update to latest pack CLI, since it trusts the builder:22 image by default. (The reason you are seeing the `[builder]` prefix is since the builder is being treated as untrusted.)

PR #721, opened 2023-11-06
Add `BuildpackOutput`

schneems:
This adds a new `build_output` module containing a `BuildOutput` interface, whose goal is to provide a standard format for outputting build information to the end user while a buildpack is running.

This implementation is an evolution of the approach prototyped in the Ruby CNB:
https://github.com/heroku/buildpacks-ruby/tree/dda4ede413fc3fe4d6d2f2f63f039c7c1e5cc5fd/commons/src/output


libherokubuildpack/Cargo.toml
```diff
@@ -45,6 +49,14 @@ termcolor = { version = "1.4.0", optional = true }
 thiserror = { version = "1.0.50", optional = true }
 toml = { workspace = true, optional = true }
 ureq = { version = "2.9.1", default-features = false, features = ["tls"], optional = true }
+lazy_static = { version = "1", optional = true }
+regex = { version = "1", optional = true }
```
edmorley, line 53: I'm pretty against adding a dependency on the `regex` crate here - since:
- it's a very large crate given unicode support (particularly for smaller buildpacks), and whilst some other parts of libcnb use `fancy_regex` (a) I have plans to remove those, (b) it's a different crate so using both crates causes overlap since they don't fully share the same deps
- looking at the two usages of the `regex` crate in the logger, it seems a non-regex approach would both be possible and likely more appropriate/simpler/quicker

libherokubuildpack/Cargo.toml
```diff
@@ -45,6 +49,14 @@ termcolor = { version = "1.4.0", optional = true }
 thiserror = { version = "1.0.50", optional = true }
 toml = { workspace = true, optional = true }
 ureq = { version = "2.9.1", default-features = false, features = ["tls"], optional = true }
+lazy_static = { version = "1", optional = true }
```
edmorley, line 52: This dependency can be removed once the regex usage in `libherokubuildpack/src/output/util.rs` is switched to a non-regex approach.

libherokubuildpack/src/output/util.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+/// Removes trailing whitespace from lines
+///
+/// Useful because most editors strip trailing whitespace (in test fixtures)
+/// but commands <https://github.com/heroku/libcnb.rs/issues/582> emit newlines
+/// with leading spaces. These can be sanatized by removing trailing whitespace.
+#[allow(dead_code)]
+pub(crate) fn strip_trailing_whitespace(s: impl AsRef<str>) -> String {
```
edmorley, line 48: I'm not sure I really understand the need for this feature?

It also uses the regex crate, which isn't ideal.

At the very least we should switch this to a non-regex solution, or else remove it outright.

libherokubuildpack/src/output/background.rs
```diff
@@ -0,0 +1,213 @@
...
+        /// remove it's internal state due to E0509:
+        ///
+        /// <https://github.com/rust-lang/rust/blob/27d8a577138c0d319a572cd1a464c2b755e577de/compiler/rustc_error_codes/src/error_codes/E0509.md>
+        ///
+        /// The workaround is to never allow invalid internal state by replacing the
+        /// inner value with a `None` when removing it. We don't want to expose this
+        /// implementation detail to the user, so instead we accept the panic, ensure
+        /// the code is exercised under test, and exhaustively document why this panic
+        /// exists and how developers working with this code can maintain safety.
+        #[allow(clippy::panic_in_result_fn)]
```
edmorley, line 144: (Redundant if we remove `print_interval` / background threads)

I think the `stop()` function can be refactored using combinators and `.expect()` to be clearer and also avoid the need for this `allow` (since it won't be using `panic!` any more)

edmorley (changes requested):
Thank you for working on this.

I've only had an initial review look so far - at 1600+ lines long this PR is pretty large (and I see that's after several simplifications already) - which will take some time to review, and is also larger than I was expecting for an initial landing of a logger feature.

The comments I've left so far focus on ways we can try to reduce the scope of this PR to make it easier to land something sooner. For example, by removing the `print_interval` feature (and all the associated background threads, locking, ... handling). 

I think when first landing a new feature it's best to keep the feature set smaller, and then incrementally add additional features in follow-up PRs, since:
- the time taken to review a PR is exponential wrt PR size
- each new feature is another potential place where design decisions/discussions need to be had, particularly for advanced features - such as the `print_interval` feature
- the `print_interval` feature is also something that not all buildpacks will even need (as mentioned in a meeting a few months ago) - so the chances are that it will require more discussion than some of the other more core logger features

Once those removals/simplifications are made, it will be easier for me to continue the review.

libherokubuildpack/src/output/util.rs
Malax: This should be functionally equivalent to [trim_end](https://doc.rust-lang.org/std/primitive.str.html#method.trim_end)

schneems: This function operates on all lines within the string, not just the trailing whitespace on the string itself. It is only used for testing. I'm:

- [ ] Changing the name to include "lines" and to reflect the similarity to `trim_end` I'll call it `trim_end_lines`
- [ ] Make it `cfg(test)` instead of `allow(dead_code)`

schneems: - [ ] Remove regex and lazy_static crates

libherokubuildpack/examples/print_style_guide.rs
```diff
@@ -0,0 +1,224 @@
...
+use toml as _;
+use ureq as _;
+
+#[allow(clippy::too_many_lines)]
+fn main() {
+    println!(
+        "{}",
+        formatdoc! {"
+
+            Living build output style guide
```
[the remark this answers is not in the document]
schneems: > Or at the very least, I think it would simplify/speed up this PR if this addition were kept separate, so we can have the discussion about it another time :-)

Done in manuel's PR, which I've pulled into this PR. 

This is my attempt at having non-stale guides. I'll save the reasons for that to a later discussion.

libherokubuildpack/Cargo.toml
```diff
@@ -45,6 +49,14 @@ termcolor = { version = "1.4.0", optional = true }
...
+regex = { version = "1", optional = true }
+const_format = { version = "0.2", optional = true }
 
 [dev-dependencies]
 tempfile = "3.8.1"
+libcnb-test = {workspace = true}
+indoc = "2"
+pretty_assertions = "1"
+fun_run = "0.1.1"
+ascii_table = { version = "4", features = ["color_codes"] }
```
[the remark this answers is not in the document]
schneems: I removed `ascii_table`, but `fun_run` is used in doctests.

libherokubuildpack/Cargo.toml
```diff
@@ -45,6 +49,14 @@ termcolor = { version = "1.4.0", optional = true }
...
 ureq = { version = "2.9.1", default-features = false, features = ["tls"], optional = true }
+lazy_static = { version = "1", optional = true }
+regex = { version = "1", optional = true }
+const_format = { version = "0.2", optional = true }
 
 [dev-dependencies]
 tempfile = "3.8.1"
+libcnb-test = {workspace = true}
+indoc = "2"
+pretty_assertions = "1"
```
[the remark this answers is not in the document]
schneems: > Switching to use pretty_assertions is unrelated to the build output work 

It's a major quality of life improvement for developing this feature. I'm open to removing it, but, like the print style guide, it served a useful development tool and would want to wait until all feedback is received to cut it out. It's not that I'm randomly throwing things into the PR; it's that these things were instrumental in building the feature.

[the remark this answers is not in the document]
Malax: I removed it for now in [2f7a13e](https://github.com/heroku/libcnb.rs/pull/721/commits/2f7a13e2799e36b2f6a430d3a39337e6227e3906)

Malax (approved):
After pairing sessions and subsequent iterations by both @schneems and me where we cut the diff in half, reduced complexity and redundancy, I consider this ready-to-merge at this point.

Re-adding the threading features will follow in a separate PR.

Since I was so heavily involved with the actual code, I defer the final review to @edmorley.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -227,10 +227,6 @@ where
         prefix_first_rest_lines(Self::PREFIX_FIRST, Self::PREFIX_REST, s.as_ref())
     }
 
-    pub fn mut_step(&mut self, s: impl AsRef<str>) {
-        writeln_now(&mut self.state.write, Self::style(s));
-    }
```
schneems: This served a purpose: when we want to log in a context that we cannot easily get the log from, this allows us to emit an output without changing the state.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -11,7 +11,7 @@
 //!
 //! output = output
 //!     .section("Ruby version")
-//!     .end_section();
+//!     .finish();
```
schneems, line 14: [comment] In the voice of subzero 🥶 finish him 🥶. 

I'm fine with this. Originally I wanted the code to not have the "finish" repetition, but I can go either way. One name makes it easier to remember

Malax: We could also suffix the `finish` or `end` functions with _what_ they finish. My thought process was that it's inconsistent to have "finish" and "end". From there, I thought that the suffixes aren't necessary. But not a hill I wanna die on. TLDR: I could live with `finish_section` or `end_section` as long as we use the same word across the whole API to denote the end of a state or the whole output.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    pub(crate) started: Option<Instant>,
+    pub(crate) state: T,
+}
+
+/// Various states for [`BuildpackOutput`] to contain.
+///
+/// The [`BuildpackOutput`] struct acts as an output state machine. These structs
+/// are meant to represent those states.
+#[doc(hidden)]
+pub mod state {
```
edmorley, line 44: It doesn't seem like the `state` module needs to be public?

If I remove the `pub` here all of `clippy --all-targets`, `cargo test` and `cargo test -- --ignored` succeed. (And the rustdocs for the `build_output` module is exercising the code as a consumer would - rustdocs can't access private APIs etc.)

If this module does need to be public, could you add something that will fail if the `pub` is removed (eg rustdocs somewhere, an integration test, a brief usage in `examples/basics/` or similar etc), or at least a comment explaining why this is pub+doc-hidden?

```suggestion
mod state {
```

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    pub struct Started<W> {
+        pub(crate) write: ParagraphInspectWrite<W>,
+    }
+
+    #[derive(Debug)]
+    pub struct Section<W> {
+        pub(crate) write: ParagraphInspectWrite<W>,
+    }
+
+    pub struct Stream<W: std::io::Write> {
```
edmorley, line 64: Similarly, this still compiles when `pub(crate)` is used:

```suggestion
    pub(crate) struct NotStarted<W> {
        pub(crate) write: ParagraphInspectWrite<W>,
    }

    #[derive(Debug)]
    pub(crate) struct Started<W> {
        pub(crate) write: ParagraphInspectWrite<W>,
    }

    #[derive(Debug)]
    pub(crate) struct Section<W> {
        pub(crate) write: ParagraphInspectWrite<W>,
    }

    pub(crate) struct Stream<W: std::io::Write> {
```

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+        pub(crate) started: Instant,
+        pub(crate) write: MappedWrite<ParagraphInspectWrite<W>>,
+    }
+}
+
+#[doc(hidden)]
```
edmorley, line 70: The `AnnounceSupportedState` trait is private, so I don't think this `#[doc(hidden)]` is needed?

```suggestion
```

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
+//! # Buildpack output
+//!
+//! Use the [`BuildpackOutput`] to output structured text as a buildpack is executing.
```
edmorley, line 3:
```suggestion
//! Use [`BuildpackOutput`] to output structured text as a buildpack is executing.
```

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+use std::fmt::Debug;
+use std::io::Write;
+use std::time::Instant;
+
+mod ansi_escape;
+mod duration_format;
+pub mod style;
+mod util;
+
+/// See the module docs for example usage.
```
edmorley, line 31: For UX on docs.rs and in an IDE, what do you think about duplicating the module docs here? (Longer term I think it's fine for both sections to differ - eg one of the two to be the canonical location for full examples, and the other a short example.)

Otherwise I get this in my IDE:

Screenshot 2024-02-02 at 15 32 07

And this in the generated rustdocs:

Screenshot 2024-02-02 at 15 33 58

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+        );
+        writeln_now(io, "");
+    }
+}
+
+impl<W> BuildpackOutput<state::NotStarted<W>>
+where
+    W: Write,
+{
+    pub fn new(io: W) -> Self {
```
edmorley, line 148: Could you add rustdocs for all of these public APIs?

Currently the generated docs are completely empty, eg:

Screenshot 2024-02-02 at 15 47 28

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,500 @@
...
+    fn flush(&mut self) -> std::io::Result<()> {
+        self.state.write.flush()
+    }
+}
+
+/// Internal helper, ensures that all contents are always flushed (never buffered).
+fn writeln_now<D: Write>(destination: &mut D, msg: impl AsRef<str>) {
+    writeln!(destination, "{}", msg.as_ref()).expect("Output error: UI writer closed");
+
+    destination.flush().expect("Output error: UI writer closed");
```
edmorley, line 306: (nit) The error message style used for these `.expect()`s don't match the guidelines here:
https://doc.rust-lang.org/std/error/index.html#common-message-styles

edmorley (commented):
I'm still reviewing, but I'm posting this partway through so you can start addressing these (if you wanted; it's fine to wait to the end too), plus since GitHub's review comment collapsing makes reviews with too many comments slight unwieldy, and hopefully splitting them up into two batches will help.

libherokubuildpack/src/buildpack_output/mod.rs
schneems: The states must be exposed so people can use them in their type signatures. For example:

```
//! use libherokubuildpack::buildpack_output::{BuildpackOutput, state::{Section, Started}};
//! use std::io::Write;
//!
//! let mut output = BuildpackOutput::new(std::io::stdout())
//!     .start("Heroku Ruby Buildpack")
//!     .warning("No Gemfile.lock found");
//!
//!      let output = install_ruby(output.section("Ruby Version"));
//!      output.finish();
//!
//! fn install_ruby<W>(mut output: BuildpackOutput<Section<W>>) -> BuildpackOutput<Started<W>>
//! where
//!    W: Write + Send + Sync + 'static
//! {
//!     output = output.step("Installing Ruby");
//!     // ...
//!    output.finish()
//! }
```

Changing that module to a different visibility makes it error:

```
5  | use libherokubuildpack::buildpack_output::{BuildpackOutput, state::{Section, Started}};
   |                                                             ^^^^^ private module
```

I'll add some more failing doc tests that exercise each state

libherokubuildpack/src/buildpack_output/mod.rs
schneems: Resolved via adding doc tests that exercise these structs

libherokubuildpack/src/buildpack_output/mod.rs
schneems: I had an issue with it that I mentioned in the commit https://github.com/heroku/libcnb.rs/commit/b9f6e2fdf8fcf0bf9d61480899912afce869e7bc. 

Though now when I delete the line, it is indeed fine:

```
$ git diff
diff --git a/libherokubuildpack/src/buildpack_output/mod.rs b/libherokubuildpack/src/buildpack_output/mod.rs
index d9c43dc..03e229b 100644
--- a/libherokubuildpack/src/buildpack_output/mod.rs
+++ b/libherokubuildpack/src/buildpack_output/mod.rs
@@ -160,7 +160,6 @@ pub mod state {
     }
 }
 
-#[doc(hidden)]
 trait AnnounceSupportedState {
     type Inner: Write;
 
⛄️ 3.1.4 🚀 /Users/rschneeman/Documents/projects/work/libcnb.rs (schneems/output)
$ cargo clippy --all-targets --all-features -- --deny warnings
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
```

I'm not 100% certain I know what caused the original problem or what change Manuel (or I) made that fixed it.

libherokubuildpack/src/buildpack_output/mod.rs
schneems: I think we could move it to an external document, like a README.md in the folder than use str include like https://github.com/schneems/magic_migrate/blob/3e43a72c329a2f03c016e0d37cd8ec085af7982f/src/lib.rs#L1 (though I'm not 100% sure that also works on a struct doc, but I'll try).

libherokubuildpack/src/buildpack_output/mod.rs
schneems: The prior docs largely referenced the print_style_guide.rs, which also existed as tests like struct visibility. I'll give everything a pass, but might defer some examples to bringing the style guide back in a future PR.

schneems: I've updated the docs, please take a look.

libherokubuildpack/src/buildpack_output/mod.rs
schneems: I've updated the docs, please take a look.

schneems:
I mentioned privately, but I'm leaving a note here as well. I've addressed all issues. I'm ready for another round.

libherokubuildpack/src/buildpack_output/style.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Decorate a URL for the build output.
+pub fn url(contents: impl AsRef<str>) -> String {
+    ansi_escape::inject_default_ansi_escape(ansi_escape::CYAN, contents)
+}
+
+/// Decorate the name of a command being run i.e. `bundle install`.
+pub fn command(contents: impl AsRef<str>) -> String {
+    value(ansi_escape::inject_default_ansi_escape(
+        ansi_escape::BOLD_CYAN,
+        contents.as_ref(),
```
edmorley, line 14: (the `body` argument to `inject_default_ansi_escape` is of type `impl AsRef<str>`, making the `as_ref()` here redundant)

```suggestion
        contents,
```

libherokubuildpack/src/buildpack_output/style.rs
```diff
@@ -0,0 +1,28 @@
...
+    ))
+}
+
+/// Decorate an important value i.e. `2.3.4`.
+pub fn value(contents: impl AsRef<str>) -> String {
+    let contents = ansi_escape::inject_default_ansi_escape(ansi_escape::YELLOW, contents.as_ref());
```
edmorley, line 20:
```suggestion
    let contents = ansi_escape::inject_default_ansi_escape(ansi_escape::YELLOW, contents);
```

libherokubuildpack/src/buildpack_output/util.rs
```diff
@@ -0,0 +1,223 @@
...
+
+        let newline_index = self.input.find('\n').map_or(self.input.len(), |i| i + 1);
+
+        let (line, rest) = self.input.split_at(newline_index);
+        self.input = rest;
+        Some(line)
+    }
+}
+
+pub(crate) fn prefix_lines<F: Fn(usize, &str) -> String>(contents: &str, f: F) -> String {
```
edmorley, line 33: I found this function confusing, since at first I thought it was the function that would be used by other parts of the `BuildOutput` module, rather than an private helper function only. 

A few things led to this misunderstanding:
- the fact this is `pub(crate) fn` rather than `fn` (it's not used outside of this file)
- the fact that it's ordered first in this file, above `prefix_first_rest_lines` (ie: I thought `prefix_lines` would be calling `prefix_first_rest_lines` not vice versa)

Could you remove the `pub(crate)` and move this function below `prefix_first_rest_lines`?

edmorley (changes requested):
Thank you for the updates. I spent ~5 more hours reading/paging in/reviewing today and am now at the end of my day, so leaving another interim review so you have the option to work through the comments sooner if you'd like :-)

libherokubuildpack/src/buildpack_output/mod.rs
edmorley: I'm not sure I like splitting out the `build_output` module docs to a `README.md`. The only reason we do that for the crate docs, is that we have to have a README.md for crates.io / GitHub, so it makes sense to combine the two (and we want to have the README examples covered in CI via doctests). 

For the `build_output` case, it means a README nested in a random directory, and means people reading the Rust code have to switch to another file to see the contents.

Also, I think longer term it's likely we'll want different contents for the module vs struct docs, so there won't be the need to de-duplicate the two. Plus duplication isn't really a problem anyway IMO.

libherokubuildpack/src/buildpack_output/util.rs
```diff
@@ -0,0 +1,223 @@
...
+        if index == 0 {
+            first_prefix.clone()
+        } else {
+            rest_prefix.clone()
+        }
+    })
+}
+
+#[derive(Debug)]
+pub(crate) struct LockedWriter<W> {
```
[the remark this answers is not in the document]
schneems: [TLDR] I'm okay making it test only for now, and I don't want to delay this PR any more than needed, but I'm also not happy with the ergonomics of this interface for a gen 1 experience.

-----
Full: Originally I was wrapping the Write instance in this and returning it back to the user. Manuel changed that behavior so that `BuildpackOutput<state::Stream<W>>` is `Write` instead. It's currently not very ergonomic to use the streaming code with `std::process::Command` streaming because it requires two `Write` instances and we only give you one. So either we'll make the user have to implement something like this, or we will provide them with something like this, or we'll give them a different abstraction (Like a version of Command.stream that writes to both stdout and stderr with the same (locked) writer.

Another way to think about it: When I wrote all those doc tests, I did not show streaming a `Command` on purpose because it's currently very gnarly with this stripped down implementation and missing other pieces of support from the rest of libherokubuildpack.

Manuel expressed interest in this concept of a LockedWriter, presumably for the purpose of implementing a `Command` stream that accepts a single writer.

libherokubuildpack/src/buildpack_output/ansi_escape.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+pub(crate) const RESET: &str = "\x1B[0m";
+
+pub(crate) const RED: &str = "\x1B[0;31m";
+pub(crate) const YELLOW: &str = "\x1B[0;33m";
+pub(crate) const CYAN: &str = "\x1B[0;36m";
+
+pub(crate) const BOLD_CYAN: &str = "\x1B[1;36m";
+pub(crate) const BOLD_PURPLE: &str = "\x1B[1;35m"; // Magenta
```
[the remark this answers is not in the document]
schneems: It's technically not purple, it's magenta `Magenta: \u001b[35m.` https://www.lihaoyi.com/post/BuildyourownCommandLinewithANSIescapecodes.html. However, people don't think of Heroku as being "magenta" we think of it as a purple color. So when you read it being used in the code you'll know the intent, when you look at the details of the implementation you'll (be hinted) that there is another term for the color that's used elsewhere.

libherokubuildpack/src/buildpack_output/ansi_escape.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
schneems: Done

libherokubuildpack/src/buildpack_output/duration_format.rs
```diff
@@ -0,0 +1,38 @@
+use std::time::Duration;
+
+pub(crate) fn human(duration: &Duration) -> String {
```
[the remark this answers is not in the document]
schneems: Seems good. I updated it to use this scheme.

libherokubuildpack/src/buildpack_output/util.rs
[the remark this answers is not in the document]
schneems: Made it cfg(test)

libherokubuildpack/src/buildpack_output/util.rs
```diff
@@ -0,0 +1,223 @@
+use std::fmt::Debug;
+use std::io::Write;
+use std::sync::{Arc, Mutex};
+
+/// Iterator yielding every line in a string. Every line includes existing newline character(s).
+pub(crate) struct LineIterator<'a> {
```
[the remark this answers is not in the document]
schneems: This is such a brilliant find. Thanks a ton. Great catch

libherokubuildpack/src/buildpack_output/ansi_escape.rs
```diff
@@ -0,0 +1,78 @@
+/// Smartly injects an ANSI escape sequence as the default into the given string.
```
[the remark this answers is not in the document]
schneems: Manuel refactored this code and wrote docs for it. I re-wrote the docs and changed the interface to restrict the codes that it supports. I also re-named the method accordingly.

libherokubuildpack/src/buildpack_output/ansi_escape.rs
```diff
@@ -0,0 +1,78 @@
...
+pub(crate) const BOLD_CYAN: &str = "\x1B[1;36m";
+pub(crate) const BOLD_PURPLE: &str = "\x1B[1;35m"; // Magenta
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn handles_explicitly_removed_colors() {
+        // Differentiate between color clear and explicit no color https://github.com/heroku/buildpacks-ruby/pull/155#discussion_r1260029915
```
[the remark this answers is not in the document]
schneems: It's no-longer used, can be removed

libherokubuildpack/src/buildpack_output/ansi_escape.rs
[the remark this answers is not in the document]
schneems: Removed the comment

libherokubuildpack/src/buildpack_output/mod.rs
schneems: I put the readme in the module docs and duplicated the example to the BuildOutput struct

edmorley (changes requested):
Thank you for the updates!

I've now been able to fully review everything, so hopefully these should now be all the comments :-)

libherokubuildpack/src/buildpack_output/util.rs
```diff
@@ -0,0 +1,183 @@
...
+            let _ = write!(acc, "{prefix}{line}");
+            acc
+        },
+    );
+
+    if lines.is_empty() {
+        f(0, "")
+    } else {
+        lines
+    }
```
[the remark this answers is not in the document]
schneems: > Re the intended behaviour, I'm curious why adding a prefix is still desirable when passed an empty str? "Blank" lines will actually have a contents of "\n" so this special casing isn't needed for them. ie: If I remove the special-casing here, the tests in the build_output/mod.rs still pass?

If someone calls step or section I think it would be weirder to see:

```

  - Seemingly floating bullet
```

Than this:

```
-
  - Sub-step under an empty bullet above
```

It's an exercise in determining the "least wrong" thing to do. I think showing the bullet point with nothing behind it at least gives a hint to the programmer that they emitted an empty output as opposed to "where on earth did this newline come from.

libherokubuildpack/src/buildpack_output/ansi_escape.rs
```diff
@@ -0,0 +1,115 @@
...
+const BOLD_PURPLE: &str = "\x1B[1;35m";
+
+#[derive(Debug)]
+#[allow(clippy::upper_case_acronyms)]
+pub(crate) enum ANSI {
+    Red,
+    Yellow,
+    BoldCyan,
+    BoldPurple,
+}
```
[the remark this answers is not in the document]
schneems: It's an internal interface that the end user never touches. We can refactor it at will, and I have no intention of opening this function up to end-user consumption. I don't think we need to block on it.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+    /// A warning should be used to emit a message to the end user about a potential problem.
+    ///
+    /// Multiple warnings can be emitted in sequence. The buildpack author should take care not to overwhelm the end user with
+    /// unnecessary warnings.
+    ///
+    /// When emitting a warning, describe the problem to the user, if possible, and tell them how to fix it or where to look next.
+    ///
+    /// Warnings should often come with some disabling mechanism, if possible. If the user can turn off the warning,
+    /// that information should be included in the warning message. If you're confident that the user should not be able to
+    /// turn off a warning; use an [`BuildpackOutput::error`] instead
```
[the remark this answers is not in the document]
schneems: Short: Added a word

Long:
Most people who aren't accustomed to warnings and errors and the distinctions between them tend to be shy about errors so they use a warning as an "unsure error." I'm writing docs with them in mind, to help give them more of a mental model for when it's okay to error. Likewise, I feel when people add warnings, they do so without consideration of edge cases and often without a mechanism to turn it off. My main goals here are "it's okay to error if you feel strong about it" and "you should consider edge cases and add escape valves, probably even if you don't think you'll need them." 

In the case of EOL/Insecure Ruby, if I KNOW that there's a RCE that will own their machine I will be fine to block them from deploying (I did this with sprockets, but only because I also could detect that a specific config flag was enabled as well). But most of the time I don't. I just know that their COULD be a problem with an insecure system. So I think that's a valid use of a warning in that scenario. 

> However, I don't think we can turn every non-disableable warning into an error.

I agree. I think adding the word "consider" here accomplishes that wiggle room. 

> Warnings should often come with some disabling mechanism, if possible. If the user can turn off the warning, that information should be included in the warning message. If you're confident that the user should not be able to turn off a warning, **consider using** a [`BuildpackOutput::error`] instead.

I think I would be okay upgrading some of those "cannot disable this warning" into a error, but only if it's a disable-able error. I don't know that I want to encourage "disable-able" errors though. It seems like an advanced topic.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+    ///
+    /// The most common use case is to stream the output of a running `std::process::Command` to the end user.
+    /// Streaming lets the end user know that something is happening and provides them with the output of the process.
+    ///
+    /// The result of this function is a `BuildpackOutput<state::Stream<W>>` which implements [`std::io::Write`].
+    ///
+    /// If you do not wish the end user to view the output of the process, consider using a `step` instead.
+    ///
+    /// This function will transition your buildpack output to [`state::Stream`].
+    pub fn start_stream(mut self, s: impl AsRef<str>) -> BuildpackOutput<state::Stream<W>> {
```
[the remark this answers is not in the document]
schneems: IDK when clippy decides to recommend it versus not. I added to all the other methods minus the finish ones that return the original Write.

Good catch.

I would like the streaming to also offer a mutating variant that accepts a closure function, but one PR at a time.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+
+/// Internal helper, ensures that all contents are always flushed (never buffered).
+fn writeln_now<D: Write>(destination: &mut D, msg: impl AsRef<str>) {
+    writeln!(destination, "{}", msg.as_ref()).expect("Output error: UI writer closed");
+
+    destination.flush().expect("Output error: UI writer closed");
+}
+
+#[cfg(test)]
+mod test {
```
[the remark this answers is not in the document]
schneems: > .write_paragraph()'s special handling of the empty line case (adding a blank line to the text used for one of the existing .warning() tests seems fine to test this)

This is tested via double warning and warning after buildpack.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+            - Done (finished in < 0.1s)
+        "};
+
+        assert_eq!(
+            expected,
+            strip_ansi_escape_sequences(String::from_utf8_lossy(&io))
+        );
+    }
+
+    fn strip_ansi_escape_sequences(contents: impl AsRef<str>) -> String {
```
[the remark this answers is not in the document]
schneems: Short: I'll add a unit test for warn/important/error/start color codes

Long:

I'm against testing color codes everywhere from personal experience in other libraries. I've said so to Manuel as well. In an ideal state I want the end developer to be able to write a test before any code executes. I also don't want to make the tests so cumbersome that future features or code aren't tested if testing is too cumbersome. I also don't want the state where one change in the style file requires many mechanical changes in other files that will produce PRs with large diffs that obscure the focus.

From this zoom level, the color is a UI concern. In the same way most developers don't assert that a certain style is bold in unit tests, they just make it bold. The color is tested at a lower level. If there's a regression or a problem, then we can revisit what kind of test would best capture that regression and where.

When I test, I test for behavior. I don't really **want** ansi codes, I want colored text, which is covered by earlier tests. I want an easy way to evaluate visual changes as a whole qualitatively. I.e. "Someone changed something in the build output, does it still look okay?" A test won't answer that, but a mechanism that outputs all the various styles for visual inspection will. That was one original purpose of the style guide.

Asserting against color codes yields test output that looks like this:

```
---- buildpack_output::test::test_color_codes stdout ----
thread 'buildpack_output::test::test_color_codes' panicked at libherokubuildpack/src/buildpack_output/mod.rs:558:9:
assertion `left == right` failed
  left: "\n# Heroku Ruby Buildpack\n\n! I am a warning\n\n! I am important\n\n! I am an error\n\n"
 right: "\n\u{1b}[1;35m# Heroku Ruby Buildpack\u{1b}[0m\n\n\u{1b}[0;33m! I am a warning\u{1b}[0m\n\n\u{1b}[1;36m! I am imporant\u{1b}[0m\n\n\u{1b}[0;31m! I am an error\u{1b}[0m\n\n"
```

If I REALLY wanted to test this behavior, I would rather want to test that a specific interface was exercised with a specific input. In Ruby I would do this with stubs/mocks.

All of the colored text is within:

- start
- warn
- important
- error

Of these three of them warn/important/error use the same interface. I wouldn't be against a unit test that asserts the behavior of `write_paragraph` given it passes in a specific color. Though that doesn't cover `start`. I'll do that now.

Eh, once I've gone that far, I might as well just assert the actual behavior. I'll add one test that exercises all these 4 interfaces. But as Error consumes, it must use a temp file

libherokubuildpack/src/buildpack_output/ansi_escape.rs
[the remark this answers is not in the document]
Malax: > then I realized that would mean my method signature was lying because it says it takes an &str.

Made me smile and I _really appreciate_ this train of thought! :) And I do agree it is lying and if we're talking about perfection I'd love to see a `AnsiEscape` type that neatly encapsulates everything that can be in an ANSI escape sequence, allows for parsing, validation and nice testing. I actually very briefly tried this in the initial review and refactor cycle before moving on. I realised that designing and implementing a full `AnsiEscape` is a big task and IMO out of scope for this PR. If we really want this, we should consider using [termcolor](https://docs.rs/termcolor/latest/termcolor/) where it's already implemented for us.

For the implementation here, I don't think we get enough value for the added code. The old constants are still around (for testing) hinting that we didn't really abstract. The new `ANSI` type is just a collection of specific combinations that we happen to need for the current functionality. Since this only internally used and the new `ANSI` type only guards what can be passed into `inject_default_ansi_escape` (not adding actual abstraction for downstream code) going the pragmatic route and leaving it as `&str` seems preferable to me in this case.

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `libherokubuildpack`:
   - `MappedWrite::unwrap` for getting the wrapped `Write` back out. ([#765](https://github.com/heroku/libcnb.rs/pull/765))
+  - Added build `buildpack_output` module. This will help buildpack authors provide consistent and delightful output to their buildpack users ([#721](https://github.com/heroku/libcnb.rs/pull/721))
```
edmorley, line 21: This will need rebasing on `main` and the changelog entry adding to the new `unreleased` section, now that 0.18.0 has been released.

libherokubuildpack/src/buildpack_output/duration_format.rs
```diff
@@ -0,0 +1,65 @@
...
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_display_duration() {
+        let duration = Duration::ZERO;
+        assert_eq!(human(&duration), "< 0.1s");
+        let duration = Duration::from_millis(99);
```
edmorley, line 29:
```suggestion
        assert_eq!(human(&duration), "< 0.1s");

        let duration = Duration::from_millis(99);
```

libherokubuildpack/src/buildpack_output/ansi_escape.rs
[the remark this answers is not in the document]
edmorley: > It's an internal interface that the end user never touches. We can refactor it at will, and I have no intention of opening this function up to end-user consumption. I don't think we need to block on it.

The reason we're discussing this is because after we'd reviewed this implementation, you changed the implementation even though it wasn't something we'd requested in the review comments. Changing things after initial review is not necessarily an issue, but I think it's then fair for us to discuss the changes, and not surprising if we ask you to change things back if we think the changes weren't for the better. And therefore, the "I don't think we need to block on it" seems like a bit of an unfair framing of any delay here.

Re this being internal only - I agree - but I think that strengths the position of the original API being fine. That is, since it's not a public API, it doesn't need to defend against mis-use quite as much, so using `&str` is fine (given the trade-offs Manuel mentioned).

libherokubuildpack/src/buildpack_output/mod.rs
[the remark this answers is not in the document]
edmorley: > This is tested via double warning and warning after buildpack.

I think that's a different case. I'm talking about a blank line partway through the `.warning()` message, ie:

```
            ! This is important
            !
            ! So is this
```

And not this (which is two `.warning()` calls in a row):

```
            ! This is important

            ! So is this
```

libherokubuildpack/src/buildpack_output/util.rs
```diff
@@ -33,6 +33,14 @@ pub(crate) fn prefix_lines<F: Fn(usize, &str) -> String>(contents: &str, f: F) -
...
 }
 
+/// A trailing newline aware writer
+///
+/// A paragraph style block of text has an empty newline before and after the text.
+/// When multiple paragraphs are emitted, it's important that they don't double up on empty newlines
+/// or the output will look off.
+///
+/// This writer seeks to solve that problem by preserving knowledge of prior newline writes and
+/// exposing that information to the caller.
```
edmorley, line 43: This new rustdocs is great - thank you!

libherokubuildpack/src/buildpack_output/mod.rs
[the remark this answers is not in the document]
edmorley: > I'm against testing color codes everywhere

Yeah I don't think we should test against them everywhere, but we should at least have some tests. (Which we now do, thank you.)

> In an ideal state I want the end developer to be able to write a test before any code executes

I don't understand this phrasing - what does "before any code executes" mean?

> From this zoom level, the color is a UI concern. In the same way most developers don't assert that a certain style is bold in unit tests, they just make it bold. The color is tested at a lower level. If there's a regression or a problem, then we can revisit what kind of test would best capture that regression and where.

I agree, however:
1. For libcnb, the UI **is** a concern, since buildpack maintainers are relying on us to get the output/style/colours right. If we were talking about the tests written in individual buildpacks, then yes, they shouldn't have to test against colours at all.
2. For libcnb, there was a gap in test coverage until the new test was added. (The lower level tests only tested eg ANSI escaping, they didn't actually show the style working as a whole.

libherokubuildpack/src/buildpack_output/util.rs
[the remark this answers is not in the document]
edmorley: > If someone calls step or section

I think there are some words missing here - did this mean "If someone calls step or section with an empty string argument"?

eg: `.section("")`

If so, I agree it improves the UX in that scenario :-)

Would you mind adding a small comment explaining why this conditional is needed?

edmorley:
There are some unresolved conversation threads that GitHub is collapsing in the UI, linking here in case it helps find them:
https://github.com/heroku/libcnb.rs/pull/721#discussion_r1486400980
https://github.com/heroku/libcnb.rs/pull/721#discussion_r1486455981
https://github.com/heroku/libcnb.rs/pull/721#discussion_r1486044249

libherokubuildpack/src/buildpack_output/ansi_escape.rs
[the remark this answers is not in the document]
schneems: My main point is that it's relatively orthogonal to the whole PR review **AT ALL**. it's an internal interface that does not leak to the end user. Focusing our time here, unless it's to strengthen the guarantee that it doesn't leak in the future is...I'll not say a waste of time, though perhaps not a great use of time. Feel free to change my mind.

libherokubuildpack/src/buildpack_output/mod.rs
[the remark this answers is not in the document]
schneems: Ahh. I didn't write that code. Manuel moved some other logic which was tested to this location, which isn't.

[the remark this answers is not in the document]
schneems: Short: I added a test.

Long: Thanks for the clarification. I thought that logic was tested elsewhere.

It was originally tested in https://github.com/heroku/libcnb.rs/blob/d4b83f99b736840a4ae09b13a21bad84b6975ae1/libherokubuildpack/src/output/style.rs#L229-L235 but manual didn't like the name of that method and removed it in https://github.com/heroku/libcnb.rs/commit/6b1cea99b9937a35a15bb1bb2d4da80aed803723.

Funny enough the logic mostly only existed to make testing easier. But it's good we added a test now so future tests are easier to make later.

libherokubuildpack/src/buildpack_output/mod.rs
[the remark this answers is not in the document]
schneems: > I don't understand this phrasing - what does "before any code executes" mean?

TDD style. As in, write the test before you write the code. The harder you make it to write a test, the harder it is to do TDD. I mostly use it as a heuristic that could also approximated as "could someone send me a meaningful PR with a failing test and let me fix it?"

> For libcnb, there was a gap in test coverage until the new test was added. (The lower level tests only tested eg ANSI escaping, they didn't actually show the style working as a whole.)

The original was designed such that testing all the permutations at the highest level wasn't needed, and someone working in this file can choose to trust the style interfaces. 

Previously, the individual style methods were tested, and then those methods were composed into another method that asserted the composition behaved as expected. I didn't realize we had lost some of that coverage when we moved away from the specialized methods (like bangify above).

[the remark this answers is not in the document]
schneems: Short: I think we can resolve this.

CHANGELOG.md
schneems: Conflicts in cargo.toml for features that were added then removed, all while I'm trying to not screw the PR history here so we can turn around and squash merge it in.  

I'll wait until all comments are final before trying.

libherokubuildpack/src/buildpack_output/util.rs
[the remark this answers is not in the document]
schneems: I pushed a commit. I refactored to make it clearer that the split_inclusive might not yield anything. Then I added tests and some docs for all these methods.

libherokubuildpack/src/buildpack_output/ansi_escape.rs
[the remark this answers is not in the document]
schneems: >  you changed the implementation even though it wasn't something we'd requested in the review comments

This is a response to a refactor by Manuel posed as a refactor rather than a comment. I originally did not expose this function even to the rest of the crate (It is `pub(crate)` here, but that's an oversite https://github.com/heroku/buildpacks-ruby/blob/dda4ede413fc3fe4d6d2f2f63f039c7c1e5cc5fd/commons/src/output/fmt.rs#L189). I think it was okay before (after visibility would have been adjusted), but I worry about accidental use in the future.

> If we really want this, we should consider using [termcolor](https://docs.rs/termcolor/latest/termcolor/) where it's already implemented for us.

I forgot we have a `log.rs`, I can delete that and we can shave off yet another dependency.

schneems:
Alright. I'm ready for another round. We all missed a HUGE piece of the public interface being that `log` was still in place. I feel like that's the level I'm aiming to get feedback on originally.

Really I'm asking for help getting this over the line, whatever that looks like. I know you're out of steam (maybe projecting? I'm out of steam).

libherokubuildpack/src/buildpack_output/util.rs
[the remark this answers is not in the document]
edmorley: Thank you for adding the test and comments in ab5d9a5579b3d9b1a7b54da80f27c9134f0d9b3b.

I see that commit also adjusted the implementation, with the reasoning of:

> I think it makes more logical sense to encode that information in the code, so I refactored

I absolutely agree in the general case that it makes sense to think about adjusting the code first, then documenting after, since clearer code can mean it becomes self-documenting.

However, in this specific case I found the code harder to understand (not from "what" it does, but the "why") after the refactor than it was before.

I think this is highlighted most clearly by looking at the rustdocs that added to explain what's happening:

> If an empty string is provided, ...

You'll note that they describe it as an "if an empty string" (which matches perfectly the conditional that was used before), and not "if there are no lines after splitting" (which is what the new implementation suggests). I know both implementations are equivalent functionality wise, but I think if the rustdocs are clearer when they use the former wording, then the code is too (and it's clearer when both the code and the rustdocs match in their explanation as it were).

I spoke with Manuel about this and said the previous implementation was clearer to him too.

As such in the interests of saving time here, I've reverted the implementation change (but left the docs and tests) in:
7d8fa16e84febe240ad2dfa4eb9587e5ee5666b0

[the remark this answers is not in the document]
edmorley: (Feel free to mark this as resolved when you've read it; leaving it open for now otherwise it's hard to find/see my reply)

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,719 @@
...
+        if !io.was_paragraph {
+            writeln_now(io, "");
+        }
+
+        writeln_now(
+            io,
+            ansi_escape::wrap_ansi_escape_each_line(
+                color,
+                prefix_lines(s.as_ref(), |_, line| {
+                    if line.chars().all(char::is_whitespace) {
```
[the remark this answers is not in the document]
edmorley: I see this change to `is_empty()` was reverted in 8be7f9069f71c1ff9505ed7b3f3d64569201cc0cto fix an additional case identified in a new test.

However, switching back to `line.chars().all(char::is_whitespace)` reintroduces a different bug.

I've fixed that bug by switching to `line.is_empty() || line == "\n"`, and updated the tests to catch this additional case, in:
5e20c4154d31fe55891419f1c479996af177c90c

(Feel free to mark this as resolved when you've read it; leaving it open for now otherwise it's hard to find/see my reply)

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -0,0 +1,819 @@
...
+                    // which preserves any trailing newline characters if present.
+                    if line.is_empty() || line == "\n" {
+                        String::from("!")
+                    } else {
+                        String::from("! ")
+                    }
+                }),
+            ),
+        );
+        writeln_now(io, "");
```
edmorley, line 294: I think there might be a bug here - if there is a trailing newline in the input text to `write_paragraph`, then a final newline is still added here, causing doubled up newlines.

For example, if a modify one of the tests to do this:

```rust
            .warning("hello\nworld\n")
```

(note the trailing `\n\)

Then the output I get is like this:

```
            # Example Buildpack

            ! hello
            ! world


            - Done (finished in < 0.1s)
```

(with two newlines before the "done")

Now in the `.warning("hello\nworld\n")` example it seems like calling the function like that is user error, since they shouldn't add the trailing newline. In which case it would seem reasonable to not support this case.

However, I wonder whether in practice when people start using `build_output`, whether they will write things using `indoc!` like so:

```rust
            .start("Example Buildpack")
            .warning(indoc! {"
                Some intro/title

                A longer explanation of what is wrong...
            "})
            .finish();
```

Or like this:

```rust
        let message = indoc! {"
            Some intro/title

            A longer explanation of what is wrong...
        "};

        // ...
            .start("Example Buildpack")
            .warning(message)
            .finish();
```

And in both of those cases, the resultant message has a trailing newline, and the only way to prevent the trailing newline is to use `indoc!` like so:

```rust
            .start("Example Buildpack")
            .warning(indoc! {"
                Some intro/title

                A longer explanation of what is wrong..."
            })
            .finish();
```

(note the quote mark is now on the line before)

Or this:

```rust
        let message = indoc! {"
            Some intro/title

            A longer explanation of what is wrong..."
        };

        // ...
            .start("Example Buildpack")
            .warning(message)
            .finish();
```

I wonder if it might be easy for people to misplace the quote mark in this case, and not realise they are introducing an additional newline?

A fix might be to wrap this `writeln_now(io, "")` call in a `if !io.was_paragraph` check, like is used higher up in the `write_paragraph` implementation (for the paragraph's opening newline).

What are your thoughts? :-)

edmorley (commented):
Thank you for the updates!

To help speed things up, I've pushed some commits to resolve a few things - more on those in these comments (feel free to resolve the comments once you've had a chance to read them):
- https://github.com/heroku/libcnb.rs/pull/721#discussion_r1487904397 (about `line.chars().all(char::is_whitespace)` in `write_paragraph`)
- https://github.com/heroku/libcnb.rs/pull/721#discussion_r1487834383 (about `prefix_lines` and peekable)
- https://github.com/heroku/libcnb.rs/pull/721#discussion_r1488031778 (about `log` re-addition)

This PR now very close to being done :-D The one last part to resolve is this possible bug (but the fix should be easy, if we decide it's something that should be fixed):
https://github.com/heroku/libcnb.rs/pull/721#pullrequestreview-1878199508

libherokubuildpack/src/buildpack_output/mod.rs
schneems: Great catch. I believe it's the build output's responsibility to handle this. Similar to input sanitization of a form element in a web app. I fixed by trimming the input in shared functions for all inputs.

libherokubuildpack/src/buildpack_output/mod.rs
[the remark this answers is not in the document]
schneems: Short: 👍🏻 I'm good with your change.

Long:
I reviewed the commit. I think this is something (hopefully) no one will ever care about or need to care about. Since it's whitespace either way, the end user should never see it. The primary driver of the original logic was test consistency which, btw, I like your testing strategy with indoc here. It's clear and robust.

libherokubuildpack/src/buildpack_output/util.rs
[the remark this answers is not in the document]
schneems: I don't care strongly about it either way. Thank you for updating! I'm resolving this conversation.

schneems:
I've reviewed your changes. Thank you for the help! I pushed a commit to handle more cases where stray trailing newlines can affect the output.

Regarding `log` and `error` I believe there's a bug in both of them when pushing to Heroku where the color bleeds over the `remote:` prefixes for any output that goes over 1 line. That's existed on main for a long time ago, it can be a bug for a bit longer.

edmorley:
> Regarding `log` and `error` I believe there's a bug in both of them when pushing to Heroku where the color bleeds over the `remote:` prefixes for any output that goes over 1 line. That's existed on main for a long time ago, it can be a bug for a bit longer.

Yeah I believe that existing bug is this one (whilst the issue filed here is about the prefix added by Pack when the builder is "untrusted", it ends up having a similar impact to the "remote:" prefix added by the Heroku Git server):
https://github.com/heroku/libcnb.rs/issues/555

We can close out that GitHub issue (plus also #712) when we remove `log` later on :-)

libherokubuildpack/src/buildpack_output/mod.rs
edmorley: Thank you for fixing this in 3ad2008e41dd1d4584142a8692f3aef677d2db96.

I spotted another case that wasn't tested (but already worked, so only needed updating the test):
https://github.com/heroku/libcnb.rs/pull/721/commits/aa19940d2b9c52eaad70468909eebf55c9b3e9ed

I also spotted what appears to be a limitation of the current `start_stream` / `line_mapped` / `MappedWrite` implementation. Specifically, if the streamed command's output has a trailing newline, then there are double newlines in the output. I think the fix for this will be more involved, so I would suggest we look at it later after merging this PR (but up to you). If you agree this is something we should fix longer term, would you mind filing a GitHub issue for it?

In the meantime I've added a testcase that demonstrates the issue (plus also tests the previously untested-but-working "newline in the middle of the output" case):
https://github.com/heroku/libcnb.rs/pull/721/commits/67e67aac889902924ad80d6a63b8b40fcd245940

Lastly, I spotted another small correctness/parity issue relating to streaming command output of blank lines (see commit message for more details) - I've fixed and updated the tests for it in:
https://github.com/heroku/libcnb.rs/pull/721/commits/d423789eee05fb9e108455ee64c0a403a7f25fdb

Malax (approved):
Re-approving to signal that changes made since my last approval are also approved :)

edmorley (commented):
@schneems I've pushed some additional commits / left one last comment:
https://github.com/heroku/libcnb.rs/pull/721#discussion_r1489694186

If you're happy to defer fixing the issue I found until later (and are happy with the other commits I pushed), I'm happy to call this done and I will approve/merge the PR :-)

libherokubuildpack/src/buildpack_output/mod.rs
schneems: >  I think the fix for this will be more involved, so I would suggest we look at it later after merging this PR (but up to you)

I pushed a fix that checks if a double newline was already added. We already had the infrastructure in place, it just wasn't being used. I modified your test case to remove that newline.

I'm good with everything else. Sorry for the delay this morning. I think we're good to go

schneems:
I commented inline but will again here as well:

> @schneems I've pushed some additional commits / left one last comment:
https://github.com/heroku/libcnb.rs/pull/721#discussion_r1489694186

If you're happy to defer fixing the issue I found until later (and are happy with the other commits I pushed), I'm happy to call this done and I will approve/merge the PR :-)

I pushed a fix for this case. It was fairly easy. If you really don't like it for some reason you can revert and merge and we can file an issue for it.

libherokubuildpack/src/buildpack_output/mod.rs
edmorley: > I pushed a fix that checks if a double newline was already added. We already had the infrastructure in place, it just wasn't being used. I modified your test case to remove that newline.

Oh fantastic - glad it was an easy fix - the fix looks great :-)

edmorley (approved):
Everything from the review is now resolved - let's merge!

Thank you for persevering through this 😄

PR #722, opened 2023-11-07
Improve consistency of error message wordings

edmorley:
* Use contractions per [Simplified Technical English](https://en.wikipedia.org/wiki/Simplified_Technical_English) guidelines (e.g. "Couldn't" instead of "Could not").
* Prefer using past tense for potentially transient failures (e.g. "Couldn't read buildpack.toml" rather than "Can't").
* Prefer using actual filenames rather than their spec names (e.g. "Couldn't read buildpack.toml" rather than "Couldn't read buildpack descriptor").
* Use "I/O error" instead of "IO error" or "IOError"
* Use display representations when including underlying error messages in the current error message where possible (ie: For most errors apart from `io::Error`). This also means preferring `panic!` over `.expect()`, since the latter uses the debug representation instead of display.
* Use "Error ..." as the prefix for all top level error message (i.e. at the `panic!` or `.expect()` call sites).
* Use `#[error(transparent)]` instead of `#[error("{0}")]` per: https://github.com/heroku/libcnb.rs/pull/720#discussion_r1383822568

GUS-W-14445374.

### DIFF
--- a/libcnb-test/src/test_runner.rs
+++ b/libcnb-test/src/test_runner.rs
@@ -59,9 +59,10 @@ impl TestRunner {
     ) {
         let config = config.borrow();
 
-        let cargo_manifest_dir = env::var("CARGO_MANIFEST_DIR")
-            .map(PathBuf::from)
-            .expect("Could not determine Cargo manifest directory");
+        let cargo_manifest_dir = env::var("CARGO_MANIFEST_DIR").map_or_else(
+            |error| panic!("Error determining Cargo manifest directory: {error}"),
+            PathBuf::from,
+        );
 
         let app_dir = {
             let normalized_app_dir_path = if config.app_dir.is_relative() {
@@ -80,7 +81,7 @@ impl TestRunner {
             // preprocessor. Skip app copying if no changes to the app will be made.
             if let Some(app_dir_preprocessor) = &config.app_dir_preprocessor {
                 let temporary_app_dir = app::copy_app(&normalized_app_dir_path)
-                    .expect("Could not copy app to temporary location");
+                    .expect("Error copying app fixture to temporary location");
 
                 (app_dir_preprocessor)(temporary_app_dir.as_path().to_owned());
 
@@ -91,7 +92,7 @@ impl TestRunner {
         };
 
         let buildpacks_target_dir =
-            tempdir().expect("Could not create a temporary directory for compiled buildpacks");
+            tempdir().expect("Error creating temporary directory for compiled buildpacks");
 
         let mut pack_command = PackBuildCommand::new(&config.builder_name, &app_dir, &image_name);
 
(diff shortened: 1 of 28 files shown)


libcnb-test/src/test_runner.rs
```diff
@@ -59,9 +59,10 @@ impl TestRunner {
     ) {
         let config = config.borrow();
 
-        let cargo_manifest_dir = env::var("CARGO_MANIFEST_DIR")
-            .map(PathBuf::from)
-            .expect("Could not determine Cargo manifest directory");
+        let cargo_manifest_dir = env::var("CARGO_MANIFEST_DIR").map_or_else(
+            |error| panic!("Error determining Cargo manifest directory: {error}"),
+            PathBuf::from,
+        );
```
Malax, line 65: Then I first read this, it made me stop for a moment and made me think about what's happening here. `.expect` seems a lot more natural.

However, `.expect` doesn't use the display representation which I assume is the reason we changed this to the new form. Can we add a short comment explaining this?

Malax (approved):

libcnb-test/src/test_runner.rs
edmorley: It's explained in the PR description

edmorley: I agree it's not immediately obvious about the implications between `.expect()` and a hand-written `panic!` (which is why I covered it in the PR description), but I think it's a bit too far towards "explaining how Rust works" to include an inline comment?

edmorley: Also, this is using the same style already used in other places, such as:
https://github.com/heroku/libcnb.rs/blob/5ab4b036a8dd18dc2d25a3d3ee49ed66ef3ec9d3/libcnb-test/src/test_runner.rs#L111-L127

Malax: I'll leave the final decision up to you - it's not a big deal. The other example seems even less problematic since only one operation takes place. I think what made me pause for a moment is the `map_or_else` where the else case is a side-effect.

edmorley: The initial edit I made was only `expect` -> `panic!`, however, Clippy then reported:

```
warning: called `map(<f>).unwrap_or_else(<g>)` on a `Result` value. This can be done more directly by calling `.map_or_else(<g>, <f>)` instead
  --> libcnb-test/src/test_runner.rs:62:34
   |
62 |           let cargo_manifest_dir = env::var("CARGO_MANIFEST_DIR")
   |  __________________________________^
63 | |             .map(PathBuf::from)
64 | |             .unwrap_or_else(|error| panic!("Error determining Cargo manifest directory: {error}"));
   | |__________________________________________________________________________________________________^ help: try: `env::var("CARGO_MANIFEST_DIR").map_or_else(|error| panic!("Error determining Cargo manifest directory: {error}"), PathBuf::from)`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#map_unwrap_or
```

PR #723, opened 2023-11-08
Add optional opentelemetry file export tracing

joshwlewis:
This is an alternative implementation of #705. 

This PR adds opt-in functionality (with the `trace` feature) that writes [OpenTelemetry File Export](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/) tracing information during build and compile. The tracing data includes buildpack attributes (like buildpack id, name), spans for build and detect, events for `build-success`, `detect-pass`, `detect-fail`, and also sets errors and span status in the case of errors.

GUS-W-14348835.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -5,6 +5,8 @@ use crate::detect::{DetectContext, InnerDetectResult};
 use crate::error::Error;
 use crate::platform::Platform;
 use crate::sbom::cnb_sbom_path;
+#[cfg(feature = "trace")]
+use crate::tracing::start_trace;
 use crate::util::is_not_found_error_kind;
 use crate::{exit_code, TomlFileError, LIBCNB_SUPPORTED_BUILDPACK_API};
 use libcnb_common::toml_file::{read_toml_file, write_toml_file};
@@ -138,14 +140,32 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
         buildpack_descriptor: read_buildpack_descriptor()?,
     };
 
-    match buildpack.detect(detect_context)?.0 {
-        InnerDetectResult::Fail => Ok(exit_code::DETECT_DETECTION_FAILED),
+    #[cfg(feature = "trace")]
+    let mut trace = start_trace(&detect_context.buildpack_descriptor.buildpack, "detect");
+
+    let detect_result = buildpack.detect(detect_context);
+
+    #[cfg(feature = "trace")]
+    if let Err(ref err) = detect_result {
+        trace.set_error(err);
+    }
+
+    match detect_result?.0 {
+        InnerDetectResult::Fail => {
+            #[cfg(feature = "trace")]
+            trace.add_event("detect-failed");
+            Ok(exit_code::DETECT_DETECTION_FAILED)
+        }
         InnerDetectResult::Pass { build_plan } => {
             if let Some(build_plan) = build_plan {
-                write_toml_file(&build_plan, build_plan_path)
-                    .map_err(Error::CannotWriteBuildPlan)?;
+                write_toml_file(&build_plan, build_plan_path).map_err(|err| {
+                    #[cfg(feature = "trace")]
+                    trace.set_error(&err);
+                    Error::CannotWriteBuildPlan(err)
+                })?;
             }
-
+            #[cfg(feature = "trace")]
+            trace.add_event("detect-passed");
             Ok(exit_code::DETECT_DETECTION_PASSED)
         }
     }
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -158,10 +158,11 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
         }
         InnerDetectResult::Pass { build_plan } => {
             if let Some(build_plan) = build_plan {
-                write_toml_file(&build_plan, build_plan_path).map_err(|err| {
+                write_toml_file(&build_plan, build_plan_path).map_err(|inner_err| {
+                    let err = Error::CannotWriteBuildPlan(inner_err);
                     #[cfg(feature = "trace")]
                     trace.set_error(&err);
-                    Error::CannotWriteBuildPlan(err)
+                    err
                 })?;
             }
             #[cfg(feature = "trace")]
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -132,7 +132,7 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
     #[cfg(feature = "trace")]
     let mut trace = start_trace(&buildpack_descriptor.buildpack, "detect");
 
-    let mut record_trace_err = |err: &dyn std::error::Error| {
+    let mut trace_error = |err: &dyn std::error::Error| {
         #[cfg(feature = "trace")]
         trace.set_error(err);
     };
@@ -140,13 +140,13 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
         .map_err(Error::CannotDetermineStackId)
         .and_then(|stack_id_string| stack_id_string.parse().map_err(Error::StackIdError))
         .map_err(|err| {
-            record_trace_err(&err);
+            trace_error(&err);
             err
         })?;
 
     let platform = B::Platform::from_path(&args.platform_dir_path).map_err(|inner_err| {
         let err = Error::CannotCreatePlatformFromPath(inner_err);
-        record_trace_err(&err);
+        trace_error(&err);
         err
     })?;
 
@@ -161,7 +161,7 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
     };
 
     let detect_result = buildpack.detect(detect_context).map_err(|err| {
-        record_trace_err(&err);
+        trace_error(&err);
         err
     });
 
@@ -175,7 +175,7 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
             if let Some(build_plan) = build_plan {
                 write_toml_file(&build_plan, build_plan_path).map_err(|inner_err| {
                     let err = Error::CannotWriteBuildPlan(inner_err);
-                    record_trace_err(&err);
+                    trace_error(&err);
                     err
                 })?;
             }
@@ -206,7 +206,7 @@ pub fn libcnb_runtime_build<B: Buildpack>(
     #[cfg(feature = "trace")]
     let mut trace = start_trace(&buildpack_descriptor.buildpack, "build");
 
-    let mut record_trace_err = |err: &dyn std::error::Error| {
+    let mut trace_error = |err: &dyn std::error::Error| {
         #[cfg(feature = "trace")]
         trace.set_error(err);
     };
@@ -215,19 +215,19 @@ pub fn libcnb_runtime_build<B: Buildpack>(
         .map_err(Error::CannotDetermineStackId)
         .and_then(|stack_id_string| stack_id_string.parse().map_err(Error::StackIdError))
         .map_err(|err| {
-            record_trace_err(&err);
+            trace_error(&err);
             err
         })?;
 
     let platform = Platform::from_path(&args.platform_dir_path).map_err(|inner_err| {
         let err = Error::CannotCreatePlatformFromPath(inner_err);
-        record_trace_err(&err);
+        trace_error(&err);
         err
     })?;
 
     let buildpack_plan = read_toml_file(&args.buildpack_plan_path).map_err(|inner_err| {
         let err = Error::CannotReadBuildpackPlan(inner_err);
-        record_trace_err(&err);
+        trace_error(&err);
         err
     })?;
 
@@ -237,7 +237,7 @@ pub fn libcnb_runtime_build<B: Buildpack>(
     }
     .map_err(Error::CannotReadStore)
     .map_err(|err| {
-        record_trace_err(&err);
+        trace_error(&err);
         err
     })?;
 
@@ -253,7 +253,7 @@ pub fn libcnb_runtime_build<B: Buildpack>(
     };
 
     let build_result = buildpack.build(build_context).map_err(|err| {
-        record_trace_err(&err);
+        trace_error(&err);
         err
     });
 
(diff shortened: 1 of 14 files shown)


libcnb/src/runtime.rs
```diff
@@ -138,14 +140,32 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
...
 
-    match buildpack.detect(detect_context)?.0 {
-        InnerDetectResult::Fail => Ok(exit_code::DETECT_DETECTION_FAILED),
+    #[cfg(feature = "trace")]
+    let mut trace = start_trace(&detect_context.buildpack_descriptor.buildpack, "detect");
+
+    let detect_result = buildpack.detect(detect_context);
+
+    #[cfg(feature = "trace")]
+    if let Err(ref err) = detect_result {
```
joshwlewis, line 149: There's a bunch of code that looks like this. I'm really looking forward to `inspect_err` becoming usable: https://github.com/rust-lang/rfcs/issues/3190.

edmorley: Looks very close to being stabilised at least 😄 
https://github.com/rust-lang/rust/pull/116866

edmorley:
@joshwlewis It looks like there were new versions of the opentelemetry crates released in the last couple of days:
https://github.com/open-telemetry/opentelemetry-rust/releases/tag/v0.21.0

Maybe worth picking those new versions up now?
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/CHANGELOG.md
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-stdout/CHANGELOG.md

joshwlewis:
> It looks like there were new versions of the opentelemetry crates released in the last couple of days

@edmorley thanks. Brought them in with [54f8fe4](https://github.com/heroku/libcnb.rs/pull/723/commits/54f8fe49d7818590454de604e9fc5b3e98d19607).

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+    trace::{Span as SpanTrait, Status, Tracer, TracerProvider as TracerProviderTrait},
+    KeyValue,
+};
+use opentelemetry_sdk::{
+    trace::{Config, Span, TracerProvider},
+    Resource,
+};
+use std::path::Path;
+
+const TELEMETRY_EXPORT_ROOT: &str = "/tmp/cnb-telemetry";
```
Malax, line 13: This hardcoded path still makes me a bit uneasy. But I also don't have a great solution either. Could we:

- Add `#[cfg(target_family = "unix")]` to compilation will at least fail on non-unixes. This way, it wouldn't compile for Windows instead of failing with an unsupported path at runtime.
- Include `libcnb` in the path since it is highly libcnb specific.

Malax (approved):
I only have one remaining comment. I think the proposed changes aren't too controversial so I'm approving already with the expectation that we resolve the comment before merging. :)

libcnb/src/tracing.rs
edmorley: A code comment here explaining what the seemingly arbitrary path is all about would be useful too :-)

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        println!("tracing_contents: {tracing_contents}");
+        assert!(tracing_contents.contains(phase));
+        assert!(tracing_contents.contains(event));
+        assert!(tracing_contents.contains(error_message));
+        assert!(tracing_contents.contains(buildpack.id.as_str()));
+        assert!(tracing_contents.contains(&buildpack.version.to_string()));
+        assert!(
+            tracing_contents.contains(&buildpack.name.expect("Expected buildpack.name to exist"))
+        );
```
edmorley, line 134: I'm a bit uneasy about this test. It doesn't actually test that the fields are set correctly, or that the file is valid JSON - only that substrings exist in the file in general.

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+impl Drop for BuildpackTrace {
+    fn drop(&mut self) {
+        self.span.end();
+        self.provider.force_flush();
+        global::shutdown_tracer_provider();
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
edmorley, line 88: I think we need some libcnb-test integration tests for this - perhaps by having the integration test run two buildpacks, with the second buildpack printing the output file contents from `/tmp`? (Since I presume there is no other easy way to get the contents out locally)

edmorley (changes requested):

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+    let lib_name = option_env!("CARGO_PKG_NAME").unwrap_or("libcnb");
+    let lib_version = option_env!("CARGO_PKG_VERSION");
+
+    let provider = TracerProvider::builder()
+        .with_simple_exporter(exporter)
+        .with_config(Config::default().with_resource(Resource::new(vec![
+            KeyValue::new("service.name", lib_name),
+            KeyValue::new("service.version", lib_version.unwrap_or_default()),
+        ])))
+        .build();
```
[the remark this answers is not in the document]
edmorley: Hmm so is `CARGO_PKG_NAME` the name of the top level crate only, or for dependencies is it set to the name of the dependency? When I wrote the comment above, I was presuming it was the former. Some code comments here would help with understanding :-)

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+    let exporter = match std::fs::File::options()
+        .create(true)
+        .append(true)
+        .open(&tracing_file_path)
+    {
+        Ok(file) => opentelemetry_stdout::SpanExporter::builder()
+            .with_writer(file)
+            .build(),
+        Err(_) => opentelemetry_stdout::SpanExporter::builder()
```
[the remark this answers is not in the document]
joshwlewis: A logger might help, but we'd probably only want to log in a verbose scenario, here. I don't think users would care if the telemetry is working or not. A log message about failed telemetry might even confuse users; they might not know telemetry exists or think it's their own telemetry.

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    let mut span = tracer.start(trace_name);
+    span.set_attributes(vec![
+        KeyValue::new("buildpack_id", buildpack.id.to_string().clone()),
+        KeyValue::new("buildpack_name", buildpack.name.clone().unwrap_or_default()),
+        KeyValue::new("buildpack_version", buildpack.version.to_string()),
+        KeyValue::new(
+            "buildpack_homepage",
+            buildpack.homepage.clone().unwrap_or_default(),
+        ),
```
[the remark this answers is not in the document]
joshwlewis: I added these mostly because they were available, and I've been told in the past that it's ok for wide events to be really wide (assuming you aren't generating keys). I'm also unsure how useful they'll be. Maybe lets try it and we can remove later?

libcnb/src/tracing.rs
joshwlewis: Added in ab24d40370f85bf88735399e14b24a00c8fbc40b.

libcnb/src/tracing.rs
joshwlewis: IMO, the keys/fields are the business of `opentelemetry`, and I don't think we should be re-testing it's serialization. 
I don't mind validating that it is JSON, though. Additonally, traversing to the values is a bit laborious, since object serialization looks like `[{"key":"resource-attr","value":{"stringValue":"resource-attr-val-1"}}]` rather than something a little more JSON-native like `{resource-attr: "resource-attr-val-1"}`.

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,136 @@
...
+
+const TELEMETRY_EXPORT_ROOT: &str = "/tmp/cnb-telemetry";
+
+pub(crate) struct BuildpackTrace {
+    provider: TracerProvider,
+    span: Span,
+}
+
+pub(crate) fn start_trace(buildpack: &Buildpack, phase_name: &'static str) -> BuildpackTrace {
+    let trace_name = format!("{}-{phase_name}", buildpack.id.replace(['/', '.'], "_"));
```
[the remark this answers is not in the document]
joshwlewis: Fixed in 65e851e.

libcnb/src/tracing.rs
joshwlewis: Addressed in [7bbf0e7](https://github.com/heroku/libcnb.rs/pull/723/commits/7bbf0e710aaa80b4925453509fe4520a6a7523c1).

libcnb/src/tracing.rs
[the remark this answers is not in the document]
joshwlewis: In my testing, `CARGO_PKG_NAME` resolves to the local package name (`libcnb`), and not the root package (not `heroku-some-buildpack`). In any case, I have added some commentary to the file in [7bbf0e7](https://github.com/heroku/libcnb.rs/pull/723/commits/7bbf0e710aaa80b4925453509fe4520a6a7523c1).

test-buildpacks/tracing/tests/fixtures/buildpacks/tracing-reader/bin/detect
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+set -eo pipefail
```
edmorley, line 2:
```suggestion
set -euo pipefail
```

test-buildpacks/tracing/tests/integration_test.rs
```diff
@@ -0,0 +1,32 @@
...
+fn test_tracing_export_file() {
+    let empty_app_dir = temp_dir().join("empty-app-dir");
+    fs::create_dir_all(&empty_app_dir).unwrap();
+
+    let mut build_config = BuildConfig::new("heroku/builder:22", &empty_app_dir);
+
+    // Telemetry file exports are not persisted to the build's resulting image,
+    // so to test that contents are emitted, a second buildpack is used to read
+    // the contents during the build.
+    build_config.buildpacks(vec![
```
edmorley, line 18: (nit; optional)

```suggestion
    build_config.buildpacks([
```

edmorley (commented):
I've run out of time tonight to completely finish the re-review, but this should be 95% of it...

test-buildpacks/tracing/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "tracing"
+version = "0.0.0"
+edition.workspace = true
+rust-version.workspace = true
+publish = false
+
+[dependencies]
+libcnb = { path = "../../libcnb", features = ["trace"] }
```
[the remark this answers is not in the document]
joshwlewis: Oh neat. I didn't even try, and assumed this wouldn't work.

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,172 @@
...
+    trace::{Span as SpanTrait, Status, Tracer, TracerProvider as TracerProviderTrait},
+    KeyValue,
+};
+use opentelemetry_sdk::{
+    trace::{Config, Span, TracerProvider},
+    Resource,
+};
+use std::{io::BufWriter, path::Path};
+
+// This is the directory in which `BuildpackTrace` stores Open Telemetry File
```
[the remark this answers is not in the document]
joshwlewis: I did this purposefully. Clippy will want me to wrap OpenTelemetry in backticks, because it thinks I'm referencing a type. If I fix the casing, I'll also have to add a lint block above it. You're probably right, using the correct casing is more correct.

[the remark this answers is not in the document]
joshwlewis: Fixed in [0c18d96](https://github.com/heroku/libcnb.rs/pull/723/commits/0c18d964ad0c9103ee3d5969da998326503ba3d9).

libcnb/src/tracing.rs
```diff
@@ -0,0 +1,172 @@
...
+// Exports. Services which intend to export the tracing data from libcnb.rs
+// (such as [cnb-otel-collector](https://github.com/heroku/cnb-otel-collector))
```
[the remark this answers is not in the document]
joshwlewis: Fixed in [0c18d96](https://github.com/heroku/libcnb.rs/pull/723/commits/0c18d964ad0c9103ee3d5969da998326503ba3d9).

libcnb/src/tracing.rs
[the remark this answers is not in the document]
edmorley: Ah I didn't realise that lint would trigger.

Looking now I see that lint has an allowlist - see the "Configuration" section of:
https://rust-lang.github.io/rust-clippy/master/index.html#/doc_markdown

I've opened an upstream PR to add "OpenTelemetry" to the default list:
https://github.com/rust-lang/rust-clippy/pull/11926

In the meantime until that change is on `stable`, we can override the default config using:
https://github.com/heroku/libcnb.rs/blob/main/clippy.toml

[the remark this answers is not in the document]
edmorley: (Marking as unresolved for now, since CI is failing at the moment due to this lint)

libcnb/src/runtime.rs
```diff
@@ -188,31 +209,53 @@ pub fn libcnb_runtime_build<B: Buildpack>(
         buildpack_dir: read_buildpack_dir()?,
         buildpack_descriptor: read_buildpack_descriptor()?,
         store,
-    })?;
+    };
+    #[cfg(feature = "trace")]
+    let mut trace = start_trace(&build_context.buildpack_descriptor.buildpack, "build");
```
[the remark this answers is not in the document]
joshwlewis: Yeah makes sense. Done in [f6486ea](https://github.com/heroku/libcnb.rs/pull/723/commits/f6486ea3c516a7e91efea5928189ca3c215f37d9).

libcnb/src/tracing.rs
joshwlewis: Yeah, fair enough on those points. I can't check the entire full file contents because 1) the keys are not consistently ordered and 2) the payload has timestamps for spans and events. I did add some rigor to these tests in 74a542092172fa26eccb8f5a6554e45707456850, though. It's now a bit more rigorous, checking text against expected keys and values. I feel pretty good about this balance now, especially given the test also validates the json.

edmorley: The new assertions look great, thank you!

edmorley (approved):
This is great - thank you so much for working on this! I can't wait to start using this in the CNBs themselves :-)

Issue #746, opened 2023-11-17
Example Ruby buildpack uses `unwrap()` in non-test code


edmorley:
We should not be using `.unwrap()` outside of tests.

These were found by the `unwrap_used` lint (which is about to be enabled across the whole repo).

https://github.com/heroku/libcnb.rs/blob/335b3514ade67db6e96df59d12e2a226325a42e2/examples/ruby-sample/src/layers/bundler.rs#L55-L57
https://github.com/heroku/libcnb.rs/blob/335b3514ade67db6e96df59d12e2a226325a42e2/examples/ruby-sample/src/layers/bundler.rs#L96
https://github.com/heroku/libcnb.rs/blob/335b3514ade67db6e96df59d12e2a226325a42e2/examples/ruby-sample/src/layers/bundler.rs#L108

edmorley:
Wontfix since the Ruby example has been removed in #757.

PR #747, opened 2023-11-17
Switch to workspace based lint configuration

edmorley:
As of the Cargo included in Rust 1.74, lints can now be configured in `Cargo.toml` across whole crates/workspaces:
https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html https://doc.rust-lang.org/stable/cargo/reference/manifest.html#the-lints-section https://doc.rust-lang.org/stable/cargo/reference/workspaces.html#the-lints-table

This reduces the boilerplate, and chance that we forget to enable lints in some targets. The only thing we need to remember is to add the `[lints] workspace = true` to any new crates in the future.

Making this switch exposed a few places where lints weren't enabled and issues had been missed, eg:
https://github.com/heroku/libcnb.rs/issues/746

Since this feature requires Rust 1.74, the MSRV has also been bumped. (Though we will have had to do so soon anyway to be able to start using `Result::inspect_err`, which is due in Rust 1.76, xref: https://github.com/heroku/libcnb.rs/pull/723#discussion_r1387095456)


Malax (approved):

Malax:
Woohoo! :)

PR #753, opened 2023-12-12
[Stacked PR] Refactor background timer logic

schneems:
The background timer prints dots to the UI while the buildpack performs some expensive operation. This requires we use threading and Rust really wants to be sure that we're not going to have data races and that our data is secure.

The original version of the code relied on wrapping our IO in an Arc<Mutex<W>> so that we don't "lose" it when we pass it to the background thread. This worked, but was overly complicated based on my limited understanding of working with lifetimes and threads. This version instead relies on the ability to retrieve a struct when we join the thread to get the IO object back out. This reduces complexity that the BuildLog interface needs to know about.


libherokubuildpack/src/output/background_dots.rs
```diff
@@ -0,0 +1,161 @@
...
+    #[allow(clippy::panic_in_result_fn)]
+    pub(crate) fn stop(mut self) -> std::thread::Result<W> {
+        // Ignore if the channel is closed, likely means the thread died which
+        // we want in this case.
+        match self.join_handle.take() {
+            Some(join_handle) => {
+                let _ = self.stop_signal.send(());
+                join_handle.join()
+            }
+            None => panic!("Internal error: Dot print internal state should never be None"),
```
schneems, line 111: Another alternative is to use something like 

- https://crates.io/crates/consume_on_drop

Which uses `ManuallyDrop` under  the hood https://github.com/markcsaving/consume_on_drop/blob/2da5decaba986313e8db1d0693b70058dd1757bd/src/lib.rs#L35-L37

In a similar territory I also found https://crates.io/crates/dispose which allows us to use the consuming interface for drop, though I'm not quite clear if it also allows you to manually get the value. This also uses ManuallyDrop https://github.com/ray-kast/dispose/blob/bd3675c164f1135ac39d35925c9d16a4497776be/src/dispose/disposable.rs#L20. 

The implementation is small enough we could re-write something similar without pulling in a dependency.

The benefit is we would get rid of this theoretically possible panic. The downside is that we trade it for an unsafe block. I'm unsure how exactly ManuallyDrop interacts with Drop i.e. if we ManuallyDrop::take the inner, then Drop is called on the outer struct, I don't know exactly how those play together.

I think this `Option` is good enough for our use case but I wanted to note that I looked into ManuallyDrop to understand all my options.

schneems: We talked about it. The panic is still not ideal, but it's not much code and seems okay. I added a debug_assert and moved it into it's own module for stronger guarantees only the public interfaces were being used.

libherokubuildpack/src/output/background.rs
```diff
@@ -0,0 +1,213 @@
...
+        fn drop(&mut self) {
+            // A note on correctness. It might seem that it's enough to signal the thread to
+            // stop, that we don't also have to join and wait for it to finish, but that's not
+            // the case. The printer can still emit a value after it's told to stop.
+            //
+            // When that happens the output can appear in the middle of another output, such
+            // as an error message if a global writer is being used such as stdout.
+            // As a result we have to signal AND ensure the thread is stoped before
+            // continuing.
+            if let Some(join_handle) = self.join_handle.take() {
```
icefoxen, line 98: Random passer-by with a random sanity check; is it desirable to do nothing if the `Option` is `None` on drop, or is that an invalid state that should fail noisily?

schneems: Thanks for the comments and cursory look!

It should be `None` if the `fn stop` is called as that's a consuming interface https://github.com/heroku/libcnb.rs/pull/753/files/87f4d6ca2ada895445c2899045eccd8468befcf0#diff-ac625dd49f4a759f34b78c2641dfe4e1b4b9b332b1f77c37caa18138a99c5b3eR145.

Possibly I can do a better job cross-referencing that functionality in the documentation as someone else suggested the same thing in the lobste.rs thread.

PR #757, opened 2024-01-02
Fix CI failures on `main`

### LINKED ISSUE #398: Example Ruby buildpack actually uses system Ruby
When I updated all usages of `heroku/buildpacks:20` in this repository to instead use `heroku/builder:22` the integration test for the example Ruby buildpack started failing.

After investigation, it turns out that the example Ruby buildpack is actually pretty broken at the moment.

For example, when the test runs, the image actually uses system Ruby, and so when run under `heroku/builder:22` (which is based on Heroku-22, which no longer has system Ruby), the image fails at runtime.

The reason it's falling back to system Ruby seems to be due to the manual `PATH` definition here, since it's missing the delimiter:
https://github.com/heroku/libcnb.rs/blob/6452900d4d5d3e198c8acc6fddb9d2b1e8f4cf4f/examples/ruby-sample/src/layers/ruby.rs#L46-L57

This then causes a broken `PATH`:

```
$ dr --entrypoint launcher test-ruby -- bash
heroku@b4cc8af292e1:/workspace$ env
...
PATH=/workspace/.gem/ruby/2.6.6/bin/layers/libcnb-examples_ruby/ruby/bin:/layers/libcnb-examples_ruby/bundler/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
LD_LIBRARY_PATH=/layers/libcnb-examples_ruby/ruby/layers/libcnb-examples_ruby/ruby/lib
```

(See also https://github.com/buildpacks/spec/issues/285)

I removed this manual `PATH` definition (which was also referencing Ruby 2.6, so completely wrong?), along with the seemingly also redundant `LD_LIBRARY_CONFIG` env var.

However then ran into another issue:

```
$ dr test-ruby
Could not find minitest-5.14.4 in any of the sources
Run `bundle install` to install missing gems.
```

As such, I'm going to exclude this example buildpack from the PR for `heroku/builder:22` -- we'll have to fix this example buildpack first.

cc @schneems @Malax

### LINKED ISSUE #479: Fix issues with `ruby-sample` integration test
The test has several issues:
1. The test uses `call_test_fixture_service` rather than something more realistic like `ureq`
2. The web server isn't compliant with the HTTP spec (so attempts to use `ureq` to make a request to it fail)
3. The app is unnecessarily complicated - it does not need to accept a posted value and invert it - replying with a fixed string to a GET would do (eg just using a Ruby stdlib HTTP server)
4. It uses a hardcoded sleep, rather than something like: https://github.com/heroku/libcnb.rs/blob/2a75bb6fc6f3817b07891bd66a944bb9b3508323/libcnb-test/tests/integration_test.rs#L77-L87
5. It uses Ruby 2.7, so has to use the `heroku/buildpacks:20` builder image (rather than the `heroku/builder:22` used by other tests), which slows down CI as it means another builder image has to be pulled.

### LINKED ISSUE #746: Example Ruby buildpack uses `unwrap()` in non-test code

### LINKED ISSUE #755: Integration tests failing on `main` due to example Ruby CNB bundler error
The integration tests in CI on `main` are failing with:

```
---- basic stdout ----
thread 'basic' panicked at /home/runner/work/libcnb.rs/libcnb.rs/libcnb-test/src/test_runner.rs:163:17:
Error performing pack build:

pack command failed with exit code 1!
error: test failed, to rerun pass `-p examples-ruby-sample --test integration_test`

## stderr:

[DEPRECATED] The `--path` flag is deprecated because it relies on being remembered across bundler invocations, which bundler will no longer do in future versions. Instead please use `bundle config set path '/layers/libcnb-examples_ruby/bundler'`, and stop using this flag
[DEPRECATED] The --binstubs option will be removed in favor of `bundle binstubs --all`
bundler-2.5.3 requires ruby version >= 3.0.0, which is incompatible with the
current version, 2.7.4
Unhandled error:
> BuildpackError(BundleInstallUnexpectedExitStatus(ExitStatus(unix_wait_status(1280))))
Buildpack will exit!
ERROR: failed to build: executing lifecycle: failed with status code: 51
```

(from https://github.com/heroku/libcnb.rs/actions/runs/7374728375/job/20065448224?pr=754#step:8:536)

IMO this is yet another reason to just delete the example Ruby CNB from this repo for now (see #398 for more reasons).

### LINKED ISSUE #756: Lint errors with Rust 1.75 / 1.76
The lint CI job on `main` is failing after the Rust 1.75 release, with:

```
warning: all fields have the same postfix: `name`
   --> libcnb-test/src/test_runner.rs:179:1
    |
179 | / pub(crate) struct TemporaryDockerResources {
180 | |     pub(crate) build_cache_volume_name: String,
181 | |     pub(crate) image_name: String,
182 | |     pub(crate) launch_cache_volume_name: String,
183 | | }
    | |_^
    |
    = help: remove the postfixes
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#struct_field_names
    = note: `-W clippy::struct-field-names` implied by `-W clippy::pedantic`
    = help: to override `-W clippy::pedantic` add `#[allow(clippy::struct_field_names)]`
```

In addition, locally using Rust 1.76 beta I get:

```
warning: items after a test module
   --> libcnb/src/layer_env.rs:598:1
    |
598 | mod tests {
    | ^^^^^^^^^
...
930 | const PATH_LIST_SEPARATOR: &str = ":";
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#items_after_test_module
    = note: `#[warn(clippy::items_after_test_module)]` on by default
    = help: move the items to before the test module was defined
```

edmorley:
Fixes the following CI failures on `main`:
1. Lint failures on the newly released Rust 1.75 (see #756) (plus pre-emptively fixes another for Rust 1.76 beta).
2. Integration test failures due to the Ruby example having picked up newer bundler, which is not compatible with the Ruby 2.7 it uses (see #755).

I've opted to delete the Ruby example entirely, since:
1. This is currently blocking libcnb.rs PRs/development.
2. It has multiple issues that really require a complete rewrite:
   - https://github.com/heroku/libcnb.rs/issues/398
   - https://github.com/heroku/libcnb.rs/issues/479
   - https://github.com/heroku/libcnb.rs/issues/746
   - https://github.com/heroku/libcnb.rs/issues/755
3. IMO including a full language CNB example in this repo is not something we should do, since it's both never going to fully support the language in a best practice way, and also will be too complicated to demonstrate libcnb concepts in the simplified way that we should be using in an example. Instead, we should stick to simple examples of concepts (in the spirit of the `basic` and `execd` examples, but add more of them), and then separately link out to our real-world CNB repos for users who want to do further reading.

As an added bonus, the removal will speed up CI a fair amount, since the Ruby integration test was very slow (due to it bundler plus the test using a different builder image, so another docker pull).

Fixes #756.
Fixes #755.
Closes #398.
Closes #479.
Closes #746.
GUS-W-14739082.
GUS-W-14739086.


Malax (approved):
Yeah, removing the Ruby example is a good call at this point. We could (in another PR) link some of our Heroku CNBs as examples for complete buildpacks written with libcnb for people that need/want it.

PR #765, opened 2024-01-30
Add `MappedWrite::unwrap`

Malax:
Currently, there is no way to get the wrapped `Write` value back out of a `MappedWrite`. In more complex scenarios where the mapping changes without closing the `Write`, we need to be able to get the wrapped value back out.

### DIFF
--- a/libherokubuildpack/src/write.rs
+++ b/libherokubuildpack/src/write.rs
@@ -13,12 +13,7 @@ pub fn mapped<W: io::Write, F: (Fn(Vec<u8>) -> Vec<u8>) + Sync + Send + 'static>
     marker_byte: u8,
     f: F,
 ) -> MappedWrite<W> {
-    MappedWrite {
-        inner: w,
-        marker_byte,
-        buffer: Vec::new(),
-        mapping_fn: Arc::new(f),
-    }
+    MappedWrite::new(w, marker_byte, f)
 }
 
 /// Constructs a writer that buffers written data until an ASCII/UTF-8 newline byte (`0x0A`) is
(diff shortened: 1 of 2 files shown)


libherokubuildpack/src/write.rs
```diff
@@ -57,10 +62,44 @@ pub struct TeeWrite<A: io::Write, B: io::Write> {
     inner_b: B,
 }
 
-impl<W: io::Write> MappedWrite<W> {
+impl<W> MappedWrite<W>
+where
+    W: io::Write,
+{
+    pub fn new<F: (Fn(Vec<u8>) -> Vec<u8>) + Sync + Send + 'static>(
```
edmorley, line 69: Should `new` instead be private? Previously it wasn't possible to create `MappedWrite` manually (only via `mapped()`) since there was no `MappedWrite::new` and the `MappedWrite` struct fields are private.

```suggestion
    pub new<F: (Fn(Vec<u8>) -> Vec<u8>) + Sync + Send + 'static>(
```

edmorley: Or to put it another way: If `MappedWrite::new` is now the recommended way for end users to construct a `MappedWrite` struct, then why do we still also have `mapped()`? For backwards compat and we'll remove later? (In which case perhaps we just remove now?) 

Or else, if `MappedWrite::new` isn't the recommended way for end users to construct a `MappedWrite` (and we still want them to use `mapped()`, then it seems `MappedWrite::new` should not be `pub`.

Malax: Yeah, it should not be `pub` here. I created it so that the weirdness with the `Option` wrapping is contained in one area as much as possible. Since we have `mapped` and `line_mapped` that configure `MappedWriter` in certain ways, using those is preferred. 

I removed the `pub` in [934ab1f](https://github.com/heroku/libcnb.rs/pull/765/commits/934ab1f3263f4f313880d56f3fd9d253a76c9d99).

edmorley (approved):

PR #766, opened 2024-01-31
Fix compile warnings when `trace` feature is disabled

Malax:
When the `trace` feature is disabled, several warnings will be emitted during compilation. These warnings started to appear after #723 was merged.

```
warning: unused variable: `err`
   --> libcnb/src/runtime.rs:135:28
    |
135 |     let mut trace_error = |err: &dyn std::error::Error| {
    |                            ^^^ help: if this is intentional, prefix it with an underscore: `_err`
    |
    = note: `#[warn(unused_variables)]` on by default

warning: variable does not need to be mutable
   --> libcnb/src/runtime.rs:135:9
    |
135 |     let mut trace_error = |err: &dyn std::error::Error| {
    |         ----^^^^^^^^^^^
    |         |
    |         help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: unused variable: `err`
   --> libcnb/src/runtime.rs:209:28
    |
209 |     let mut trace_error = |err: &dyn std::error::Error| {
    |                            ^^^ help: if this is intentional, prefix it with an underscore: `_err`

warning: variable does not need to be mutable
   --> libcnb/src/runtime.rs:209:9
    |
209 |     let mut trace_error = |err: &dyn std::error::Error| {
    |         ----^^^^^^^^^^^
    |         |
    |         help: remove this `mut`

warning: `libcnb` (lib) generated 4 warnings (run `cargo fix --lib -p libcnb` to apply 4 suggestions)
```

This PR fixes those.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -132,10 +132,14 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
     #[cfg(feature = "trace")]
     let mut trace = start_trace(&buildpack_descriptor.buildpack, "detect");
 
+    #[cfg(feature = "trace")]
     let mut trace_error = |err: &dyn std::error::Error| {
-        #[cfg(feature = "trace")]
         trace.set_error(err);
     };
+
+    #[cfg(not(feature = "trace"))]
+    let trace_error = |_: &dyn std::error::Error| {};
+
     let stack_id: StackId = env::var("CNB_STACK_ID")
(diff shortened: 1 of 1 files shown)


edmorley (approved):
Thank you - these were bugging me in CI logs.

(Once Rust 1.76 is released in Feb we can also switch to `inspect_err` which will clean up some of the rest of this handling)

Issue #728, opened 2023-11-09
Cross-compilation assistance: Help text "cross-compile" references are misleading


edmorley:
The current cross compilation assistance text uses phases like:

> For cross-compilation from Linux to x86_64-unknown-linux-musl, a C compiler and
linker for the target platform must be installed on your computer.

Whilst that's accurate for some platforms and host architectures, it's not correct for cases where the host architecture matches the target architecture - since in those cases no cross-compilation is taking place. Instead the key takeaway should be "using musl".

Rather than try and customise the message here for every host/target combination, I think we could instead still use a generic message, but phrase it with the emphasis on "in order to make the packaged buildpack portable, we need to use musl rather than glibc, which requires the relevant musl toolchain" etc.

See:
https://github.com/heroku/libcnb.rs/blob/main/libcnb-package/src/cross_compile.rs

Issue #726, opened 2023-11-09
Cross-compilation assistance: Env vars not set when targeting `x86_64_unknown_linux_musl` from Linux


edmorley:
Currently the cross-compile assistance does not set any env vars in this branch:
https://github.com/heroku/libcnb.rs/blob/4ec39b6fdf171f4f241f2d1fb3c3443b3cec53e6/libcnb-package/src/cross_compile.rs#L49-L53

Whilst some compiles will succeed with that implementation (since Rust will automatically use `musl-gcc` if it's found), it breaks when cross-compiling for a different architecture.

All `musl-gcc` does is compile using musl rather than glibc, using the same architecture as the host. (As opposed to the wrappers like `x86_64-unknown-linux-musl` which also target the different architecture.)

In addition, since none of the `CC_*` type env vars are set, if the buildpack uses any Rust dependencies that contain C components (such as the `ring` crate used for TLS), then the build will fail like so:

```
The following warnings were emitted during compilation:

warning: cc1: error: unrecognized command-line option '-m64'

error: failed to run custom build command for `libz-sys v1.1.12`

Caused by:
  process didn't exit successfully: `/buildpacks-python/target/debug/build/libz-sys-bfd25c9fd11b1325/build-script-build` (exit status: 1)
  --- stdout
  cargo:rerun-if-env-changed=LIBZ_SYS_STATIC
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-env-changed=ZLIB_NO_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_x86_64-unknown-linux-musl
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS_x86_64_unknown_linux_musl
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG_ALLOW_CROSS
  cargo:rerun-if-env-changed=PKG_CONFIG_ALLOW_CROSS
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64-unknown-linux-musl
  cargo:rerun-if-env-changed=PKG_CONFIG_x86_64_unknown_linux_musl
  cargo:rerun-if-env-changed=TARGET_PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG
  cargo:rerun-if-env-changed=PKG_CONFIG_SYSROOT_DIR_x86_64-unknown-linux-musl
[...]

Issue #724, opened 2023-11-09
libcnb-package: Incorrect GCC binary used for target `aarch64_unknown_linux_musl`


edmorley:
In #577 support was added for Linux aarch64, however, the GCC binary name used was `aarch64-linux-gnu-gcc` rather than the MUSL variant (in this case `aarch64-linux-musl-gcc`) used by all of the other target/host combinations:
https://github.com/heroku/libcnb.rs/blob/4ec39b6fdf171f4f241f2d1fb3c3443b3cec53e6/libcnb-package/src/cross_compile.rs#L64

I believe this is an unintentional bug. (If instead there is a reason for it, it needs to have a comment added.)

edmorley:
Checking an ARM64 Ubuntu 22.04 image with `musl-tools` installed, I see that `aarch64-linux-musl-gcc` does exist, and that it does differ from `aarch64-linux-gnu-gcc` - which confirms my belief that we should be using the former instead.

The `gnu` variant is simply a symlink to standard GCC:

```
root@5d2e9841af3b:/# ls -al /usr/bin/aarch64-linux-gnu-gcc
lrwxrwxrwx 1 root root 6 Aug  5  2021 /usr/bin/aarch64-linux-gnu-gcc -> gcc-11
``

Whereas the `musl` variant is this wrapper script:

```
root@5d2e9841af3b:/# cat /usr/bin/aarch64-linux-musl-gcc
#!/bin/sh
exec "${REALGCC:-aarch64-linux-gnu-gcc}" "$@" -specs "/usr/lib/aarch64-linux-musl/musl-gcc.specs"
```

Issue #725, opened 2023-11-09
Cross-compilation assistence: Set the `CXX_*` and `AR_*` environment variables too


edmorley:
The cross compile toolchain docs at https://github.com/messense/homebrew-macos-cross-toolchains suggest setting all three of these env vars (adjusted appropriately for the name of the target):

```
CC_x86_64_unknown_linux_gnu=x86_64-linux-gnu-gcc
CXX_x86_64_unknown_linux_gnu=x86_64-linux-gnu-g++
AR_x86_64_unknown_linux_gnu=x86_64-linux-gnu-ar
```

However, currently we're only setting the `CC_*` variant, not the `CXX_*` and `AR_*` env vars.

Whilst this works for all of the Rust crates we happen to use so far, for completeness we should probably set them all, in case someone uses a crate that relies on a tool that needs them to be set.

See:
https://github.com/heroku/libcnb.rs/blob/main/libcnb-package/src/cross_compile.rs

PR #769, opened 2024-02-01
Refactor cross compile assistance

runesoerensen:
We currently support targeting `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` on macOS and Linux, but the cross compile assistance provided is inconsistent depending on the host's CPU architecture. This refactoring aims to improve that across all supported `target_triple` and host OS/architecture combinations by providing helpful assistance for supported scenarios.

* Fixed https://github.com/heroku/libcnb.rs/issues/728 by improving the help text to include accurate advice for the `target_triple` relative to the current host OS *and* architecture.
* Fixed https://github.com/heroku/libcnb.rs/issues/726 by adding support for targeting `x86_64-unknown-linux-musl` from AARCH64 Linux.
* Improves maintainability by sharing CrossCompileAssistance::Configuration initialization with relevant environment variables (when appropriate) in one place (which will be useful if we want to set additional environment variables in the future, i.e. https://github.com/heroku/libcnb.rs/issues/725).
* Fixes part of https://github.com/heroku/libcnb.rs/issues/724 by adding support for targeting `aarch64-unknown-linux-musl` from AARCH64 Linux. The gcc binary used when targeting the same architecture is now `musl-gcc` on both `aarch64` and `x86-64`.
    * However, the binary used when the architectures don't match is still the `*-gnu-gcc` variant (as for instance `aarch64-linux-musl-gcc` is only available on aarch64 Linux (this requires further investigation, and a separate issue should be created for that)).
* Adds support for targeting `aarch64-unknown-linux-musl` from aarch64 macOS.
* Adds strict host architecture checks to ensure that `CrossCompileAssistance::Configuration` or `CrossCompileAssistance::HelpText` is only returned when supported. Otherwise `CrossCompileAssistance::NoAssistance` is returned.

It also:
* Drops support for a gcc binary (`x86_64-linux-musl-gcc`) installed by a previously recommended macOS toolchain. Since the current toolchain has been recommended for nearly 2 years it's probably safe to assume (or fair to require) that users have update to the recommended toolchain. 
The brew formula/revision has also been bumped, so the missing dependencies mentioned here https://github.com/heroku/libcnb.rs/pull/312#issuecomment-1038044636 are now included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,11 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `libcnb`:
   - Changed `Layer` interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
+- `libcnb-package`:
+  - Improved the consistency of cross-compilation assistance provided across all supported `target_triple` and host OS/architecture combinations. [#769](https://github.com/heroku/libcnb.rs/pull/769)
 
 ### Added
 
 - `libherokubuildpack`:
   - `MappedWrite::unwrap` for getting the wrapped `Write` back out. ([#765](https://github.com/heroku/libcnb.rs/pull/765))
+- `libcnb-package`:
+  - Added cross-compilation assistance for `aarch64-unknown-linux-musl` (on macOS and ARM64 Linux) and `x86_64-unknown-linux-musl` (on ARM64 Linux). [#769](https://github.com/heroku/libcnb.rs/pull/769)
 
 ## [0.17.0] - 2023-12-06
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,17 +11,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improved the consistency of cross-compilation assistance provided across all supported `target_triple` and host OS/architecture combinations. [#769](https://github.com/heroku/libcnb.rs/pull/769)
+- Added cross-compilation assistance for `aarch64-unknown-linux-musl` (on macOS and ARM64 Linux) and `x86_64-unknown-linux-musl` (on ARM64 Linux). [#769](https://github.com/heroku/libcnb.rs/pull/769)
 - `libcnb`:
   - Changed `Layer` interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
-- `libcnb-package`:
-  - Improved the consistency of cross-compilation assistance provided across all supported `target_triple` and host OS/architecture combinations. [#769](https://github.com/heroku/libcnb.rs/pull/769)
 
 ### Added
 
 - `libherokubuildpack`:
   - `MappedWrite::unwrap` for getting the wrapped `Write` back out. ([#765](https://github.com/heroku/libcnb.rs/pull/765))
-- `libcnb-package`:
-  - Added cross-compilation assistance for `aarch64-unknown-linux-musl` (on macOS and ARM64 Linux) and `x86_64-unknown-linux-musl` (on ARM64 Linux). [#769](https://github.com/heroku/libcnb.rs/pull/769)
 
 ## [0.17.0] - 2023-12-06
 
--- a/libcnb-package/src/cross_compile.rs
+++ b/libcnb-package/src/cross_compile.rs
@@ -40,35 +40,37 @@ fn generate_assistance(
     help_text: &str,
     target_triple: &str,
 ) -> CrossCompileAssistance {
-    if which(gcc_path).is_err() {
-        CrossCompileAssistance::HelpText(help_text.to_string())
-    } else if gcc_path == "musl-gcc" {
-        CrossCompileAssistance::Configuration {
-            cargo_env: Vec::new(),
+    match which(gcc_path) {
+        Err(_) => CrossCompileAssistance::HelpText(help_text.to_string()),
+        Ok(_) => {
+            if gcc_path == "musl-gcc" {
+                CrossCompileAssistance::Configuration {
+                    cargo_env: Vec::new(),
+                }
+            } else {
+                let cargo_env = vec![
+                    (
+                        // Required until Cargo can auto-detect the musl-cross gcc/linker itself,
+                        // since otherwise it checks for a binary named 'musl-gcc' (which is handled above):
+                        // https://github.com/rust-lang/cargo/issues/4133
+                        OsString::from(format!(
+                            "CARGO_TARGET_{}_LINKER",
+                            target_triple.to_uppercase().replace('-', "_")
+                        )),
+                        OsString::from(gcc_path),
+                    ),
+                    (
+                        // Required so that any crates that call out to gcc are also cross-compiled:
+                        // https://github.com/alexcrichton/cc-rs/issues/82
+                        OsString::from(format!("CC_{target_triple}")),
+                        OsString::from(gcc_path),
+                    ),
+                ];
+                CrossCompileAssistance::Configuration { cargo_env }
+            }
         }
-    } else {
-        let cargo_env = vec![
-            (
-                // Required until Cargo can auto-detect the musl-cross gcc/linker itself,
-                // since otherwise it checks for a binary named 'musl-gcc' (which is handled above):
-                // https://github.com/rust-lang/cargo/issues/4133
-                OsString::from(format!(
-                    "CARGO_TARGET_{}_LINKER",
-                    target_triple.to_uppercase().replace('-', "_")
-                )),
-                OsString::from(gcc_path),
-            ),
-            (
-                // Required so that any crates that call out to gcc are also cross-compiled:
-                // https://github.com/alexcrichton/cc-rs/issues/82
-                OsString::from(format!("CC_{target_triple}")),
-                OsString::from(gcc_path),
-            ),
-        ];
-        CrossCompileAssistance::Configuration { cargo_env }
     }
 }
-
 pub enum CrossCompileAssistance {
     /// No specific assistance available for the current host and target platform combination.
     NoAssistance,
--- a/libcnb-package/src/cross_compile.rs
+++ b/libcnb-package/src/cross_compile.rs
@@ -64,7 +64,7 @@ pub fn cross_compile_assistance(target_triple: impl AsRef<str>) -> CrossCompileA
         }
         Err(_) => CrossCompileAssistance::HelpText(format!(
             r"For cross-compilation from {0} {1} to {target}, a C compiler and
-linker for the target platform must be installed on your computer:
+linker for the target platform must be installed:
 
 {help_text}
             
--- a/libcnb-package/src/cross_compile.rs
+++ b/libcnb-package/src/cross_compile.rs
@@ -12,23 +12,23 @@ pub fn cross_compile_assistance(target_triple: impl AsRef<str>) -> CrossCompileA
     let (gcc_path, help_text) = match (target, consts::OS, consts::ARCH) {
         (AARCH64_UNKNOWN_LINUX_MUSL, OS_LINUX, ARCH_X86_64) => (
             "aarch64-linux-gnu-gcc",
-            "Install aarch64 cross-compiler:\nsudo apt-get install g++-aarch64-linux-gnu",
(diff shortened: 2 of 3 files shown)


libcnb-package/src/cross_compile.rs
```diff
@@ -111,5 +88,12 @@ pub enum CrossCompileAssistance {
     },
 }
 
+// Constants for supported target triples
 const AARCH64_UNKNOWN_LINUX_MUSL: &str = "aarch64-unknown-linux-musl";
 const X86_64_UNKNOWN_LINUX_MUSL: &str = "x86_64-unknown-linux-musl";
+
+// Constants for OS and ARCH
```
Malax, line 95: Nit:
```suggestion
// Constants for `std::env::consts::OS` and `std::env::consts::ARCH`
```

libcnb-package/src/cross_compile.rs
```diff
@@ -7,95 +8,71 @@ use which::which;
...
-        }
-    } else {
-        CrossCompileAssistance::NoAssistance
+{help_text}
+            
+You will also need to install the Rust target:
+rustup target add {target}",
+            consts::ARCH,
+            consts::OS
+        )),
```
Malax, line 75: This seems like a good use-case for [indoc](https://docs.rs/indoc/latest/indoc/index.html). We use that crate a lot and it helps making these multi-line strings much nicer in the code as it will strip away the common whitespace for your. And since it's a macro, there is no impact on the resulting code.

libcnb-package/src/cross_compile.rs
```diff
@@ -7,95 +8,71 @@ use which::which;
 /// look for the required tools and returns a human-readable help text if they can't be found or
 /// any other issue has been detected.
 pub fn cross_compile_assistance(target_triple: impl AsRef<str>) -> CrossCompileAssistance {
-    // Background: https://omarkhawaja.com/cross-compiling-rust-from-macos-to-linux/
-    if target_triple.as_ref() == X86_64_UNKNOWN_LINUX_MUSL && cfg!(target_os = "macos") {
-        // There is more than just one binary name here since we also support the binary name for
-        // an older version cross_compile_assistance which suggested installing a different
-        // toolchain.
-        let possible_gcc_binary_names = ["x86_64-unknown-linux-musl-gcc", "x86_64-linux-musl-gcc"];
+    let target = target_triple.as_ref();
```
Malax, line 11: Rust allows to re-declare variables with the same name. This would allow us to do: 

```suggestion
    let target_triple = target_triple.as_ref();
```

No need to call [`AsRef::as_ref`](https://doc.rust-lang.org/std/convert/trait.AsRef.html#tymethod.as_ref) everywhere but also no loss of information calling the target triple just `target` which can be ambiguous to readers.

libcnb-package/src/cross_compile.rs
```diff
@@ -7,95 +8,71 @@ use which::which;
...
+            "To install musl-tools on Ubuntu:\nsudo apt-get install musl-tools",
+        ),
+        (X86_64_UNKNOWN_LINUX_MUSL, OS_LINUX, ARCH_AARCH64) => (
+            "x86_64-linux-gnu-gcc",
+            "To install an x86_64 cross-compiler on Ubuntu:\nsudo apt-get install g++-x86_64-linux-gnu",
+        ),
+        (X86_64_UNKNOWN_LINUX_MUSL, OS_MACOS, _) => (
+            "x86_64-unknown-linux-musl-gcc",
+            "To install an x86_64 cross-compiler on macOS:\nbrew install messense/macos-cross-toolchains/x86_64-unknown-linux-musl",
+        ),
```
Malax, line 32: For these multiline help texts, using [indoc](https://docs.rs/indoc/latest/indoc/index.html) and writing them with actual newlines in the code should improve readability a little bit.

libcnb-package/src/cross_compile.rs
```diff
@@ -7,95 +8,71 @@ use which::which;
...
-
-You will also need to install the Rust target, using:
-rustup target add x86_64-unknown-linux-musl",
-            )), |gcc_binary_path| {
+    match which(gcc_path) {
+        Ok(_) => {
+            if gcc_path == "musl-gcc" {
+                CrossCompileAssistance::Configuration {
+                    cargo_env: Vec::new(),
+                }
```
Malax, line 41: Do we actually need this branch or could the code in the `else` branch work for `musl-gcc` as well? If we do need this branch, can we add a short comment why?

libcnb-package/src/cross_compile.rs
```diff
@@ -7,95 +8,71 @@ use which::which;
...
-        // an older version cross_compile_assistance which suggested installing a different
-        // toolchain.
-        let possible_gcc_binary_names = ["x86_64-unknown-linux-musl-gcc", "x86_64-linux-musl-gcc"];
+    let target = target_triple.as_ref();
+    let (gcc_path, help_text) = match (target, consts::OS, consts::ARCH) {
+        (AARCH64_UNKNOWN_LINUX_MUSL, OS_LINUX, ARCH_X86_64) => (
+            "aarch64-linux-gnu-gcc",
+            "To install an aarch64 cross-compiler on Ubuntu:\nsudo apt-get install g++-aarch64-linux-gnu",
+        ),
+        (AARCH64_UNKNOWN_LINUX_MUSL, OS_MACOS, _) => (
```
Malax, line 17: Since the `_` will match any value, wouldn't it be better to put `ARCH_X86_64 | ARCH_AARCH64` there to be explicit and to ensure we'll end up in the catch-all for unknown architectures?

(There is another `_` match later as well)

Malax (changes requested):
Thanks for the PR! 🎉 

I gave this a skim-read to provide some early feedback since this is one of your first PRs around here. I understand that this is a draft still.

Apart from the mostly minor code comments, I was missing some context due to the fact there is not PR description and the title being very vague (what does it fix, is there an existing issue somewhere?) Again, this is a draft PR and I write my descriptions at the very end as well. If that was to come later - ignore my comment! :)

There also seem to be some slight different in packages recommended. Have you verified those on the respective systems? For example, `libc6-dev-arm64-cross` is no longer mentioned. It's possible it is not needed, but from the information I have it could also be an oversight. :)

libcnb-package/src/cross_compile.rs
runesoerensen: For sure -- addressed in https://github.com/heroku/libcnb.rs/pull/769/commits/cab120b23e7965ef932119fbe813df3ce74e31f5

runesoerensen:
> Apart from the mostly minor code comments, I was missing some context due to the fact there is not PR description and the title being very vague (what does it fix, is there an existing issue somewhere?) Again, this is a draft PR and I write my descriptions at the very end as well. If that was to come later - ignore my comment! :)

I have added a more elaborate description now with links to relevant issues. Merging this pull request should close those issues, but let me know if it's better to unlink them and manually close the issues with reference to this pull request!

> There also seem to be some slight different in packages recommended. Have you verified those on the respective systems? For example, `libc6-dev-arm64-cross` is no longer mentioned. It's possible it is not needed, but from the information I have it could also be an oversight. :)

I initially removed the `libc6-dev-arm64-cross` package from the help text as during testing it was installed along with `g++-aarch64-linux-gnu` - but only as a recommended package - https://github.com/heroku/libcnb.rs/pull/769/commits/c5a046b0d7699e95351e3c25b03e5267130b9db7 adds it back in, and explains why. The following commit adds the equivalent packages for amd64 cross compilation from aarch64 Linux

I'm not sure if `musl-tools` is still necessary, but also added that to both help texts. I will test further, but overall it's probably best to keep the changes introduced by this pull request minimal (and improve on the help text in another PR).

Malax (changes requested):
Thanks for the changes! Looking great! 👍🏻 Let's add a brief `CHANGELOG` entry and ship it!

CHANGELOG.md
```diff
@@ -13,11 +13,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `libcnb`:
   - Changed `Layer` interface from `&self` to `&mut self`. ([#669](https://github.com/heroku/libcnb.rs/pull/669))
+- `libcnb-package`:
```
edmorley, line 16: As of #662, we treat `libcnb-package` (and `libcnb-common`) as an internal packages that has no stability guarantees. As such, we don't mention them directly in the changelog, but instead mention any end-user impact in terms of what the user would experience - which in this case is via `libcnb-cargo` and `libcnb-test`.

However, rather than duplicate the entries under both a `libcnb-cargo` bullet and a `libcnb-test` bullet, what we did for the last cross-compile assistance change was to mention it at the top level, like so:
https://github.com/heroku/libcnb.rs/blob/2907c182b4aaf72d90a312e95455e3b51c0399a8/CHANGELOG.md?plain=1#L36

Could you remove the `libcnb-package` bullet and move the entries to the top level?

edmorley (approved):
Thank you for this! The implementation is both cleaner and more correct now :-)

I tested this manually locally using on macOS by packaging the Python CNB (which includes C deps, such as `ring` to exercise the env vars being set), in the following scenarios:
- the default target (`x86_64-unknown-linux-musl`) with and without the correct toolchain installed
- `--target aarch64-unknown-linux-musl` with and without the correct toolchain installed

...and it worked great for all of those.

(Longer term we have #730 and #731 for exploring trying to test more scenarios in CI)

> However, the binary used when the architectures don't match is still the *-gnu-gcc variant (as for instance aarch64-linux-musl-gcc is only available on aarch64 Linux (this requires further investigation, and a separate issue should be created for that)).

The intention of #724 was to be about the general case of "we probably shouldn't be using the non-MUSL gcc ever". This PR helps with a subset of cases, but not all (as you mentioned in the PR description). Rather than close #724 out, I've morphed the title to make it clearer that it's about the general case :-)

libcnb-package/src/cross_compile.rs
```diff
@@ -7,95 +9,89 @@ use which::which;
...
 /// any other issue has been detected.
 pub fn cross_compile_assistance(target_triple: impl AsRef<str>) -> CrossCompileAssistance {
-    // Background: https://omarkhawaja.com/cross-compiling-rust-from-macos-to-linux/
-    if target_triple.as_ref() == X86_64_UNKNOWN_LINUX_MUSL && cfg!(target_os = "macos") {
-        // There is more than just one binary name here since we also support the binary name for
-        // an older version cross_compile_assistance which suggested installing a different
-        // toolchain.
-        let possible_gcc_binary_names = ["x86_64-unknown-linux-musl-gcc", "x86_64-linux-musl-gcc"];
+    let target_triple = target_triple.as_ref();
+    let (gcc_path, help_text) = match (target_triple, consts::OS, consts::ARCH) {
```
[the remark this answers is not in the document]
runesoerensen: Agree - forgot to make that change, but was thinking the same thing! Addressed in https://github.com/heroku/libcnb.rs/pull/769/commits/e040fee6ff05cce89cc83312f64eb98dc3a5ce0a

CHANGELOG.md
runesoerensen: Ah didn't know we made that distinction (I modeled the entries after [this entry](https://github.com/heroku/libcnb.rs/pull/577/commits/fc4f442b5d47f30c85a92b9af32aac9ccab12f02), which seems to have been the last one prior to marking the `libcnb-package` internal :)).

I [moved the entries up](https://github.com/heroku/libcnb.rs/pull/769/commits/d71b6c5c53cdea63953379969b87a2872bc0af77) to the top level (both under the `Changed` header) without the `libcnb-package` bullet.

CHANGELOG.md
```diff
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- Improved the consistency of cross-compilation assistance provided across all supported `target_triple` and host OS/architecture combinations. [#769](https://github.com/heroku/libcnb.rs/pull/769)
+- Added cross-compilation assistance for `aarch64-unknown-linux-musl` (on macOS and ARM64 Linux) and `x86_64-unknown-linux-musl` (on ARM64 Linux). [#769](https://github.com/heroku/libcnb.rs/pull/769)
+
```
edmorley, line 16:
```suggestion
```

runesoerensen: Lets pretend that never happened ;-) https://github.com/heroku/libcnb.rs/compare/d71b6c5c53cdea63953379969b87a2872bc0af77..f9fb1cfcc4030551941c860ad55004b44d90085b

PR #773, opened 2024-02-06
Move to buildpack API `0.10`

### LINKED ISSUE #772: Add support for buildpack.toml targets for Buildpack API 0.10
Buildpack API 0.10 has been released as of August 3, 2023: https://github.com/buildpacks/spec/releases/tag/buildpack%2Fv0.10.

It introduces the concept of buildpack `targets`, which replace the concepts `stacks` and `mixins`:
- https://github.com/buildpacks/spec/pull/336
- https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md

Malax:
### Stack and Mixin Removal ([RFC 096](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md), [Spec PR](https://github.com/buildpacks/spec/pull/336))

The `stack` field is sill in the spec as a deprecated field. I decided to not keep it around for libcnb. libcnb always just supported one specific version and when moving to the new buildpack API, uses should fully migrate to the `target` concept. This simplifies the implementation greatly (fewer permutations). If a blocker emerges, we can add support for deprecated stacks later down the line.

#### Open Questions

- [x] Are `target`s and `order` mutually exclusive as `stack` and `order` were? ([slack question](https://cloud-native.slack.com/archives/C033DV9EBDF/p1707401225376109))

This seems to be an oversight in the spec. `pack` does not allow for this ([code](https://github.com/buildpacks/pack/blob/5cbc93efc42e3aad4dd991d7a5437596aae9a4fd/pkg/buildpack/buildpack.go#L287-L295)) and it seems like a to big of a change to silently make. Semantics of this are also unclear. I added code comments to the respective areas to clarify why libcnb.rs differs from the spec. I'm in contact with the maintainers to fix this spec issue upstream.

### Runtime Base Image Extensions ([RFC 105](https://github.com/buildpacks/rfcs/blob/main/text/0105-dockerfiles.md), [Spec PR](https://github.com/buildpacks/spec/pull/353))

Will be implemented at a later time. The spec is marked as experimental and the feature is basically a new kind of buildpack that shares some of the spec for regular buildpacks, but with slight differences. See [the image extension spec](https://github.com/buildpacks/spec/blob/7eb38d15fd3bd5bb5240584f0e70beb61913032c/image_extension.md) for details. Tracking issue for this feature: #776

### Build Image Environment Variables ([RFC 109](https://github.com/buildpacks/rfcs/blob/main/text/0109-build-config.md), [Spec PR](https://github.com/buildpacks/spec/pull/349))

This does not affect the implementation in libcnb.rs. These new environment variables will always be set (regardless of the `clear-env` value and cannot be read manually from a directory. This is different from `$CNB_PLATFORM_DIR/env/`. However, this might affect buildpacks in other ways. For example, If a buildpack choses to set `clear-env` to `true` to ensure that no Maven specific environment variables are set that might mess with buildpack execution, build image environment variables can now mess with this assumption. 
 
However, there are many other ways environment variables can end up in the buildpack's executable environment (from outside of `$CNB_PLATFORM_DIR/env/`). Buildpack authors should not rely on `clear-env` to avoid clobbering of important environment variables in any case.

## Notes for reviewers
- A lot of `stack = "*"` occurrences in tests and docs have been removed. Having no `target` specified in `buildpack.toml` is equivalent as it matches any target.

Closes #772.


README.md
```diff
@@ -13,7 +13,7 @@
 It is an opinionated implementation adding language constructs and convenience methods for working
 with the spec. It values strong adherence to the spec and data formats.
 
-It currently targets version `0.9` of the CNB [Buildpack API specification](https://github.com/buildpacks/spec/blob/buildpack/0.9/buildpack.md).
+It currently targets version `0.10` of the CNB [Buildpack API specification](https://github.com/buildpacks/spec/blob/buildpack/0.9/buildpack.md).
```
runesoerensen, line 16:
```suggestion
It currently targets version `0.10` of the CNB [Buildpack API specification](https://github.com/buildpacks/spec/blob/buildpack/0.10/buildpack.md).
```

edmorley:
> This does not affect the implementation in libcnb.rs. These new environment variables will always be set (regardless of the `clear-env` value and cannot be read manually from a directory. This is different from `$CNB_PLATFORM_DIR/env/`.

Is this correct? Iirc there was a bit of uncertainty around this (inconsistency between the rfc, spec and implementation) - though it's been so long I can't quite remember the outcome. See:
https://github.com/buildpacks/spec/issues/330#issuecomment-1634336710
https://github.com/buildpacks/spec/pull/367

Malax:
> Is this correct?

The buildpack API spec `0.10` says:
> During the detection and build phases, the lifecycle MUST provide as environment variables any operator-provided files in \<build-config\>/env with environment variable names and contents matching the file names and contents. This applies for all values of clear-env or if clear-env is undefined in buildpack.toml.

Which is quite clear IMO. The `<build-config>` placeholder is only mentioned once (not defined, nor re-used) in the Buildpack API spec and there is no environment variable for it like for the other directories such as `CNB_PLATFORM_DIR` or `CNB_LAYERS_DIR`. I interpret this as the strong intent for the "operator defined environment variables" to be magic from the buildpack POV.

The [PR that implements this feature into lifecycle](https://github.com/buildpacks/lifecycle/pull/962) does not seem to make any attempt to expose build config environments separately. The current implementation in `main` isn't different in that regard, platform env and build config env are merged together: https://github.com/buildpacks/lifecycle/blob/751c7a2737be2e6a8f66bfcc2e0cf0058661d6ff/env/env.go#L81-L112

With the Buildpack API spec and lifecycle implementation agreeing, there doesn't seem to be any follow-up required for libcnb.rs here. If anyone has contradicting information I'm happy to stand corrected! :)

libcnb/src/runtime.rs
```diff
@@ -140,21 +140,38 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
...
-        .and_then(|stack_id_string| stack_id_string.parse().map_err(Error::StackIdError))
-        .inspect_err(|err| trace_error(err))?;
-
     let platform = B::Platform::from_path(&args.platform_dir_path)
         .map_err(Error::CannotCreatePlatformFromPath)
         .inspect_err(|err| trace_error(err))?;
 
     let build_plan_path = args.build_plan_path;
 
+    let target_os = env::var("CNB_TARGET_OS")
```
edmorley, line 149: I'm not sure I quite follow the spec about these env vars:
https://github.com/buildpacks/spec/blob/main/buildpack.md#targets

It says:

> The following environment variables MUST be set by the lifecycle during the detect and build phases to describe the target runtime image, if inputs are provided.

However:
1. What does "if inputs are provided" mean? Does it mean "some of these might not be set if they are not set in the builder image"? (In which case, do all of these need to be `Option<String>`? Also, are any of these automatically derived from the builder image, even if they are not set as labels there? eg there is mention of [this bionic backwards compat thing](https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml-stacks-array), but it presumably only matches the "bionic" case exactly, rather than any other builder IDs?
2. If the spec says MUST why does it then say "optional" for some of the env vars, like `CNB_TARGET_ARCH_VARIANT`

I guess I'm not clear how the tests are passing, given that our build images don't set the new labels yet, and only the legacy `io.buildpacks.stack.id` label and `CNB_STACK_ID` env var?
https://github.com/heroku/base-images/blob/0874d827bb161c4df57580840010063122e01130/heroku-22-cnb-build/Dockerfile#L12-L13

I also tried reading https://github.com/buildpacks/spec/blob/main/platform.md#target-data but that didn't really help either.

edmorley (commented):
Thank you for tackling this!

README.md
```diff
@@ -13,7 +13,7 @@
 It is an opinionated implementation adding language constructs and convenience methods for working
 with the spec. It values strong adherence to the spec and data formats.
 
-It currently targets version `0.9` of the CNB [Buildpack API specification](https://github.com/buildpacks/spec/blob/buildpack/0.9/buildpack.md).
+It currently targets version `0.10` of the CNB [Buildpack API specification](https://github.com/buildpacks/spec/blob/buildpack/0.10/buildpack.md).
```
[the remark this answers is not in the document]
Malax: I'd even say they **MUST**. :) Thanks!

libcnb/src/runtime.rs
edmorley: > Are any of these automatically derived from the builder image, even if they are not set as labels there? eg there is mention of [this bionic backwards compat thing](https://github.com/buildpacks/spec/blob/main/buildpack.md#buildpacktoml-toml-stacks-array), but it presumably only matches the "bionic" case exactly, rather than any other builder IDs?

I also found this backwards compat matrix (which was linked from one of the RFC or spec issues, I forget where) that suggests maybe the env vars are derived from the "image config" after all?
https://user-images.githubusercontent.com/6894483/222824654-09bd9c3f-26a8-44a1-90c6-f78d37bc4213.png

But does "image config" mean the manually provided labels in the `Dockerfile`, or the OCI metadata that automatically gets set when `docker build`ing the build image?

libcnb/src/build.rs
```diff
@@ -16,7 +15,11 @@ pub struct BuildContext<B: Buildpack + ?Sized> {
     pub layers_dir: PathBuf,
     pub app_dir: PathBuf,
     pub buildpack_dir: PathBuf,
-    pub stack_id: StackId,
+    pub target_os: String,
+    pub target_arch: String,
+    pub target_arch_variant: Option<String>,
+    pub target_distro_name: Option<String>,
+    pub target_distro_version: Option<String>,
```
edmorley, line 22: I think some rustdocs for these fields (and the equivalents in `DetectContext`) would help buildpack maintainers with the transition from `stack_id` to them. I think we should include:
- a brief description of the field (could be taken from the spec)
- an example value
- where the value originates from in practice (eg is it from the metadata set in the build image, derived, or ..?)

libcnb/src/runtime.rs
Malax: My understanding is that the `CNB_TARGET_*` environment variables will be filled from various sources if not explicitly set. Examples:

- [From systemd](https://github.com/buildpacks/lifecycle/blob/751c7a2737be2e6a8f66bfcc2e0cf0058661d6ff/platform/target_data.go#L60-L73)
- [From the run image metadata](https://github.com/buildpacks/lifecycle/blob/751c7a2737be2e6a8f66bfcc2e0cf0058661d6ff/platform/target_data.go#L27-L56)

Malax: Agreed, the spec is very unclear there. Lifecycle also sets `CNB_TARGET_DISTRO_NAME` to an empty string instead of not setting it at all when no name can be found. Unclear if that is intentional or not. Since empty strings in Go are are treated as Nil, this might a Go (yet again) leaking into the spec.

libcnb-data/src/buildpack/mod.rs
```diff
@@ -461,22 +486,26 @@ checksum = "abc123"
...
 
 [buildpack]
 id = "foo/bar"
 version = "0.0.1"
 
-[[stacks]]
-id = "*"
+[[targets]]
+os = "linux"
+arch = "amd64"
```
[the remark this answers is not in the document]
Malax: It could be, I added this in since the test previously also asserted against the `buildpack_descriptor.targets` field. We can change this to asserting if that field is empty. Works for me either way.

libcnb-data/src/buildpack/target.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub arch: Option<String>,
+    pub variant: Option<String>,
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    #[serde(default)]
+    pub distros: Vec<Distro>,
+}
+
+#[derive(Debug, Eq, PartialEq, Clone, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct Distro {
```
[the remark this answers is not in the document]
Malax: There is another bit in the spec that made me interpret this as "If you specify distros, then all keys under it are mandatory":

> For Windows-based images, distros.name SHOULD be empty; distros.versions SHOULD contain the value of os.version in the image config (e.g., 10.0.14393.1066)

Note that is says "empty" and not "omitted".

README.md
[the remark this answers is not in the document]
Malax: d1ac93c

README.md
```diff
@@ -148,7 +145,7 @@ impl Buildpack for HelloWorldBuildpack {
     // build phase (`bin/build`).
     fn build(&self, context: BuildContext<Self>) -> libcnb::Result<BuildResult, Self::Error> {
         println!("Hello World!");
-        println!("Build runs on stack {}!", context.stack_id);
+        println!("Build runs on: {} ({})!", context.target_os, context.target_arch);
```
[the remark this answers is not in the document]
Malax: Deferring to the native speaker, fbe3a90

libcnb-data/src/buildpack/mod.rs
```diff
@@ -102,22 +97,35 @@ impl<BM> BuildpackDescriptor<BM> {
...
+///     }]
+/// );
 /// ```
 #[derive(Deserialize, Debug)]
 #[serde(deny_unknown_fields)]
 pub struct ComponentBuildpackDescriptor<BM = GenericMetadata> {
     pub api: BuildpackApi,
     pub buildpack: Buildpack,
-    pub stacks: Vec<Stack>,
+    #[serde(default = "Vec::new")]
```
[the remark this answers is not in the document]
Malax: Good catch!

libcnb-data/src/buildpack/mod.rs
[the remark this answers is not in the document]
Malax: 1f9d849

libcnb/src/runtime.rs
Malax: Based on the fact that this works in practice with `CNB_TARGET_OS` and `CNB_TARGET_ARCH` being required and lifecycle actively setting these values even if not provided, I think we can continue with the PR as-is and have to raise these questions to upstream separately.

edmorley: > I think we can continue with the PR as-is and have to raise these questions to upstream separately.

Yeah agree it needn't block here - after this merges it would be good to make sure we track (via a TODO or GUS WI or GH issue in this repo) raising these upstream sometime soon before it fades too much from our memories :-)

Malax: Relevant spec Slack thread with upstream: https://cloud-native.slack.com/archives/C033DV9EBDF/p1707493563612439

libcnb/src/build.rs
Malax: Added in [dfdd601](https://github.com/heroku/libcnb.rs/pull/773/commits/dfdd60124aa67fdb2376e55f2a12732b9b29a5c6).

When adding these doc strings, the amount of duplication for detect and build became too large and I moved all target information into a new `ContextTarget` struct and added a shared initialiser in [0104524](https://github.com/heroku/libcnb.rs/pull/773/commits/0104524b6a54d3f535bba46efdf0953276217a91).

libcnb-data/src/buildpack/mod.rs
```diff
@@ -245,21 +257,21 @@ uri = "https://example.tld/my-license"
 [[buildpack.licenses]]
 uri = "https://example.tld/my-license"
 
-[[stacks]]
-id = "heroku-20"
-
-[[stacks]]
-id = "io.buildpacks.stacks.bionic"
-mixins = []
+[[targets]]
```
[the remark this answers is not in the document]
Malax: Added via 6a82f93, empty distros should not be valid, see comment https://github.com/heroku/libcnb.rs/pull/773#discussion_r1484345493.

examples/basics/src/main.rs
```diff
@@ -15,7 +15,11 @@ impl Buildpack for BasicBuildpack {
     }
 
     fn build(&self, context: BuildContext<Self>) -> libcnb::Result<BuildResult, Self::Error> {
-        println!("Build runs on stack {}!", context.stack_id);
+        println!(
+            "Build runs on: {} ({})!",
```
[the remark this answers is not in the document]
Malax: f13069d

libcnb/src/target.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// The variant of the specified CPU architecture.
+    ///
+    /// The value should conform to [OCI image spec platform variants](https://github.com/opencontainers/image-spec/blob/main/image-index.md#platform-variants), for example
+    /// `v7` or `v8`.
+    ///
+    /// CNB `lifecycle` sources this value from the build OCI image's [`variant` property](https://github.com/opencontainers/image-spec/blob/main/config.md#properties).
+    pub arch_variant: Option<String>,
+    /// The name of the operating system distribution. Should be empty for windows.
+    ///
+    /// For example: `ubuntu` or `arch`.
```
edmorley, line 26: (Optional) I found the usage of "arch" in this example confusing for a split second, since I read it as not an example, but referring the the `arch` field above. Whilst I then realised it meant the Arch distro, perhaps using another distro name here would be clearer?

Also, I wasn't sure what the correct capitalisation would be for these fields (sadly it's another thing the spec doesn't advise on - perhaps we should file this upstream?).

I presume lowercase is recommended, since I found lowercase for a slightly similar opencontainer field that some images already set:

```
$ docker inspect ubuntu | rg opencontainers
                "org.opencontainers.image.ref.name": "ubuntu",
                "org.opencontainers.image.version": "22.04"
```

Plus the OS release metadata uses lowercase for the `ID` field for several OSes:

```
$ dr ubuntu sh -c "cat /etc/*-release | sort | grep -v https"
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
ID=ubuntu
ID_LIKE=debian
NAME="Ubuntu"
PRETTY_NAME="Ubuntu 22.04.3 LTS"
UBUNTU_CODENAME=jammy
VERSION="22.04.3 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
VERSION_ID="22.04"
```

```
$ dr debian sh -c "cat /etc/*-release | sort | grep -v https"
ID=debian
NAME="Debian GNU/Linux"
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
VERSION_ID="12"
```

```
$ dr alpine sh -c "cat /etc/*-release | sort | grep -v https"
3.19.1
ID=alpine
NAME="Alpine Linux"
PRETTY_NAME="Alpine Linux v3.19"
VERSION_ID=3.19.1
```

libcnb/src/target.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// ``
+    pub arch: String,
+    /// The variant of the specified CPU architecture.
+    ///
+    /// The value should conform to [OCI image spec platform variants](https://github.com/opencontainers/image-spec/blob/main/image-index.md#platform-variants), for example
+    /// `v7` or `v8`.
+    ///
+    /// CNB `lifecycle` sources this value from the build OCI image's [`variant` property](https://github.com/opencontainers/image-spec/blob/main/config.md#properties).
+    pub arch_variant: Option<String>,
+    /// The name of the operating system distribution. Should be empty for windows.
```
edmorley, line 24:
```suggestion
    /// The name of the operating system distribution. Should be empty for Windows.
```

libcnb/src/build.rs
edmorley: Thank you - these new rustdocs are great :-)

edmorley (approved):
🎉

libcnb/src/target.rs
Malax: Switched to `alpine` in [c5cb54c](https://github.com/heroku/libcnb.rs/pull/773/commits/c5cb54ce9285983331e3afb08accbb45e970c542).

libcnb/src/target.rs
Malax: Fixed in [c5cb54c](https://github.com/heroku/libcnb.rs/pull/773/commits/c5cb54ce9285983331e3afb08accbb45e970c542)

PR #783, opened 2024-02-14
Add missing `derive(Debug)` for `buildpack_output::state::Stream`

schneems:


edmorley (approved):
Good spot, thank you!

PR #784, opened 2024-02-14
Add background dot printer

schneems:
The `start_timer` emits dots in the background while the buildpack does work. This is important for things like downloading a file while on spotty internet so that the user knows the buildpack isn't stuck.

### DIFF
--- a/libherokubuildpack/src/buildpack_output/background_printer.rs
+++ b/libherokubuildpack/src/buildpack_output/background_printer.rs
@@ -1,11 +1,11 @@
 //! This module is responsible for the logic involved in the printing to output while
-//! other work is being performed. Such as printing dots while a download is being performed
+//! other work is being performed. Such as printing dots while a download is being performed.
 use std::io::Write;
 use std::sync::mpsc::{channel, Sender};
 use std::thread::JoinHandle;
 use std::time::Duration;
 
-/// Print dots to the given buffer at the given interval
+/// Repeatedly prints `tick` to the given buffer at the given interval. The `start` argument will be printed before the first `tick` and the `end` argument will be printed after the last `tick` when the timer is stopped.
 ///
 /// Returns a struct that allows for manually stopping the timer or will automatically stop
 /// the timer if the guard is dropped. This functionality allows for errors that trigger
@@ -46,7 +46,7 @@ where
     PrintGuard::new(join_handle, sender)
 }
 
-/// Holds the reference to the background printer
+/// Holds the reference to the background printer.
 ///
 /// Ensures that the dot printer is stopped in the event of an error. By signaling
 /// it and joining when this struct is dropped.
@@ -60,7 +60,7 @@ where
 /// documentation in `PrintGuard::stop` below for more details.
 #[derive(Debug)]
 pub(crate) struct PrintGuard<W> {
-    /// Holds the handle to the thread printing dots
+    /// Holds the handle to the thread printing ticks in the background.
     ///
     /// Structs that implement `Drop` must ensure a valid internal state at
     /// all times due to E0509. The handle is wrapped in an option to allow the
@@ -101,7 +101,7 @@ impl<W> PrintGuard<W> {
         guard
     }
 
-    /// The only thing a consumer can do is stop the dots and receive
+    /// The only thing a consumer can do is stop the background printer and receive
     /// the original buffer.
     ///
     /// # Panics
@@ -111,10 +111,10 @@ impl<W> PrintGuard<W> {
     /// To avoid a panic, developers modifying this file must obey the following
     /// rules:
     ///
-    /// - Always consume the struct when accessing the Option value outside of Drop
+    /// - Always consume the struct when accessing the Option value outside of Drop.
     /// - Never construct this struct with a `None` option value.
     ///
-    /// This `Option` wrapping is needed to support a implementing Drop to
+    /// This `Option` wrapping is needed to support implementing Drop to
     /// ensure the printing is stopped. When a struct implements drop it cannot
     /// remove it's internal state due to E0509:
     ///
--- a/libherokubuildpack/src/buildpack_output/background_printer.rs
+++ b/libherokubuildpack/src/buildpack_output/background_printer.rs
@@ -1,0 +1,195 @@
+//! This module is responsible for the logic involved in the printing to output while
+//! other work is being performed. Such as printing dots while a download is being performed
+use std::io::Write;
+use std::sync::mpsc::{channel, Sender};
+use std::thread::JoinHandle;
+use std::time::Duration;
+
+/// Print dots to the given buffer at the given interval
+///
+/// Returns a struct that allows for manually stopping the timer or will automatically stop
+/// the timer if the guard is dropped. This functionality allows for errors that trigger
+/// an exit of the function to not accidentally have a timer printing in the background
+/// forever.
+#[must_use]
+pub(crate) fn print_interval<W>(
+    mut buffer: W,
+    interval: Duration,
+    start: String,
+    tick: String,
+    end: String,
+) -> PrintGuard<W>
+where
+    W: Write + Send + 'static,
+{
+    let (sender, receiver) = channel::<()>();
+
+    let join_handle = std::thread::spawn(move || {
+        write!(buffer, "{start}").expect("Writer should not be closed");
+        buffer.flush().expect("Writer should not be closed");
+
+        loop {
+            write!(buffer, "{tick}").expect("Writer should not be closed");
+            buffer.flush().expect("Writer should not be closed");
+
+            if receiver.recv_timeout(interval).is_ok() {
+                break;
+            }
+        }
+
+        write!(buffer, "{end}").expect("Writer should not be closed");
+        buffer.flush().expect("Writer should not be closed");
+
+        buffer
+    });
+
+    PrintGuard::new(join_handle, sender)
+}
+
+/// Holds the reference to the background printer
+///
+/// Ensures that the dot printer is stopped in the event of an error. By signaling
+/// it and joining when this struct is dropped.
+///
+/// Gives access to the original io/buffer struct passed to the background writer
+/// when the thread is manually stopped.
+///
+/// # Panics
+///
+/// Updates to this code need to take care to not introduce a panic. See
+/// documentation in `PrintGuard::stop` below for more details.
+#[derive(Debug)]
+pub(crate) struct PrintGuard<W> {
+    /// Holds the handle to the thread printing dots
+    ///
+    /// Structs that implement `Drop` must ensure a valid internal state at
+    /// all times due to E0509. The handle is wrapped in an option to allow the
+    /// inner value to be removed while preserving internal state.
+    join_handle: Option<JoinHandle<W>>,
+
+    /// Holds the signaling method to tell the background printer
+    /// to stop emitting.
+    stop_signal: Sender<()>,
+}
+
+impl<W> Drop for PrintGuard<W> {
+    fn drop(&mut self) {
+        // A note on correctness. It might seem that it's enough to signal the thread to
+        // stop, that we don't also have to join and wait for it to finish, but that's not
+        // the case. The printer can still emit a value after it's told to stop.
+        //
+        // When that happens the output can appear in the middle of another output, such
+        // as an error message if a global writer is being used such as stdout.
+        // As a result we have to signal AND ensure the thread is stopped before
+        // continuing.
+        if let Some(join_handle) = self.join_handle.take() {
+            let _ = self.stop_signal.send(());
+            let _ = join_handle.join();
+        }
+    }
+}
+
+impl<W> PrintGuard<W> {
+    /// Preserve internal state by ensuring the `Option` is always populated
+    fn new(join_handle: JoinHandle<W>, sender: Sender<()>) -> Self {
+        let guard = PrintGuard {
+            join_handle: Some(join_handle),
+            stop_signal: sender,
+        };
+        debug_assert!(guard.join_handle.is_some());
+
+        guard
+    }
+
+    /// The only thing a consumer can do is stop the dots and receive
+    /// the original buffer.
+    ///
+    /// # Panics
+    ///
+    /// This code can panic if it encounters an unexpected internal state.
+    /// If that happens it means there is an internal bug in this logic.
+    /// To avoid a panic, developers modifying this file must obey the following
+    /// rules:
+    ///
+    /// - Always consume the struct when accessing the Option value outside of Drop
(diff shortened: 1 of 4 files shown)


libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -491,6 +524,40 @@ where
...
+    ///
+    /// If a buildpack author wishes to start a long-running task that does not stream, starting a timer
+    /// will let the user know that the buildpack is performing work and that the UI is not stuck.
+    ///
+    /// One common use case is when downloading a file. Emitting periodic output when downloading is especially important for the local
+    /// buildpack development experience where the user's network may be unexpectedly slow, such as
+    /// in a hotel or on a plane.
+    ///
+    /// This function will transition your buildpack output to [`state::Background`].
+    #[allow(clippy::missing_panics_doc)]
```
schneems, line 537: This is because of manually writing and flushing the IO that appends to the same line. We need to do that here because `writeln_now` writes...lines. I originally had two functions `write_now` and `writeln_now` and it was a tad hard to read and know which case to expect.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -491,6 +524,40 @@ where
...
+        BuildpackOutput {
+            started: self.started,
+            state: state::Background {
+                started: Instant::now(),
+                write: background_printer::print_interval(
+                    self.state.write,
+                    std::time::Duration::from_secs(1),
+                    ansi_escape::wrap_ansi_escape_each_line(&ANSI::Dim, " ."),
+                    ansi_escape::wrap_ansi_escape_each_line(&ANSI::Dim, "."),
+                    ansi_escape::wrap_ansi_escape_each_line(&ANSI::Dim, ". "),
```
schneems, line 555: For dimming the dots:

The downside of this approach is that each character is wrapped in color codes. The upside is we never have to worry about accidentally getting in a state where the output is dimmed, and a panic happens, and the RESET is never written by accident.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -560,6 +651,38 @@ mod test {
...
+
+        assert_eq!(
+            expected,
+            strip_ansi_escape_sequences(String::from_utf8_lossy(&io))
+        );
+
+        // Test timer dot colorization
+        let expected = formatdoc! {"
+            - Background
+              - Installing\u{1b}[2;1m .\u{1b}[0m\u{1b}[2;1m.\u{1b}[0m\u{1b}[2;1m. \u{1b}[0m(< 0.1s)
```
schneems, line 679: I agree that since the logic lives here (to specify what to dim) it should be tested here. I still want to make sure the output looks okay as human readable, so I'm testing once with and once without color codes.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -560,6 +651,38 @@ mod test {
...
+            - Done (finished in < 0.1s)
+        "};
+
+        assert_eq!(expected, String::from_utf8_lossy(&io));
+    }
+
```
schneems, line 685: Intentionally not adding a changelog entry as the buildpack output is not (yet) shipped.

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
+//! This module is responsible for the logic involved in the printing to output while
```
schneems, line 1: Review order comment: Everything in this file is internal to the crate. It's about half comments.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -502,6 +569,30 @@ where
...
+    ///
+    /// Once you're finished with your long running task, calling this function
+    /// finalizes the timer's output and transitions back to a [`state::Section`].
+    #[must_use]
+    pub fn finish(self) -> BuildpackOutput<state::Section<W>> {
+        let duration = self.state.started.elapsed();
+        let mut io = match self.state.write.stop() {
+            Ok(io) => io,
+            Err(e) => std::panic::resume_unwind(e),
+        };
```
schneems, line 586: This match and resuming of the panic was suggested by clippy for handling a thread result. The alternative would be manually panicking, which...either way, is a panic.

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -502,6 +569,30 @@ where
...
+
+        writeln_now(&mut io, style::details(duration_format::human(&duration)));
```
schneems, line 588: This is where the newline is added at the end. Everything up to this point prints inline.

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
...
+#[must_use]
+pub(crate) fn print_interval<W>(
+    mut buffer: W,
+    interval: Duration,
+    start: String,
+    tick: String,
+    end: String,
+) -> PrintGuard<W>
+where
+    W: Write + Send + 'static,
```
schneems, line 23: FWIW the `'static` lifetime requirement really tripped me up. It doesn't mean "lives statically for as long as the program" it means: "Contains only owned values, OR is a staticly declared value." For our use case it means we cannot pass a reference to a `Write` we must always pass the full struct.

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
...
+
+            if receiver.recv_timeout(interval).is_ok() {
+                break;
+            }
+        }
+
+        write!(buffer, "{end}").expect("Writer should not be closed");
+        buffer.flush().expect("Writer should not be closed");
+
+        buffer
```
schneems, line 43: Once the thread is joined it returns the `Write` struct we gave it so we can keep using it.

edmorley:
Thank you for the PR. Just to set expectations timing wise - I won't have a chance to look at this PR until next week, since there are several things in my queue (both review-wise, and the Python `WEB_CONCURRENCY` work) ahead of it.

libherokubuildpack/src/buildpack_output/mod.rs
colincasey: I think the upside is worth it.

libherokubuildpack/src/buildpack_output/mod.rs
colincasey: Why not use `let mut io = self.state.write.stop().expect("...");` instead and provide a message stating that the internal restrictions of the `PrintGuard` have been violated?

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
+//! This module is responsible for the logic involved in the printing to output while
+//! other work is being performed. Such as printing dots while a download is being performed
+use std::io::Write;
+use std::sync::mpsc::{channel, Sender};
+use std::thread::JoinHandle;
+use std::time::Duration;
+
+/// Print dots to the given buffer at the given interval
```
colincasey, line 8: Printing dots is given as an example of how this can be used but, for this method, there are no dots. 

```suggestion
/// Repeatedly prints `tick` to the given buffer at the given interval. The `start` argument will be printed before the first `tick` and the `end` argument will be printed after the last `tick` when the timer is stopped.
```

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
...
+/// Gives access to the original io/buffer struct passed to the background writer
+/// when the thread is manually stopped.
+///
+/// # Panics
+///
+/// Updates to this code need to take care to not introduce a panic. See
+/// documentation in `PrintGuard::stop` below for more details.
+#[derive(Debug)]
+pub(crate) struct PrintGuard<W> {
+    /// Holds the handle to the thread printing dots
```
colincasey, line 63:
```suggestion
    /// Holds the handle to the thread printing ticks in the background.
```

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
...
+        let guard = PrintGuard {
+            join_handle: Some(join_handle),
+            stop_signal: sender,
+        };
+        debug_assert!(guard.join_handle.is_some());
+
+        guard
+    }
+
+    /// The only thing a consumer can do is stop the dots and receive
```
colincasey, line 104:
```suggestion
    /// The only thing a consumer can do is stop the background printer and receive
```

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
+//! This module is responsible for the logic involved in the printing to output while
+//! other work is being performed. Such as printing dots while a download is being performed
```
colincasey, line 2:
```suggestion
//! other work is being performed. Such as printing dots while a download is being performed.
```

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
...
+        write!(buffer, "{end}").expect("Writer should not be closed");
+        buffer.flush().expect("Writer should not be closed");
+
+        buffer
+    });
+
+    PrintGuard::new(join_handle, sender)
+}
+
+/// Holds the reference to the background printer
```
colincasey, line 49:
```suggestion
/// Holds the reference to the background printer.
```

libherokubuildpack/src/buildpack_output/background_printer.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// the original buffer.
+    ///
+    /// # Panics
+    ///
+    /// This code can panic if it encounters an unexpected internal state.
+    /// If that happens it means there is an internal bug in this logic.
+    /// To avoid a panic, developers modifying this file must obey the following
+    /// rules:
+    ///
+    /// - Always consume the struct when accessing the Option value outside of Drop
```
colincasey, line 114:
```suggestion
    /// - Always consume the struct when accessing the Option value outside of Drop.
```

libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -502,6 +569,30 @@ where
     }
 }
 
+impl<W> BuildpackOutput<state::Background<W>>
+where
+    W: Write + Send + Sync + 'static,
+{
+    /// Finalize a timer's output
```
colincasey, line 576:
```suggestion
    /// Finalize a timer's output.
```

libherokubuildpack/src/buildpack_output/mod.rs
colincasey: Now that the buildpack output has shipped, a changelog entry is warranted here.

colincasey (changes requested):

libherokubuildpack/src/buildpack_output/mod.rs
schneems: The more I think about it the more I like the strong guarantee to not put our user's terminal in a "bad" state with an ANSI code.

libherokubuildpack/src/buildpack_output/mod.rs
schneems: I'm not 100% sure, but this is the recommendation from stdlib
https://doc.rust-lang.org/std/thread/type.Result.html

> The value contained in the Result::Err variant is the value the thread panicked with; that is, the argument the panic! macro was called with. Unlike with normal errors, this value doesn’t implement the [Error](https://doc.rust-lang.org/std/error/trait.Error.html) trait.
>
> Thus, a sensible way to handle a thread panic is to either:
>
> propagate the panic with [std::panic::resume_unwind](https://doc.rust-lang.org/std/panic/fn.resume_unwind.html)

If we add an `.expect()` then clippy complains that this function doesn't have a `# Panics` section, but with this code recommended by stdlib docs, it doesn't. I figured it was better to go with the recommended version that requires fewer docs or lint disables to make clippy happy.

If it's eyebrow raising we could add a comment linking to those docs.

schneems:
Updated with a changelog entry. 

I didn't add one as I wasn't expecting the buildpack output changes to be released in an unusable state. I wanted to avoid any conflicts to reduce the cost of maintaining a long-lived branch until it was being actively worked on.

We might want to pre-populate the `## [Unreleased]` section with added/changed/removed along with bullet points for each of the sub libraries when we release. It won't prevent changelog conflicts, but would make them faster and more intuitive to resolve.

libherokubuildpack/src/buildpack_output/mod.rs
colincasey: Maybe, for consistency, the same should apply to the following code in this PR?
https://github.com/heroku/libcnb.rs/blob/841817f5d2645cfefcc3f2e5b6aef174be0cac91/libherokubuildpack/src/buildpack_output/mod.rs#L538-L545

schneems: > Maybe, for consistency, the same should apply to the following code in this PR?

I went ahead and annotated my code with a comment from my prior musing:

> If it's eyebrow raising we could add a comment linking to those docs.

I think that's good enough. 

In this case we're handling a different class of error. This code is otherwise identical to `writeln_now` which writes and flushes, both can panic if the writer is closed. In that case I believe we want to add an annotation here otherwise the panic wouldn't have the information that it was caused while we were trying to write to the buildpack output.

Contrast this with the panic resume_unwind. In that case either the problem happened in a different thread and has one of two causes (if I understand correctly). 

- Our code panic-d and already added an expectation message https://github.com/heroku/libcnb.rs/pull/784/files#diff-bacb3b82540d939a8804ab0a93bb0cc89eb3c098f073a0036f77dcbe660f17ffR29
- Or some unknown system level event, like an OS bug, triggered a panic in the thread, in which case we don't want to annotate the error on join, we want to effectively continue it.

I'm not 100% sure of all the cases when that second one can be triggered, but even if we ignore it, i think it makes sense to propagate the panic in the background thread case, but write an `except` message here.

schneems:
I'm going to rebase this and then I want to merge it in. If there's a refactor needed or some changes needed, then I'll own it. If it's not up to standards or quality, then we need to figure out additional ways to scale out standard and quality measures.

libherokubuildpack/src/buildpack_output/mod.rs
schneems: I added a changelog

colincasey (approved):

schneems:
Think we can close in favor of using https://crates.io/crates/bullet_stream

PR #787, opened 2024-02-15
Allows buildpack output to stream commands

schneems:
The `start_stream` interface only returns a single writer. The Command expects two writers: stdout and stderr. There's currently no good way for a buildpack author to stream a running Command, to the point that there are no examples of it in the documentation.

This PR introduces a mechanism to do this and adds a doc that executes a command streaming to the build output.

### DIFF
--- a/libherokubuildpack/src/buildpack_output/mod.rs
+++ b/libherokubuildpack/src/buildpack_output/mod.rs
@@ -477,21 +477,27 @@ where
             started: self.started,
             state: state::Stream {
                 started: Instant::now(),
-                write: line_mapped(self.state.write, |mut line| {
-                    // Avoid adding trailing whitespace to the line, if there was none already.
-                    // The `[b'\n']` case is required since `line` includes the trailing newline byte.
-                    if line.is_empty() || line == [b'\n'] {
-                        line
-                    } else {
-                        let mut result: Vec<u8> = Self::CMD_INDENT.into();
-                        result.append(&mut line);
-                        result
-                    }
-                }),
+                write: Self::format_stream_writer(self.state.write),
             },
         }
     }
 
+    fn format_stream_writer<S>(stream_to: S) -> crate::write::MappedWrite<S>
+    where
+        S: Write + Send + Sync,
+    {
+        line_mapped(stream_to, |mut line| {
+            // Avoid adding trailing whitespace to the line, if there was none already.
+            // The `[b'\n']` case is required since `line` includes the trailing newline byte.
+            if line.is_empty() || line == [b'\n'] {
+                line
+            } else {
+                let mut result: Vec<u8> = Self::CMD_INDENT.into();
+                result.append(&mut line);
+                result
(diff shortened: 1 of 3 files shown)


libherokubuildpack/src/buildpack_output/mod.rs
```diff
@@ -35,6 +35,8 @@ use crate::buildpack_output::util::{prefix_first_rest_lines, prefix_lines, Parag
 use crate::write::line_mapped;
```
schneems, line 35: I'm ready for feedback on if this is a viable path forward or what else might be needed.

## Exploring our options for solving this problem

- Give the end user a struct like `LockedWriter` that allows them to clone and manage their own IO faux duplication
  - Technically possible and quite explicit, but is gross for the buildpack maintainer. I feel we should take on this complexity, not our end users.
- Introduce a Command output function variant that only takes a single writer
  - I tried this, the function would then have to return the writer and any Result object, likely via a tuple, which is ergonomically weird.
- Introduce a non-consuming function that yields two `Write` structs that are then proxied to the actual Write output (This PR).
  - This PR: Solves the problem. The downside is that it's complex, but it's hidden from the user. There may be an easier way to implement this logic. I feel this closure type of build output streaming is the correct interface as it allows natively interfacing with the result and not having to juggle returning a build output struct.

colincasey: fwiw, i've been pretty happy with the closure variant for streaming command outputs.

colincasey (approved):
@schneems I noticed you added the "Skip Changelog" label here. Why no changelog entry?

edmorley:
> Why no changelog entry?

Good spot. I suspect it's because when this PR was opened, the `build_output` module had not been released, so this PR would have been covered by the existing "add a new build_output module" changelog entry, and that Richard forgot to account for that after the release of libcnb 0.19.0.

schneems:
Integration tests are failing. They're also failing for me on main locally https://github.com/heroku/libcnb.rs/issues/802

edmorley:
> Integration tests are failing. They're also failing for me on main locally #802

These are two separate, albeit related things.

The failures in CI on this PR are because the PR needs rebasing on `main` to pick up #801 (which makes this repo compatible with Procfile CNB 3.0.0).

The failures locally are because of:
https://github.com/heroku/libcnb.rs/issues/802#issuecomment-1969897593

edmorley:
(Ignore the previous message, I thought the PR wasn't rebased on `main`)

Ah the `Unexpected Docker volumes left behind!` is an intermittent failure that happens occasionally - I don't know why (since image cleanup works most of the time). I'd like to track it down, but I probably won't have time to do so for a few months.

schneems:
Thankfully a restart fixed the issue. Changelog is added. Tests are passing.

schneems:
Think we can close in favor of using https://crates.io/crates/bullet_stream

PR #811, opened 2024-03-28
Correct `ContextTarget` references to distro image label names

edmorley:
Since the relevant distros run image Docker label names are `io.buildpacks.base.distro.name` and `io.buildpacks.base.distro.version`. 

See:
https://github.com/buildpacks/spec/blob/buildpack/v0.10/platform.md#target-data

I believe the rustdocs here were based on the `lifecycle` codebase, which uses the wrong label name, and is about to be fixed:
https://github.com/buildpacks/lifecycle/pull/1325


edmorley:
CI is failing because of new lint errors with Rust 1.77 - I've opened #812 to fix.

colincasey (approved):

PR #813, opened 2024-04-08
Test helpers to match against regular expressions.

colincasey:
Extracted these from https://github.com/heroku/buildpacks-apt/pull/8 which uses these macros to test for output that cannot be done reliably using just `assert_contains` and `assert_not_contains` because of the presence of random variations like the names of temporary test folders.

### DIFF
--- a/libcnb-test/Cargo.toml
+++ b/libcnb-test/Cargo.toml
@@ -20,7 +20,7 @@ fs_extra = "1.3.0"
 libcnb-common.workspace = true
 libcnb-data.workspace = true
 libcnb-package.workspace = true
-regex-lite = "0.1.5"
+regex = "1.10.4"
 tempfile = "3.10.1"
 thiserror = "1.0.58"
 
--- a/libcnb-test/Cargo.toml
+++ b/libcnb-test/Cargo.toml
@@ -20,6 +20,7 @@ fs_extra = "1.3.0"
 libcnb-common.workspace = true
 libcnb-data.workspace = true
 libcnb-package.workspace = true
+regex-lite = "0.1.5"
 tempfile = "3.10.0"
 thiserror = "1.0.57"
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,181 +9,263 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `libcnb-test`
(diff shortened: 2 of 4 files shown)


Malax:
@colincasey can you get CI passing? I gave it a quick skim and it looked fine otherwise. I'll give it a quick review as soon as CI is fixed. Can you ping me as soon as that is done? :)

colincasey:
@Malax yep, forgot to check a doc test and had accidentally applied a `#[cfg(test)]` to the unused dependency the linter was complaining about.

libcnb-test/Cargo.toml
```diff
@@ -20,6 +20,7 @@ fs_extra = "1.3.0"
 libcnb-common.workspace = true
 libcnb-data.workspace = true
 libcnb-package.workspace = true
+regex-lite = "0.1.5"
```
colincasey, line 23: Question - should we use this or the full `regex` crate?

schneems: I know that clippy will check the `regex` crate for valid regex-s and that it will not check `fancy-regex`. I don't know if it will check regex-lite. Looking at the source it does not look like it would https://github.com/rust-lang/rust-clippy/issues/10205. I think we should use the `regex` crate.

colincasey: @schneems because the regex is passed to the macro as a string value I don't believe the linter will be able to check those. And I'm less concerned about having linting here because these macros are intended for testing so if an invalid regex is given, you'll end up with a panic that fails a test.

colincasey: I'm happy to swap `regex-lite` for `regex` though. For test code I don't think we need to be as concerned with compile or binary sizes.

schneems (approved):
PR review:

I asked to swap out the regex backend in another comment. Thanks for doing that.

I've looked at the interface (contents come first, then regex, then optional error message). I've looked at the docs, both macros have doc tests. I've looked at the tests: it's tested. I'm not sure what else we might want to add, it seems thorough. I looked at the macros. It supports an optional trailing comma which I like. Both macros are pretty self-explanatory.

Overall I think it looks great. ✅

Side comment:

I like the pattern of linking to a regex-explorer resource when using regular expressions. Here's an example with Rubular (for ruby regexes) https://github.com/heroku/heroku-buildpack-ruby/blob/fd6585411cbf86d363ad0824f53d8edba1d7afed/lib/language_pack/helpers/binstub_wrapper.rb#L35. If the team finds one we like, maybe we can promote using that as a part of the optional 3rd argument.

PR #814, opened 2024-04-09
Struct Layer API

Malax:
This PR adds a new API for working with CNB layers, attempting to solve the issues with the current trait based API. The current issues are mainly around flexibility and explicitness.

This PR **does not** remove the existing trait API. This allows buildpack maintainers to slowly migrate to the new API, even on a layer-by-layer basis. Eventually, the old trait API will be removed. For now, the trait API is just marked as deprecated.

## Trait API problems

The current API encourages, and in some cases forces, buildpack implementors to put all code related to a layer into the `Layer` trait implementation. As the code will be called from libcnb itself, the buildpack has no control and/or insight when the functions will be called. This is especially a problem for logging. 

Passing data into and from a `Layer` trait implementation is hard to do correctly and ergonomically. Before #669, it was impossible to use mutable values originating from outside the layer implementation because the trait was too restrictive. Coupled with the fact that most buildpack implementations chose to put business logic in `Layer` implementations made this restriction even worse.

The trait default implementation of `existing_layer_strategy` is a source of great confusion because the default is not an universally good default. More context can be found in #371.

## New API philosophy

The new API only abstracts the minimum functionality and leaves code structure for dealing with the business logic up to the implementor. All a buildpack now does is asking `BuildContext` for a layer, providing two callbacks for cache invalidation and invalid metadata handling:

```rust
let layer_ref = context.cached_layer(
    layer_name!("test"),
    CachedLayerDefinition {
        build: true,
        launch: true,
        invalid_metadata: &|_| InvalidMetadataAction::DeleteLayer,
        inspect_restored: &|_: &GenericMetadata, _| InspectRestoredAction::KeepLayer,
    },
)?;

```

Afterwards, the `LayerRef` returned by `cached_layer` (there is also an even simpler `uncached_layer`) contains the path to the layer and detailed information about the layer state. The layer state can signal that the layer was restored from cache, that the layer is empty because the cache was invalidated by the `inspect_restored` callback, that the layer is empty because it didn't exist, etc. Both `invalid_metadata` and `inspect_restored` can also return a cause value that is available in the layer state. This is useful if the cache invalidation logic has multiple cases for invalidation and the buildpack needs to know why the cache was invalidated. This is specially useful for user output.

Fallible callback implementations are also possible by wrapping the return value of the callback in a `Result` with the buildpack error type as the `Result`'s `E`. In fact, any type is possible to be used as a callback return value as long as it implements `IntoAction`.  Most buildpacks will never have to do this though.

Since the callbacks are implemented as function references with a lifetime, it is possible to access values that are valid for the lifetime of the definition and to use regular functions that can be unit tested without the baggage of the actual layer handling:

```rust
fn inspect_restored_complex(metadata: &GenericMetadata, path: &Path) -> InspectRestoredAction {
    unimplemented!()
}

context.cached_layer(
    layer_name!("test"),
    CachedLayerDefinition {
        build: true,
        launch: true,
        invalid_metadata: &|_| InvalidMetadataAction::DeleteLayer,
        inspect_restored: &inspect_restored_complex,
    },
)?;
```

Since everything that comes after the `cached_layer` call is up to the buildpack author, the appropriate abstraction can be chosen on a case-by-case basis. A quick and dirty inline layer in `main.rs`:

```rust
let layer_ref = context.cached_layer(
    layer_name!("test"),
    CachedLayerDefinition {
        build: true,
        launch: true,
        invalid_metadata: &|_| InvalidMetadataAction::DeleteLayer,
        inspect_restored: &inspect_restored_complex,
    },
)?;

match layer_ref.state {
    LayerState::Restored { .. } => {
        // Work with existing data
    }
    LayerState::Empty { .. } => {
        // Populate the layer
    }
}
```

or a more complex layer with lots of associated business logic in its own module:

```rust
let jvm_layer_ref = context.cached_layer(
    layer_name!("jvm"),
    CachedLayerDefinition {
        build: true,
        launch: true,
        invalid_metadata: &|_| InvalidMetadataAction::DeleteLayer,
        inspect_restored: &crate::layers::jvm::inspect_restored_jvm::layer,
    },
)?;

crate::layers::jvm::handle_jvm_layer(&jvm_layer_ref, &openjdk_inventory, &mut logger)?;
```

Since `LayerState` is an easily pattern matched value, buildpack authors can freely choose which layer states to combine or even care about. None of that flexibility is achieved by libcnb.rs specific mechanisms - use your entire Rust toolbox and go wild!

The PR has more examples how this new API can be used in RustDocs, example and test buildpacks.

## Refactor of the trait API implementation

This PR also refactors the implementation of the trait API. No changes to the actual API has been made though. The refactor was necessary to allow for code sharing between both APIs as the low-level operations on CNB layers are the same. As the code for those low-level operations has been thoroughly tested, it made sense to reuse it for the new API implementation as well.

## Breaking Changes

This PR is marked as a breaking change. Even though the old trait API wasn't removed, the layer errors needed to be restructured to allow code-sharing between the old and new API. For the majority of the buildpack implementations, this will not be a breaking change in practice. The JVM buildpacks built cleanly with the code in this PR without any modifications.

Yet, as this is technically a breaking change, the PR is marked as such and the changelog contains a section of the breakage.

## Follow ups

- The examples and test buildpacks in this repository are still using the old trait API. This is intentional to keep this PR as small as possible. See #833 (stacked on top of this PR) for updated examples and test buildpacks.
- Functions on `LayerRef` always replace the entirety of a thing (SBOMs, env, exec.d). There are no functions to delete or add individual entries. Initial ports of existing JVM buildpack layers did not show the need for those functions. To keep things lightweight, this PR doesn't attempt to add them. We can add them later easily should the need arise.

## Review notes

This is a really dense (packed) PR. The code for the new API is highly generic and small details can make or break the usability of this API. It might not be obvious by just reading the implementation in a web browser how the pieces fit together and how the API is used. However, I am absolutely certain that it is easy to **use**.

Please take the time and look at the usage examples in `BuildContext::cached_layer` to get a feel for the API. I'd also suggest using this branch for an existing buildpack and port a layer over. This can be achieved by "patching" in `Cargo.toml`:

```
[patch.crates-io]
libcnb = { git = "https://github.com/heroku/libcnb.rs.git", branch = "malax/layer-api" }
libcnb-data = { git = "https://github.com/heroku/libcnb.rs.git", branch = "malax/layer-api" }
```

Should you have trouble getting started with the new API, please let me know - this is the most important feedback as I can myself no longer see this new API with fresh eyes!


runesoerensen (approved):
I've used the new API for the past several weeks without any issues. Good work!

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+{
+    /// Returns the path to the layer on disk.
+    pub fn path(&self) -> PathBuf {
+        self.layers_dir.join(self.name.as_str())
+    }
+
+    /// Replaces the existing layer metadata with a new value.
+    ///
+    /// The new value does not have to be of the same type as the existing metadata.
+    pub fn replace_metadata<M>(&self, metadata: M) -> crate::Result<(), B::Error>
```
edmorley, line 172: It feels a little strange to have functions named `replace_*` (such as `replace_metadata`, `replace_env`, `replace_sboms` etc) when the layer could be new and its not replacing anything?

What do you think about either no prefix at all, or something like `set_` as a prefix?

eg:

```rust
my_layer_ref.metadata(...);
my_layer_ref.env(...);
my_layer_ref.sboms(...);
my_layer_ref.exec_d_programs(...);
```

Or:

```rust
my_layer_ref.set_metadata(...);
my_layer_ref.set_env(...);
my_layer_ref.set_sboms(...);
my_layer_ref.set_exec_d_programs(...);
```

libcnb/src/build.rs
```diff
@@ -95,18 +104,297 @@ impl<B: Buildpack + ?Sized> BuildContext<B> {
...
+    ///         CachedLayerDefinition {
+    ///             build: false,
+    ///             launch: false,
+    ///             invalid_metadata: &|_| InvalidMetadataAction::DeleteLayer,
+    ///             inspect_restored: &|metadata: &ExampleLayerMetadata, layer_dir| {
+    ///                 if metadata.lang_runtime_version.starts_with("0.") {
+    ///                     Ok((
+    ///                         InspectRestoredAction::DeleteLayer,
+    ///                         CustomCause::LegacyVersion,
+    ///                     ))
```
edmorley, line 280: I found the tuple being returned here confusing, since I was expecting to see a `From` / `Into` implementation in the rustdocs here (based on the last time I'd seen the proposed design) but there wasn't one.

As I continued the review, I found the built-in `From` implementations for the common forms (such as returning a tuple) - however, these don't appear to be documented in the rustdocs anywhere.

I'm not sure where is best to put these? Perhaps the `CachedLayerDefinition` rustdocs, and then link to those from the rustdocs of `cached_layer` with a see more at X? Or vice versa.

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Whether the layer is intended for launch.
+    pub launch: bool,
+    /// Callback for when the metadata of a restored layer cannot be parsed as `M`.
+    ///
+    /// Allows replacing the metadata before continuing (i.e. migration to a newer version) or
+    /// deleting the layer.
+    pub invalid_metadata: &'a dyn Fn(&GenericMetadata) -> MA,
+    /// Callback when the layer was restored from cache to validate the contents and metadata.
+    /// Can be used to delete existing cached layers.
+    pub inspect_restored: &'a dyn Fn(&M, &Path) -> IA,
```
edmorley, line 34: I feel like there might be a clearer name for `inspect_restored` but I can't quite come up with one. Want to iterate on this and some of the other naming comments in a huddle?

edmorley (changes requested):
Awesome! Few comments but otherwise think we're very close to merging this :-)

libcnb/src/layer/struct_api/mod.rs
Malax: Slack call outcome: we decided to use a `write_` prefix.

libcnb/src/layer/struct_api/mod.rs
edmorley: Slack call outcome: We were leaning towards something like `invalid_metadata_action` and `restored_layer_action`

Malax: [f7bc0f1](https://github.com/heroku/libcnb.rs/pull/814/commits/f7bc0f1b540f5f3253c1f411a3c57bf561bd9129)

CHANGELOG.md
```diff
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- `libcnb`:
+  - A new API for working with layers as been added. Check `BuildContext::cached_layer` for a starting point how to work with this API. ([#814](https://github.com/heroku/libcnb.rs/pull/814))
```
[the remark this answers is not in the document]
Malax: Fixed.

libcnb/src/layer/struct_api/mod.rs
Malax: [06d4668](https://github.com/heroku/libcnb.rs/pull/814/commits/06d466876f9140a27e2eef25cd79726eb4e01223)

libcnb/src/layer/struct_api/handling.rs
```diff
@@ -0,0 +1,520 @@
...
+use libcnb_data::layer_content_metadata::{LayerContentMetadata, LayerTypes};
+use serde::de::DeserializeOwned;
+use serde::Serialize;
+use std::marker::PhantomData;
+use std::path::{Path, PathBuf};
+
+pub(crate) fn handle_layer<B, M, MA, RA, MAC, RAC>(
+    layer_types: LayerTypes,
+    invalid_metadata: &dyn Fn(&GenericMetadata) -> MA,
+    inspect_restored: &dyn Fn(&M, &Path) -> RA,
```
edmorley, line 22: These need renaming to `invalid_metadata_action` / `restored_layer_action` too (not using suggested changes feature since there are a few more places below that will need updating as a result).

libcnb/src/layer/struct_api/handling.rs
```diff
@@ -0,0 +1,520 @@
...
+            (),
+            (),
+        >(
+            LayerTypes {
+                build: true,
+                launch: true,
+                cache: true,
+            },
+            &|_| panic!("invalid_metadata callback should not be called!"),
+            &|_, _| panic!("inspect_restored callback should not be called!"),
```
edmorley, line 217:
```suggestion
            &|_| panic!("invalid_metadata_action callback should not be called!"),
            &|_, _| panic!("restored_layer_action callback should not be called!"),
```

libcnb/src/layer/struct_api/handling.rs
```diff
@@ -0,0 +1,520 @@
...
+            LayerTypes {
+                build: true,
+                launch: true,
+                cache: true,
+            },
+            &|metadata| {
+                assert_eq!(metadata, &Some(toml! { answer = 42 }));
+                (InvalidMetadataAction::DeleteLayer, DELETE_CAUSE)
+            },
+            &|_, _| panic!("inspect_restored callback should not be called!"),
```
edmorley, line 398:
```suggestion
            &|_, _| panic!("restored_layer_action callback should not be called!"),
```

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+    Uncached,
+    /// The layer was cached in a previous buildpack run, but the metadata was invalid and couldn't
+    /// be converted into a valid form. Subsequently, the layer was deleted entirely.
+    ///
+    /// See: `invalid_metadata` in [`CachedLayerDefinition`].
+    MetadataInvalid { cause: MAC },
+    /// The layer was cached in a previous buildpack run, but the `inspect_restored` function
+    /// rejected the contents and/or metadata.
+    ///
+    /// See: `inspect_restored` in [`CachedLayerDefinition`].
```
edmorley, line 92:
```suggestion
    /// The layer was cached in a previous buildpack run, but the `restored_layer_action` function
    /// rejected the contents and/or metadata.
    ///
    /// See: `restored_layer_action` in [`CachedLayerDefinition`].
```

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/// Framework metadata about the layer state.
+///
+/// See: [`BuildContext::cached_layer`] and [`BuildContext::uncached_layer`]
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum LayerState<MAC, RAC> {
+    /// The layer contains validated cached contents from a previous buildpack run.
+    ///
+    /// See: `inspect_restored` in [`CachedLayerDefinition`].
```
edmorley, line 73:
```suggestion
    /// See: `restored_layer_action` in [`CachedLayerDefinition`].
```

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+
+/// The cause of a layer being empty.
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum EmptyLayerCause<MAC, RAC> {
+    /// The layer wasn't cached in a previous buildpack run and was freshly created.
+    Uncached,
+    /// The layer was cached in a previous buildpack run, but the metadata was invalid and couldn't
+    /// be converted into a valid form. Subsequently, the layer was deleted entirely.
+    ///
+    /// See: `invalid_metadata` in [`CachedLayerDefinition`].
```
edmorley, line 87:
```suggestion
    /// See: `invalid_metadata_action` in [`CachedLayerDefinition`].
```

libcnb/src/build.rs
```diff
@@ -95,18 +104,297 @@ impl<B: Buildpack + ?Sized> BuildContext<B> {
...
+
+    /// Creates a cached layer, potentially re-using a previously cached version.
+    ///
+    /// Buildpack code uses this function to create a cached layer and will get back a reference to
+    /// the layer directory on disk. Intricacies of the CNB spec are automatically handled such as
+    /// the maintenance of TOML files. Buildpack code can also specify a callback for cached layer
+    /// invalidation.
+    ///
+    /// Users of this function pass in a [`CachedLayerDefinition`] that describes the desired layer
+    /// and the returned `LayerRef` can then be used to modify the layer like any other path. This
```
schneems, line 130: Make doc reference clickable

```suggestion
    /// and the returned [`LayerRef`] can then be used to modify the layer like any other path. This
```

Make doc clickable

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,228 @@
...
+{
+    name: LayerName,
+    // Technically not part of the layer itself. However, the functions that modify the layer
+    // will need a reference to the layers directory as they will also modify files outside the
+    // actual layer directory. To make LayerRef nice to use, we bite the bullet and include
+    // the layers_dir here.
+    layers_dir: PathBuf,
+    buildpack: PhantomData<B>,
+    pub state: LayerState<MAC, RAC>,
+}
```
schneems, line 158: Suggestion: Add a `fn read_env` to LayerRef.

Context:
With the current layer design I have a pattern of applying the env of the newly created layer to future layers https://github.com/heroku/buildpacks-ruby/blob/aaa72f317ee7b4700de35ea27112c1ad15334def/buildpacks/ruby/src/main.rs#L146. This is important to do even if the buildpack did not write any `LayerEnv` data because the `bin/` path of the layer is always put on the path.

Therefore, I'm wanting some mechanism of pulling out the current `LayerEnv` of a layer. This exists:

```
  LayerEnv::read_from_layer_dir(barnes_layer.path())
      .expect("read env from a layer")
      .apply(Scope::Build, &env),
```

However, the docs say not to use it, and it returns `std::io::Error` instead of a `libcnb::Error`. 


Sorry for tacking this somewhere random, but it seemed to make the most sense to put the request at the end of `LayerRef`.

runesoerensen: I was just about to ask the same question :-) I'm currently using `LayerEnv::read_from_layer_dir(layer_ref.path()`, which works great, but the warning not to use it should maybe be changed (if not an alternative implementation that works with the new API.

libcnb/src/build.rs
Malax: 👍🏻

libcnb/src/layer/struct_api/handling.rs
```diff
@@ -0,0 +1,522 @@
...
+            layers_dir,
+            EmptyLayerCause::Uncached,
+        ),
+        Ok(Some(layer_data)) => {
+            let inspect_action =
+                restored_layer_action_fn(&layer_data.metadata.metadata, &layer_data.path)
+                    .into_action()
+                    .map_err(crate::Error::BuildpackError)?;
+
+            match inspect_action {
```
edmorley, line 45:
```suggestion
            let restored_layer_action =
                restored_layer_action_fn(&layer_data.metadata.metadata, &layer_data.path)
                    .into_action()
                    .map_err(crate::Error::BuildpackError)?;

            match restored_layer_action {
```

libcnb/src/build.rs
Malax: Added some additional docs in [400b1b3](https://github.com/heroku/libcnb.rs/pull/814/commits/400b1b334181e5d2036a4aa13a586a464ce38ae9)

libcnb/src/build.rs
```diff
@@ -95,18 +104,297 @@ impl<B: Buildpack + ?Sized> BuildContext<B> {
...
+            &self.layers_dir,
+        )
+    }
+
+    /// Creates an uncached layer.
+    ///
+    /// If the layer already exists because it was cached in a previous buildpack run, the existing
+    /// data will be deleted.
+    ///
+    /// This function is essentially the same as [`BuildContext::uncached_layer`] but simpler.
```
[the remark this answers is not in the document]
Malax: Added an example in a4877e4

libcnb/src/layer/struct_api/mod.rs
```diff
@@ -0,0 +1,235 @@
...
+        env.borrow()
+            .write_to_layer_dir(self.path())
+            .map_err(|error| {
+                crate::Error::LayerError(LayerError::WriteLayerError(WriteLayerError::IoError(
+                    error,
+                )))
+            })
+    }
+
+    /// Reads the current layer environment from disk.
```
edmorley, line 198: Maybe add a note saying this will also add the additional magic entries?

edmorley (approved):

libcnb/src/layer/struct_api/mod.rs
Malax: Good call!

PR #821, opened 2024-04-30
Improve usability of `Target`

### LINKED ISSUE #820: Usability issues with the `Target` type
The `distro_name` and `distro_version` fields of the [libcnb::Target](https://github.com/heroku/libcnb.rs/blob/v0.20.0/libcnb/src/target.rs#L1-L35) type are currently optional (`Option<String>`). 

It was implemented this way because the Buildpack API spec says that the `CNB_TARGET_DISTRO_NAME` and `CNB_TARGET_DISTRO_VERSION` env vars are optional:
https://github.com/buildpacks/spec/blob/buildpack/0.11/buildpack.md#targets

However, in practice these env vars will always be set, since:
1. If the base image has the `io.buildpacks.base.distro.*` Docker labels set, then these values will be used (now that https://github.com/buildpacks/lifecycle/issues/1324 is fixed).
2. If the labels aren't set, then lifecycle falls back to the values found in `/etc/os-release` (once the fix for https://github.com/buildpacks/lifecycle/issues/1336 is released shortly), and `/etc/os-release` exists for pretty much all of the common Linux distros (xref https://unix.stackexchange.com/q/351557).
3. Even if neither the labels nor `/etc/os-release` is found, `lifecycle` still [sets both of these env vars to the empty string](https://github.com/buildpacks/lifecycle/blob/13d3691dd017405e22ad7b5050204cc2ccdf034a/platform/target_data.go#L30-L31) anyway, so the env vars will never be unset.

Plus, for buildpacks that include the distro name/version in their `buildpack.toml` `[[targets.distros]]`, the buildpack will only ever run when `CNB_TARGET_DISTRO_NAME` and `CNB_TARGET_DISTRO_VERSION` matches those values (once the fix for https://github.com/buildpacks/lifecycle/issues/1337 is released).

If they were never not set (which can't currently happen anyway), IMO libcnb.rs should be opinionated and reject this edge case for the sake of normalising metadata to make buildpack authors lives easier. 

As-is, since both `distro_name` and `distro_version` are an `Option`, it means they are much more annoying to use. For languages that have distro-version specific binaries, the distro name/version has to be used in many places (from layer metadata, cache invalidation explanation messages, error messages explaining why a particular runtime version isn't available `"for this OS ({distro_name} {distro_version}"` etc) - and in every place I'm currently using `.expect()` since the scenario can't happen in practice.

As such, I think we should make them `String` instead of `Option<String>`.

(Longer term I'd love to advocate for changing the spec upstream to remove the "optional", but we can still improve the UX within libcnb.rs in the meantime.)

GUS-W-15639138.

edmorley:
* Makes the distro name/version fields non-optional (`String` instead of `Option<String>`), since (a) they will never be `None` in practice (see #820), and (b) it otherwise causes a lot of unnecessary friction for languages that have distro-version-specific binaries and need to use the metadata in multiple places in the buildpack.
* Implements `Clone` and `Debug` on `Target`, which unblocks a number of use-cases, such as including the target as part of a buildpack Error enum variant.

Fixes #820.
GUS-W-15639138.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -364,8 +364,10 @@ where
     let os = env::var("CNB_TARGET_OS").map_err(Error::CannotDetermineTargetOs)?;
     let arch = env::var("CNB_TARGET_ARCH").map_err(Error::CannotDetermineTargetArch)?;
     let arch_variant = env::var("CNB_TARGET_ARCH_VARIANT").ok();
-    let distro_name = env::var("CNB_TARGET_DISTRO_NAME").ok();
-    let distro_version = env::var("CNB_TARGET_DISTRO_VERSION").ok();
+    let distro_name =
+        env::var("CNB_TARGET_DISTRO_NAME").map_err(Error::CannotDetermineTargetDistroName)?;
+    let distro_version =
+        env::var("CNB_TARGET_DISTRO_VERSION").map_err(Error::CannotDetermineTargetDistroVersion)?;
 
     Ok(Target {
         os,
--- a/libcnb/src/target.rs
+++ b/libcnb/src/target.rs
@@ -1,35 +1,40 @@
+#[derive(Debug)]
 pub struct Target {
     /// The name of the target operating system.
     ///
     /// The value should conform to [Go's `$GOOS`](https://golang.org/doc/install/source#environment), for example
     /// `linux` or `windows`.
     ///
-    /// CNB `lifecycle` sources this value from the build OCI image's [`os` property](https://github.com/opencontainers/image-spec/blob/main/config.md#properties).
+    /// CNB `lifecycle` sources this value from the run OCI image's [`os` property](https://github.com/opencontainers/image-spec/blob/main/config.md#properties).
     pub os: String,
     /// The name of the target CPU architecture.
     ///
     /// The value should conform to [Go's $GOARCH](https://golang.org/doc/install/source#environment), for example
     /// `amd64` or `arm64`.
     ///
-    /// CNB `lifecycle` sources this value from the build OCI image's [`architecture` property](https://github.com/opencontainers/image-spec/blob/main/config.md#properties).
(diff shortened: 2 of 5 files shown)


libcnb/src/target.rs
```diff
@@ -1,35 +1,40 @@
...
     pub arch_variant: Option<String>,
     /// The name of the operating system distribution. Should be empty for Windows.
     ///
     /// For example: `ubuntu` or `alpine`.
     ///
-    /// CNB `lifecycle` sources this value from the build OCI image's `io.buildpacks.base.distro.name` label.
-    pub distro_name: Option<String>,
+    /// CNB `lifecycle` sources this value from either:
+    /// 1. The `io.buildpacks.base.distro.name` OCI image label, if set on the run image.
+    /// 2. Or else, the `ID` field of the `/etc/os-release` file in the build image.
```
edmorley, line 30: The labels are sourced from the run image, per:
https://github.com/buildpacks/spec/blob/platform/0.12/platform.md#target-data

However, the `/etc/os-release` fallback run against the build image, per:
https://github.com/buildpacks/lifecycle/blob/13d3691dd017405e22ad7b5050204cc2ccdf034a/platform/target_data.go#L16

libcnb/src/runtime.rs
```diff
@@ -364,8 +364,10 @@ where
     let os = env::var("CNB_TARGET_OS").map_err(Error::CannotDetermineTargetOs)?;
     let arch = env::var("CNB_TARGET_ARCH").map_err(Error::CannotDetermineTargetArch)?;
     let arch_variant = env::var("CNB_TARGET_ARCH_VARIANT").ok();
-    let distro_name = env::var("CNB_TARGET_DISTRO_NAME").ok();
-    let distro_version = env::var("CNB_TARGET_DISTRO_VERSION").ok();
+    let distro_name =
```
edmorley, line 367: I added a comment explaining this in 0303bab35956d5610d84f86d6f3bdd1e26853f2f

PR #824, opened 2024-05-16
Test fixture hack

runesoerensen:
When a git dependency is added to a Rust project, Cargo will inspect _all_ `Cargo.toml` files in the repository. This conflicts with some of our test fixtures, which a user of the library will never want to add to their project (they're either dummy projects for testing, and in one case, intentionally broken).

While it may be worthwhile to consider reporting this upstream, we may want to address it on our own in parallel. One solution would be to rename all test fixture `Cargo.toml` files to a different name, and then copy and rename them prior to running integration tests that rely on them.

This PR takes a slightly different approach - it's a quick hack that [uses this logic](https://github.com/rust-lang/cargo/blob/8d68ed4fb2554ae7ff2f0a3193c2651e31ee4e1e/src/cargo/ops/cargo_read_manifest.rs#L68) to prevent Cargo from traversing the test fixtures.


edmorley:
I'd be fine with adjusting the test with the intentionally broken `Cargo.toml` to something else that tests that packaging error handling, or if we can't come up with something else removing that specific test (I added a lot of tests around error handling after there was a regression, so losing one won't matter too much)

runesoerensen:
> (I'd prefer doing this to the current workaround in this PR.)

I'm not proud of it that's for sure :)

PR #831, opened 2024-06-11
Serialize process working directory as working-dir

runesoerensen:
Launch processes configured with a `WorkingDirectory::Directory` are currently incorrectly serialized. This PR fixes that by renaming the serialized `launch.toml` process working directory key from `working-directory` to `working-dir` ([as defined in the spec](https://github.com/buildpacks/spec/blob/7207c7dd711e28ddd045e0130e8a96e9200bf079/buildpack.md?plain=1#L909)).

This could be a breaking change for users of the library. However, a [quick GitHub search](https://github.com/search?q=WorkingDirectory%3A%3ADirectory+language%3ARust&type=code) didn't reveal any public/internal code that configures the working directory outside this repository.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- `libcnb-data`:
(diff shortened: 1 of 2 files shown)


Malax (approved):
Nice, thanks! :)

PR #844, opened 2024-07-20
Reset color settings after writing body

runesoerensen:
The color settings are currently reset immediately after writing the header value in the `log_warning` and `log_error` functions, but before setting and writing the (non-bold), colored body value.

As a result, the color setting persist after the function call (e.g. after logging a warning, subsequent `log_info` output will be yellow). I might be missing something here -- but I doubt that's the intended behavior?

### DIFF
--- a/libherokubuildpack/src/log.rs
+++ b/libherokubuildpack/src/log.rs
@@ -14,12 +14,12 @@ pub fn log_error(header: impl AsRef<str>, body: impl AsRef<str>) {
         .set_color(ColorSpec::new().set_fg(Some(Color::Red)).set_bold(true))
         .unwrap();
     writeln!(&mut stream, "\n[Error: {}]", header.as_ref()).unwrap();
-    stream.reset().unwrap();
 
     stream
         .set_color(ColorSpec::new().set_fg(Some(Color::Red)))
         .unwrap();
     writeln!(&mut stream, "{}", body.as_ref()).unwrap();
+    stream.reset().unwrap();
     stream.flush().unwrap();
 }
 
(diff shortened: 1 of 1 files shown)


libherokubuildpack/src/log.rs
```diff
@@ -14,12 +14,12 @@ pub fn log_error(header: impl AsRef<str>, body: impl AsRef<str>) {
         .set_color(ColorSpec::new().set_fg(Some(Color::Red)).set_bold(true))
         .unwrap();
     writeln!(&mut stream, "\n[Error: {}]", header.as_ref()).unwrap();
-    stream.reset().unwrap();
```
edmorley, line 17: I agree this is broken on `main` at the moment, and that we need to adjust the resets here, however, I think we need a few more changes.

Something that came up during the `buildpack_output` work, is that because other tools can end up adding prefixes to each log line (for example, endosome adds the `remote:` prefix, and Pack CLI can add the `[builder]` prefixes for an untrusted builder, or the timestamp prefixes when using `--timestamps`), the log helpers need to wrap each line separately, rather than being able to rely on the ANSI codes from an earlier line still applying for the current line. See:
https://github.com/heroku/libcnb.rs/blob/3fa2c6cd5b45f6b56e78ea69df84d838644d0e84/libherokubuildpack/src/buildpack_output/ansi_escape.rs#L1-L27

The `log_error` behaviour on `main` is currently equivalent to:

```
<colour>
[Error: Foo]
<reset><colour>Message
```

And with prefixes this ends up being:

```
[builder] <colour>
[builder] [Error: Foo]
[builder] <reset><colour>Message body
```

When really what we need is:

```
[builder] 
[builder] <colour>[Error: Foo]<reset>
[builder] <colour>Message body<reset>
```

ie: I think we need to change a few things:
1. Add the extra `.reset()` after writing the message body (as this PR does)
2. Leave the `.reset()` after writing the header (ie: skip the removal currently in this PR)
3. For the header, also change it to use two `writeln!` calls, rather than including the leading `\n` in a single call. The heading colour attributes would only then be specified on the second call.

Making those changes should fix #555. Those changes still won't fix the case where the message body is a multi-line string, but fixing the latter would be more involved and perhaps not worth doing given the `log_*` helpers are deprecated in favour of `buildpack_output`.

edmorley (changes requested):

edmorley:
Removing skip-changelog since this (a) changes user-facing behaviour, (b) will fix a bug (see #555) - so should be mentioned in the changelog.

PR #852, opened 2024-08-27
Remove Buildpack Output module

schneems:
In our July 9th meeting. We agreed to remove this library and use [bullet_stream](https://crates.io/crates/bullet_stream) for now.


schneems:
CI is fixed in https://github.com/heroku/libcnb.rs/pull/853

edmorley:
In today's meeting, I thought we said we'd leave this in for now, and people can use either this or bullet_stream as they prefer short term?

Re switching to bullet_stream, is there an easy way to see what changed between the implementation here and bullet_stream? (Ideally a list of commits or PRs, presuming the implementation here used as the base?)

colincasey:
I'm in approval of this PR since the API for [bullet_stream](https://crates.io/crates/bullet_stream) is arguably better and I would prefer we don't commit to an "official" output implementation yet (which would be upstreamed here in the future). 

That said, I think some folks were considering using the implementation here so they should probably speak up.

edmorley:
> since the API for [bullet_stream](https://crates.io/crates/bullet_stream) is arguably better

It's all very well stating this, but the rest of the team have no idea what the differences are between the libcnb.rs build_output` implementation, because there's been no communication about this, and the commit sequence at https://github.com/schneems/bullet_stream/commits/main/ doesn't make it clear what's changed since it forked etc.

The onus shouldn't be on the rest of us to diff the `docs.rs` API docs to figure this out.

schneems:
## This PR

This PR is me doing what I said I would do almost 2 months ago in our meeting. 

Prior to that:

- https://github.com/heroku/libcnb.rs/pull/784#issuecomment-2168587572
- https://github.com/heroku/libcnb.rs/pull/787#issuecomment-2168587923

## Today's meeting

In today's meeting we could have made one of two decisions:

- 1. Force everyone to use `bullet_stream`.
- 2. Encourage people to use `bullet_stream` but leave room for them to explore.

We picked number 2. Functionally, unless you'll personally be writing your own logging code for the python buildpack it would mean adopting `bullet_stream`.

> It's all very well stating this, but the rest of the team have no idea what the differences are between the libcnb.rs build_output implementation

This is irrelevant, no one on the team has used the libcnb.rs implementation. Including me. If you're using it, that would be news to me.

> , because there's been no communication about this, and the commit sequence at https://github.com/schneems/bullet_stream/commits/main/ doesn't make it clear what's changed since it forked etc.

The original PR to libcnb.rs that was merged was effectively non-functional. Bullet stream mostly my original libcnb.rs PR with the features that are needed to allow the Ruby buildpack work. But I massaged it until the API made more sense. The biggest change is that the API is now less abstract and more concrete i.e. to get a bullet point you call `bullet()` instead of `section()`.

Please let me know if you find a critical difference between libcnb.rs implementation and `bullet_stream` that you feel should be documented. I've been begging people to care about this codebase and use case for a long time. I continue to do so.

## Moving forward

It's a mistake to leave dead code in this library. I support output and logging competitors from other maintainers and community members. I discourage taking on maintenance of this very-similar-but-not-quite-the-same library.

edmorley:
> We picked number 2

No, we picked "short term people can use whatever library they want to implement the new logging style".

edmorley:
> This is irrelevant, no one on the team has used the libcnb.rs implementation

The libcnb.rs implementation is what had code review and a lot of time invested from myself and Manuel, so it's the API with which we are familiar. So I think it's perfectly understandable why we'd want to use it as the basis to understand an alternate API.

schneems:
> So I think it's perfectly understandable why we'd want to use it as the basis to understand an alternate API.

When I hear that what It sounds like is that you'll attempt to wire your entire buildpack to use this code and then if you are okay with it you won't look at `bullet_stream`. I'm worried that if that happens we'll be left with all buildpacks using one type of logging code and one outlier.

What I want is to live in a world where we all use each other's stuff and libraries live or die based on their merits. I want people to either help me write better libraries, give me better libraries, or not block me from moving forward. I'm asking you to not block me on this.

I intentionally wrote the code so it's easy to port around. If you really want to use it, a copy will exist forever in the git history, you can copy it into your buildpack. If that's not acceptable to you, then I ask for a time limit. If not now, when can we delete this?

edmorley:
Let's move the discussion to Slack :-)

edmorley (approved):

PR #861, opened 2024-09-24
Migrate inventory code

runesoerensen:
This PR imports and migrates [the inventory code here](https://github.com/Malax/inventory/tree/main/inventory/src) to the `libherokubuildpack` crate. The code was copied as-is and added in the first commit, and [I've tried to keep changes to integrate it to a minimum](https://github.com/heroku/libcnb.rs/compare/0abce6a..369a70f).

I'm not sure what the best approach is to adapt the `sha2` and `semver` feature flags from the source inventory crate - perhaps it's fine as-is, or perhaps it'd make sense to use different feature names given the new context/crate for the code. [Relevant commit with comment](https://github.com/heroku/libcnb.rs/pull/861/commits/7cb5e0f63685921d478085696a8253cc4b03f050).

I've removed a couple re-exports that appear to be unnecessary - [see explanation in this commit](https://github.com/heroku/libcnb.rs/pull/861/commits/369a70f8a1e801e365863dffabe48289e53ff93c).


libherokubuildpack/src/inventory/unit.rs
```diff
@@ -0,0 +1,11 @@
+use crate::inventory::checksum::Digest;
+
+impl Digest for () {
```
joshwlewis, line 3: As far as I can tell, this is unused. Should we remove it?

If we want to keep it - maybe we should give it a name? `pub struct NoChecksum`, `UnknownChecksum`, `AnyDigest`, or something.

Malax: This is useful in tests where you might not want to have an actual checksum or when you want to read an inventory file where the reading code doesn't have an implementation of the used digest. This will happen when we have a generic CLI tool to work with inventory files.

 Using `()`/unit seems fine to me (unsurprisingly, since I wrote the code in question). There is no need to introduce a custom type to convey "no information", it's actually the formal definition of unit.

Quoting Wikipedia:
> In the area of [mathematical logic](https://en.wikipedia.org/wiki/Mathematical_logic) and [computer science](https://en.wikipedia.org/wiki/Computer_science) known as [type theory](https://en.wikipedia.org/wiki/Type_theory), a unit type is a [type](https://en.wikipedia.org/wiki/Data_type) that allows only one value (and thus can hold no information).

Note: There is a `BogusDigest` used in tests for the inventory. It's different from this as we there actually test that digests are validated and only match if the digest implementation matches as well.

I would say lets keep it around as-is for now, maybe create an issue to re-evaluate later. I'm pretty sure we'll need it again when the CLI is re-introduced (it intentionally wasn't ported with this PR). It's sitting in its own dedicated module, isn't explicitly interwoven with code outside of the module (since it's just a trait implementation). The complexity cost on the code-base is pretty much zero.

libherokubuildpack/src/inventory.rs
```diff
@@ -0,0 +1,178 @@
...
+            .filter(|artifact| {
+                artifact.os == os
+                    && artifact.arch == arch
+                    && requirement.satisfies_version(&artifact.version)
+                    && requirement.satisfies_metadata(&artifact.metadata)
+            })
+            .max_by_key(|artifact| &artifact.version)
+    }
+
+    pub fn partial_resolve<R>(
```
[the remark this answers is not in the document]
Malax: Note the trait bound on `V`. One is `Ord` the other is `PartialOrd`. Some version implementations are only partially ordered. One example could be `f32` which is not totally ordered. As a more practical example: OpenJDK versions are only partially ordered as JEP 322 versions can only be compared to a legacy version when no version coordinates beyond the major version are specified.

libherokubuildpack/Cargo.toml
```diff
@@ -18,11 +18,14 @@ all-features = true
 workspace = true
 
 [features]
-default = ["command", "download", "digest", "error", "log", "tar", "toml", "fs", "write", "buildpack_output"]
+default = ["command", "download", "digest", "error", "inventory", "log", "semver", "sha2", "tar", "toml", "fs", "write", "buildpack_output"]
```
Malax, line 21: having `semver` and `sha2` as features made sense when they were in the inventory crate. In the context of the whole `libherokubuildpack` they seem to be worded to broadly. Maybe we can prefix them with `inventory-`?

Malax (changes requested):

schneems:
There's a few comments on docs. I took a stab at it in a stacked PR https://github.com/heroku/libcnb.rs/pull/864. Feedback/review is welcome over there or it's free to be merged in here and iterated on. The only known pending thing is i put the feature names in the docs as TODO.

libherokubuildpack/Cargo.toml
runesoerensen: Agree, it makes more sense to prefix them -- I've updated the feature names to `inventory-sha2` and `inventory-sermver`.

runesoerensen:
@joshwlewis I've merged Richard's documentation PR - I think should address your documentation comments?

libherokubuildpack/src/inventory.rs
```diff
@@ -0,0 +1,178 @@
...
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn push(&mut self, artifact: Artifact<V, D, M>) {
+        self.artifacts.push(artifact);
+    }
+
+    pub fn resolve<R>(&self, os: Os, arch: Arch, requirement: &R) -> Option<&Artifact<V, D, M>>
```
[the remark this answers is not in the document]
schneems: Docs are added and the module docs have an example that calls this function. I think that's good enough for now.

libherokubuildpack/src/inventory.rs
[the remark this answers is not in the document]
schneems: I think we can resolve this. Also. Thanks for the answer. I used it to help write documentation.

libherokubuildpack/src/inventory/unit.rs
joshwlewis: > This is useful in tests where you might not want to have an actual checksum or when you want to read an inventory file where the reading code doesn't have an implementation of the used digest. This will happen when we have a generic CLI tool to work with inventory files.

There's a lot more information in this explanation than "No Information". Quoting the Rust docs:

> The () type has exactly one value `()`, and is used when there is no other meaningful value that could be returned.

A `NoChecksum` unit could be a more meaningful value to me.

In any case, this isn't blocking for me.

joshwlewis (approved):

joshwlewis:
@Malax I think this is waiting on you now. It looks like the changes you requested have been made. Can we get a 👍🏻?

Malax (approved):

PR #864, opened 2024-10-01
[stacked] Add docs to inventory code

schneems:


joshwlewis (approved):
Works for me. What do you think about merging this to your PR @runesoerensen.

libherokubuildpack/src/inventory/artifact.rs
```diff
@@ -4,6 +4,15 @@ use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 use std::str::FromStr;
 
+/// Representation of a downloadable artifact such as a binary tarball.
+///
+/// An inventory is made up of multiple artifacts that have a version that
+/// can be compared to each other and a URL where the artifact can be downloaded.
+///
+/// Some artifacts may be OS or architectures specific. The checksum value can
+/// be used to validate an artifact once it has been downloaded.
```
runesoerensen, line 13:
```suggestion
/// Artifacts are OS and architectures specific. The checksum value can
/// be used to validate an artifact once it has been downloaded.
```

libherokubuildpack/src/inventory/artifact.rs
```diff
@@ -4,6 +4,15 @@ use serde::{Deserialize, Serialize};
 use std::fmt::Display;
 use std::str::FromStr;
 
+/// Representation of a downloadable artifact such as a binary tarball.
+///
+/// An inventory is made up of multiple artifacts that have a version that
+/// can be compared to each other and a URL where the artifact can be downloaded.
```
runesoerensen, line 10: This is a bit inaccurate.. The inventory itself doesn't actually require the artifact version to be comparable -- the `resolve` and `resolve_partial` functions add those constraints (e.g. `Ord` and `PartialOrd`) on the `Version` type.

runesoerensen:
Makes sense to merge the changes into the other branch! I've added a few comments, but this branch should also be updated to use the new feature flags (and replace the TODO in the comments).

libherokubuildpack/src/inventory/artifact.rs
schneems: The jruby binaries aren't. OS or architecture specific. However, I guess I'm nulling that out manually by writing the same URL twice. So technically your version is true and too. 👍🏻

libherokubuildpack/src/inventory.rs
```diff
@@ -31,15 +110,23 @@ impl<V, D, M> Default for Inventory<V, D, M> {
...
+    /// Creates a new empty inventory
     #[must_use]
     pub fn new() -> Self {
         Self::default()
     }
 
+    /// Add a new artifact to the in-memory inventory
+    ///
+    /// For removal, you can use the `Vec::retain` method on the artifacts attribute
+    /// to remove artifacts that match a given predicate.
```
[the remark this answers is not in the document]
schneems: I needed to delete/reject artifacts in the ruby binary builder https://github.com/heroku/docker-heroku-ruby-builder/blob/c5f2030754560bc8d830053a2935063ab14b491a/ruby_executable/src/bin/ruby_build.rs#L201-L203.

I'm fine removing this as well. It's a little weird that the field is public which allows you to push to the vec, but there's a separate push associated function.

> Otherwise, artifacts attribute should probably be changed to artifacts field

In Ruby they're called attributes i.e. `attr_accessor` also I keep wanting to call associated functions methods. Anywhoo. Deleting this all is fine.

schneems:
Updated with the latest features. Leaving the merge up to you.

libherokubuildpack/src/inventory/artifact.rs
runesoerensen: I mean you're right in the sense that most consumers of this module will likely use the `resolve` and `resolve_partial` methods on the `Inventory` struct. But this is documentation for the `Artifact` struct, which is deliberately decoupled from any `ArtifactRequirement` constraints -- so I guess that's why it stood out to me in this context.

libherokubuildpack/src/inventory.rs
```diff
@@ -1,3 +1,76 @@
...
+//! let requirement = VersionReq::parse("= 1.0.0").unwrap();
+//! if let Some(artifact) = inventory.resolve(Os::Linux, Arch::Amd64, &requirement) {
+//!     // Downloading the artifact
+//!     println!("Installing {requirement:?} from {}", artifact.url);
+//!     let path = Path::new("path/to/binary");
+//!     download_file(&artifact.url, &path)
+//!        .unwrap();
+//!
+//!     // Validating the checksum
+//!     Checksum::<Sha256>::from_str(&std::fs::read_to_string(&path).unwrap())
```
runesoerensen, line 57: The `FromStr` implementation for `Checksum` doesn't calculate the checksum for a file like this. It just parses a string containing the checksum value and name, and validating the format against the `Digest` implementation, e.g.:

    Checksum::<Sha256>::from_str(
        "sha256:2c26b46b68ffc68ff99b453c1d30413413422d706483bfa0f98a5e886266e7ae",
    )

runesoerensen: https://github.com/heroku/libcnb.rs/pull/864/commits/28204a087982c968e23e6b2b8c0e6b2b4ce26ed2 addresses this

runesoerensen:
I've updated the example usage docs in `inventory.rs` to show full end-to-end usage that works (and runs during CI) https://github.com/heroku/libcnb.rs/pull/864/commits/28204a087982c968e23e6b2b8c0e6b2b4ce26ed2 - feel free to adjust!

libherokubuildpack/src/inventory.rs
schneems: Good catch. I was wanting to show a few things that were not intuitive (clearly, since my example is wrong). I wanted to show:

- Generating a SHA from a file on disk
- Using that value to compare to the digest from the inventory
- Printing a human readable version of the sha

I will add a line to the example that shows `hex::encode` for printing.

schneems:
@runesoerensen thanks. I like that the example also now gives a sample of what the resulting inventory.toml format will look like.

I added a line that shows how to `hex::encode` the value for printing/display. I would ideally like someone to also see the file interactions i.e. `download_file` right in this example, but couldn't figure out a way to make it work and run. I would also like to show generating a sha from disk, but it's not that hard to infer that if we can get it from a string we can `std::fs::read_to_string`. 

Given all of that, I think what we've got is pretty great. Take a look at my commit and either merge or comment.

PR #870, opened 2024-10-20
Support for Docker "--volume" mounts during integration tests.

mars:
This enables using Docker volumes in integration tests, by implementing `libcnb_test::ContainerConfig::volumes()`, which generates `--volume XXXXX` arguments on the underlying `docker run` command.

### DIFF
--- a/libcnb-test/src/container_config.rs
+++ b/libcnb-test/src/container_config.rs
@@ -31,6 +31,7 @@ pub struct ContainerConfig {
     pub(crate) command: Option<Vec<String>>,
     pub(crate) env: HashMap<String, String>,
     pub(crate) exposed_ports: HashSet<u16>,
+    pub(crate) volumes: Option<Vec<String>>,
 }
 
 impl ContainerConfig {
@@ -196,6 +197,33 @@ impl ContainerConfig {
             self.env(key.into(), value.into());
         });
 
+        self
+    }
+
+    /// Attaches container volumes. Useful for integration tests that
+    /// depend on persistent storage shared between container executions.
+    ///
+    /// See: [Docker CLI, Mount Volume](https://docs.docker.com/reference/cli/docker/container/run/#volume)
+    ///
+    /// # Example
+    /// ```no_run
+    /// use libcnb_test::{BuildConfig, ContainerConfig, TestRunner};
+    ///
+    /// TestRunner::default().build(
+    ///     BuildConfig::new("heroku/builder:22", "tests/fixtures/app"),
+    ///     |context| {
+    ///         // ...
+    ///         context.start_container(
+    ///             ContainerConfig::new().volumes(["/shared/cache:/workspace/cache"]),
+    ///             |container| {
+    ///                 // ...
+    ///             },
+    ///         );
+    ///     },
+    /// );
+    /// ```
+    pub fn volumes<I: IntoIterator<Item = S>, S: Into<String>>(&mut self, volumes: I) -> &mut Self {
+        self.volumes = Some(volumes.into_iter().map(S::into).collect());
         self
     }
 }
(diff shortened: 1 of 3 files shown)


runesoerensen:
Would it perhaps make sense to use an interface similar to how we currently handle environment variables, e.g. `pub fn volume(&mut self, source: impl AsRef<Path>, destination: impl AsRef<Path>) -> &mut Self`?

See [this PR](https://github.com/heroku/libcnb.rs/pull/871) for a full implementation using that approach.

mars:
@runesoerensen if y'all would prefer that more granular interface implementation, that's your call. We can close this PR and use yours instead.

I just hope to have this implemented/merged, so that I can switch my dependent project back to using the main distribution of libcnb.rs, instead of my branch `docker-volume`.

runesoerensen:
> `docker run` supports multiple volume args: `--volume /a:/a --volume /b:/b`.
> 
> This changeset supports specifying multiple `--volume` from that vector of strings:

The implementation I proposed actually also supports specifying multiple volume/bind mounts (and arguably in a way that's more consistent with the way the parameters are passed to the docker CLI). I've since [changed the semantics a bit](https://github.com/heroku/libcnb.rs/pull/871/commits/6f4aea46cb318e42df545bfdfb618aff9207d211) to use `bind_mount` rather than `volume`, but you can still mount multiple paths by chaining `bind_mount` calls on the builder.

```rust
ContainerConfig::new()
    .bind_mount(
        PathBuf::from("/shared/cache"),
        PathBuf::from("/workspace/cache"),
    )
    .bind_mount("/shared/models", "/workspace/models"); // Less verbose, but retaining the `PathBuf` semantics.
```

My primary motivation for suggesting this approach is that it's more consistent with how `ContainerConfig` is currently used to construct Docker CLI arguments. For instance, when setting environment variables, the string manipulation required to specify `--env KEY=VALUE` is [handled here](https://github.com/heroku/libcnb.rs/blob/a3174bc77311eb8eb58b66decb9e0cbebb15d6e4/libcnb-test/src/docker.rs#L93-L95), which joins the `KEY` and `VALUE` with `=`.

On that note, calling the `ContainerConfig`'s `volumes` function [as implemented in this PR](https://github.com/heroku/libcnb.rs/pull/870/files#diff-da467816e4ab0cf374c54f28ef30e66dcf809ca854e3180b3a135f8b48c410f0R225-R228) will overwrite the `volumes` struct field. It seems this breaks a bit with the builder pattern, and may at least be less intuitive in this context (for instance, the similar `envs` function [inserts items into the existing collection](https://github.com/heroku/libcnb.rs/pull/870/files#diff-da467816e4ab0cf374c54f28ef30e66dcf809ca854e3180b3a135f8b48c410f0R192-R201), allowing for multiple chained calls to add more `envs` without dropping the previous state).

Not looking to over-engineer things here, but I changed the implementation from using the `--volume` parameter to `--mount`, which seems more accurate and allow for better semantics. Docker's `--volume` parameter is a "legacy" parameter and not recommended for new implementations, combining multiple different operations (e.g. creating the source/host directory if it doesn't exist (unlike `--mount type=bind,source={},target={}`, which will return an error if the host directory doesn't exist), allowing for both (named and anonymous) volume and bind mounts), in ways that aren't immediately obvious. So rather than having a `volumes` field on `ContainerConfiguration` there's now a `bind_mounts` field. This allows us to have a cleaner interface for the `bind_mount` function, which takes source and target `impl Into<PathBuf>` parameters. If we want to support volume mounts in the future, we could add a function for that taking `String` source and `impl Into<PathBuf>` target parameters.

mars:
Thanks @runesoerensen for the deeper explanation, evidence, and alternative PR. I am totally fine with switching from volumes to bind_mounts. That sounds good 👍

runesoerensen:
Ok great, I've [merged the PR](https://github.com/heroku/libcnb.rs/commit/7f089eb1b411153baf3f16902106443e4fd34c2b) and cut a new release, so you can switch back to using the version published on crates.io.

I'll close this PR as the functionality is now supported - thanks for your work on this @mars !

PR #874, opened 2024-10-29
Update download.rs to match non-CNB operation

usiegj00:


edmorley (changes requested):
Thank you for the PR, and sorry for the delayed reply. 

Unfortunately I don't think we could accept this as-is, since:
- it retries all errors, rather than just transient ones
- it outputs to stderr using `eprintln!` when many of our buildpacks use a more advanced logger that provides consistent output. However, libcnb/libherokubuildpack is meant to be standalone, which means we'll have to figure out how to support this case. We've had other use cases (eg #371) where we've had a need for libcnb to be able to log things - so maybe we need a logging API within libcnb itself? Or alternatively and simpler, perhaps the download helper here in libherokubuildpack needs to accept a callback function inside which the consumer could log in their desired format.

Lastly, several buildpacks currently use their own download functionality rather than that in `libherokubuildpack` (for various reasons, including needing to support zstd, checksum verification etc), so we need to decide what the long term approach for this utility should be.

These are all things above and beyond what we'd expect you to do of course - but mentioning here so you understand why it's not straightforwards to accept this change (and also why we haven't replied yet, since we were hoping we might have had a moment to figure this out). Adding retries and timeouts across all of our CNBs is definitely on our internal TODO list though.

PR #884, opened 2024-11-25
Use IPv6 in in integration tests

schneems:
Specify IPv6 host `::1` which is the equivalent of IPv4 host `127.0.0.1`.


schneems:
I made this PR to debug a failure, and this wasn't the cause. Might still want to do it, but it's not urgent.

PR #890, opened 2024-12-10
libherokubuildpack: Fix colour resetting for the `log_*` macros

### LINKED ISSUE #555: Pack output eats libherokubuildpack colors

edmorley:
This fixes some long-standing ANSI colour bugs with the `log_header`, `log_error` and `log_warning` macros. 

Whilst we soon want to move away to more advanced logging libraries that use the new logging style, there are still many buildpacks using these macros which will benefit short term from these fixes (Procfile, Go, Node.js, JVM, Python, PHP, buildpacks-release-phase, buildpacks-frontend-web).

The logging macros would previously emit output roughly like:

```
<colour>
[Error: Foo]
<reset><colour>Message line one.
Message line two.
```

This was not only missing the final `<reset>`, but also didn't wrap each line individually with colour codes/resets.

This causes issues when lines end up prefixed - such as the Git `remote:` prefix, or when using Pack CLI locally with an untrusted build (which adds the colourised `[builder]` prefixes) or `--timestamps` mode.

For example in this:

```
remote: <colour>
remote: [Error: Foo]
remote: <reset><colour>Message line one.
remote: Message line two.
```

...several of the `remote:`s would inherit the colours.

Instead what we need is:

```
remote: <colour><reset>
remote: <colour>[Error: Foo]<reset>
remote: <colour>Message line one.<reset>
remote: <colour>Message line two.<reset>
```

Fixes #555.
Closes #844.
GUS-W-17400078.

### DIFF
--- a/libherokubuildpack/src/log.rs
+++ b/libherokubuildpack/src/log.rs
@@ -1,4 +1,4 @@
-use std::io::Write;
+use std::io::{self, Write};
 use termcolor::{Color, ColorChoice, ColorSpec, StandardStream, WriteColor};
 
 /// # Panics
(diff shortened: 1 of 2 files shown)


edmorley:
### Before

before

### After

after

libherokubuildpack/src/log.rs
```diff
@@ -72,3 +74,22 @@ pub fn log_info(message: impl AsRef<str>) {
...
+// Styles each line of text separately, so that when buildpack output is streamed to the
+// user (and prefixes like `remote:` added) the line colour doesn't leak into the prefixes.
+fn write_styled_message(
+    stream: &mut StandardStream,
+    message: impl AsRef<str>,
+    spec: &ColorSpec,
+) -> io::Result<()> {
+    // Using `.split('\n')` rather than `.lines()` since the latter eats trailing newlines in
+    // the passed message, which would (a) prevent the caller from being able to add spacing at
+    // the end of their message and (b) differ from the `println!` / `writeln!` behaviour.
```
Malax, line 87: Nice attention to detail here :)

Malax (approved):

PR #908, opened 2025-02-04
Update test assertions to match new procfile output

joshwlewis:
Newer versions of the procfile buildpack have different output, so CI tests are failing. This fixes the tests.


edmorley (approved):
Thank you!

PR #907, opened 2025-02-04
Update to opentelemetry 0.24

joshwlewis:
This updates opentelemetry to the last version that supports JSONL file exports -- at least for the moment. There was an upstream change to simplify the `opentelemetry-stdout` exporter: https://github.com/open-telemetry/opentelemetry-rust/pull/2040. With that change, the exporter lost the ability to send telemetry to a generic writer (like a buffer or file), and also lost `jsonl` serialization. Based on [this comment](https://github.com/open-telemetry/opentelemetry-rust/pull/2040#issuecomment-2508805391), there is interest in bringing this functionality back in some other form. 

I've opened an issue upstream: https://github.com/open-telemetry/opentelemetry-rust/issues/2602

Supersedes #886

PR #909, opened 2025-02-06
Custom OTLP File Exporter + opentelemetry updates

joshwlewis:
As mentioned in #907, there was an upstream change to `opentelemetry-stdout`'s exporter: https://github.com/open-telemetry/opentelemetry-rust/pull/2040. With that change, the exporter lost the ability to send telemetry to a generic writer (like a buffer or file), and also lost `jsonl` serialization. This means we can't export in the correct format with newer versions of the `opentelemetry` libraries without a new exporter. This PR adds a custom OTLP File Exporter (spec is [here](https://opentelemetry.io/docs/specs/otel/protocol/file-exporter/)) for traces and updates the `opentelemetry` libraries to the latest.

Based on [this comment](https://github.com/open-telemetry/opentelemetry-rust/pull/2040#issuecomment-2508805391), and my [upstream issue](https://github.com/open-telemetry/opentelemetry-rust/issues/2602), there is interest in an OTLP File Exporter upstream, so this custom exporter shouldn't live here forever.

This exporter probably isn't quite good enough to upstream yet -- it supports traces, but not metrics or logs.

Bonus changes:

- Now uses batch export functionality, which is generally preferred according to the [spec](https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/README.md). In `opentelemetry-sdk` `0.28`, the batch export functionality no longer requires an async runtime (like `tokio`) and instead runs in a separate background thread (some details [here](https://github.com/open-telemetry/opentelemetry-rust/issues/2066)), which lowers the dependency overhead for us.
- Slightly better feature usage -- we're no longer compiling the `metrics` and `logs` features from `opentelemetry`, which we don't use anyway
- This uses a `LineWriter` rather than a `BufWriter` now. The former is better for our use case -- we want to write complete lines to the file to prevent writing partial json to the file and corrupting it.
- There were a truckload of random API changes in `opentelemetry`. Some things were moved, others renamed. Almost everything uses the builder pattern now.

### DIFF
--- a/libcnb/Cargo.toml
+++ b/libcnb/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 trace = [
     "dep:opentelemetry",
     "dep:opentelemetry_sdk",
-    "dep:opentelemetry-proto ",
+    "dep:opentelemetry-proto",
     "dep:serde_json",
 ]
 
--- a/libcnb/Cargo.toml
+++ b/libcnb/Cargo.toml
@@ -35,9 +35,9 @@ libcnb-common.workspace = true
 libcnb-data.workspace = true
 libcnb-proc-macros.workspace = true
 futures-core = { version = "0.3", optional = true }
-opentelemetry = { version = "0.27.0", optional = true, default-features = false }
-opentelemetry_sdk = { version = "0.27.0", optional = true, default-features = false }
-opentelemetry-proto = { version = "0.27.0", optional = true, default-features = false }
+opentelemetry = { version = "0.28.0", optional = true, default-features = false }
+opentelemetry_sdk = { version = "0.28.0", optional = true, default-features = false }
+opentelemetry-proto = { version = "0.28.0", optional = true, default-features = false }
 serde = { version = "1.0.215", features = ["derive"] }
 serde_json = { version = "1.0.133", optional = true }
 thiserror = "2.0.6"
--- a/libcnb/Cargo.toml
+++ b/libcnb/Cargo.toml
@@ -18,7 +18,7 @@ workspace = true
 trace = [
     "dep:opentelemetry",
     "dep:opentelemetry_sdk",
-    "dep:opentelemetry-stdout",
+    "dep:opentelemetry_proto ",
 ]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


libcnb/src/tracing.rs
```diff
@@ -40,45 +49,44 @@ pub(crate) fn start_trace(buildpack: &Buildpack, phase_name: &'static str) -> Bu
     if let Some(parent_dir) = tracing_file_path.parent() {
         let _ = std::fs::create_dir_all(parent_dir);
     }
-    let exporter = match std::fs::File::options()
+
+    let resource = Resource::new([
+        // Define a resource that defines the trace provider.
+        // The buildpac name/version seems to map well to the suggestion here
```
colincasey, line 55:
```suggestion
        // The buildpack name/version seems to map well to the suggestion here
```

joshwlewis: Thanks - fixed it myself before I saw this =)

colincasey (approved):
@joshwlewis This looks good but, as we discussed, it will need to also mark the span passed into `BuildpackTrace` as the "active span" for later use in our individual buildpacks:
https://docs.rs/opentelemetry/latest/opentelemetry/trace/fn.mark_span_as_active.html

joshwlewis:
@colincasey I'll add the global span in another PR since it's sort of a new feature for libcnb.rs, where this is more of a dependency upgrade.

libcnb/Cargo.toml
```diff
@@ -16,9 +16,16 @@ workspace = true
...
 trace = [
+    "dep:futures-core",
     "dep:opentelemetry",
+    "opentelemetry/trace",
     "dep:opentelemetry_sdk",
-    "dep:opentelemetry-stdout",
+    "opentelemetry_sdk/trace",
+    "dep:opentelemetry-proto",
+    "opentelemetry-proto/trace",
+    "opentelemetry-proto/gen-tonic-messages",
```
edmorley, line 26: There's probably not much we can/should do about it, but it seems the `gen-tonic-messages` feature pulls in a whole load of additional deps (tonic, tower, prost etc), eg:
https://github.com/heroku/buildpacks-procfile/pull/262/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eR1078

(I was hoping after https://github.com/open-telemetry/opentelemetry-rust/pull/1569 updating to newer OTel might see a saving in buildpack crates, but the dependency count has increased).

Anyway, we need this functionality (and writing the transformation implementation ourselves would be worse), so there's not much we can do...

PR #917, opened 2025-02-27
Set active span on thread instead of stack

joshwlewis:
WIP!

Alternative to #918. The idea here is to set the active span on the thread so that buildpacks can nest their own spans within the libcnb provided one - extending the waterfall trace from kodon through libcnb and into CNB specific traces. 

Usage within a buildpack might look like this:

```rust
let tracer = global::tracer("my-buildpack");
// This span will be nested within libcnb.rs-generated span
tracer.in_span("download-go", |ctx| {
    let s = ctx.span();
    s.set_attribute(KeyValue::new("go_version", go_version.to_string()));
    do_the_download();
    // This event will be affiliated with the 'download-go' span
    s.add_event("downloaded", vec![]);
});
```

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -6,7 +6,7 @@ use crate::error::Error;
 use crate::platform::Platform;
 use crate::sbom::cnb_sbom_path;
 #[cfg(feature = "trace")]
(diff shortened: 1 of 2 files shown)


joshwlewis:
Closing in favor of #918.

PR #918, opened 2025-02-27
Use `tracing` library to generate spans

joshwlewis:
Alternative to #917 -- personally, I prefer this one. This sets up `tracing` so that it emits trace data into our opentelemetry file exporter pipeline while correctly honoring the cascade of events from kodon, through libcnb, and buildpack specific spans.

Using `tracing` for recording span data in CNBs looks like this:

```rust
{
  // This span will be nested within libcnb.rs-generated span
  let _span_guard = tracing::span!(Level::INFO, "download-go", version = go_version).entered();
  do_the_download();
  // This event will be affiliated with the 'download-go' span
  tracing::event!(Level::INFO, "downloaded");
}
// The _span_guard is out of scope, so the 'download-go' span is ended
```

`tracing` is a widely used library, even within libraries. This seems like the more rust-like approach. #917 by comparison is the more `opentelemetry`-like approach and closer to that standard.

As an added bonus the `tracing` library automatically adds `code.lineno`, `code.filepath`, `code.namespace` attributes to spans and events.

There are some changes to the shape of the tracing data - some span names, some attributes have changed. Most of this is because we are accepting `tracing-opentelemetry` conventions.

### DIFF
--- a/libcnb/src/runtime.rs
+++ b/libcnb/src/runtime.rs
@@ -19,7 +19,6 @@ use std::fmt::Debug;
 use std::path::{Path, PathBuf};
 use std::process::exit;
 use std::{env, fs};
-use tracing::Level;
 
 /// Main entry point for this framework.
 ///
@@ -137,13 +136,13 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
     #[cfg(feature = "trace")]
     let _trace_guard = init_tracing(&buildpack_descriptor.buildpack, "detect");
     #[cfg(feature = "trace")]
-    let _span_guard = tracing::span!(Level::INFO, "libcnb-detect").entered();
+    let _span_guard = tracing::span!(tracing::Level::INFO, "libcnb-detect").entered();
     #[cfg(feature = "trace")]
     let trace_error =
         |error: &dyn std::error::Error| tracing::error!(?error, "libcnb-detect-error");
 
     #[cfg(not(feature = "trace"))]
-    let mut trace_error = |_: &dyn std::error::Error| {};
+    let trace_error = |_: &dyn std::error::Error| {};
 
     let platform = B::Platform::from_path(&args.platform_dir_path)
         .map_err(Error::CannotCreatePlatformFromPath)
@@ -168,7 +167,7 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
     match detect_result.0 {
         InnerDetectResult::Fail => {
             #[cfg(feature = "trace")]
-            tracing::event!(Level::INFO, "libcnb-detect-failed");
+            tracing::event!(tracing::Level::INFO, "libcnb-detect-failed");
             Ok(exit_code::DETECT_DETECTION_FAILED)
         }
         InnerDetectResult::Pass { build_plan } => {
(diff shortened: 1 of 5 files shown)


colincasey:
Ooh, I like this more than #917.

libcnb/src/runtime.rs
```diff
@@ -134,15 +134,15 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
...
-
+    let _trace_guard = init_tracing(&buildpack_descriptor.buildpack, "detect");
     #[cfg(feature = "trace")]
-    let mut trace_error = |err: &dyn std::error::Error| {
-        trace.set_error(err);
-    };
+    let _span_guard = tracing::span!(tracing::Level::INFO, "libcnb-detect").entered();
+    #[cfg(feature = "trace")]
+    let trace_error =
+        |error: &dyn std::error::Error| tracing::error!(?error, "libcnb-detect-error");
```
edmorley, line 142: When I view this file in my editor, Rust analyzer reports an error that rustc and/or Clippy don't:

```
expected &dyn Error, found &Error<<B as Buildpack>::Error>rust-analyzer[E0308](https://doc.rust-lang.org/stable/error_codes/E0308.html)
```

If I replace the type here with the type mentioned in the error (`&Error<<B as Buildpack>::Error`) then both Rust analyzer and the rustc/clippy are happy, and Clippy actually suggests some further simplifications to avoid the unnecessary closures at the points where `trace_error` is used.

Combining those and a tweak to avoid needing to declare `trace_error` twice / having to use `not(feature = "trace"))`, this was the end result:
https://github.com/heroku/libcnb.rs/commit/34811c4a3bbdc6b8df26b9bbc3e255fa92d64582

Thoughts?

I pushed it as a commit to a separate branch in case you'd prefer something different :-)

edmorley (approved):
Thank you! I really like the approach of using the `tracing` crate - it seems this will be easier for individual language CNBs to adopt :-)

libcnb/src/runtime.rs
joshwlewis: Yeah, my editor complains too. Those errors have been there prior to this branch -- I assumed it was some kind of weird bug/inconsistency in rust-analyzer. I like it, I'll bring it in.

libcnb/src/runtime.rs
```diff
@@ -134,23 +134,22 @@ pub fn libcnb_runtime_detect<B: Buildpack>(
...
+    let _trace_guard = init_tracing(&buildpack_descriptor.buildpack, "detect");
     #[cfg(feature = "trace")]
-    let mut trace_error = |err: &dyn std::error::Error| {
-        trace.set_error(err);
-    };
+    let _span_guard = tracing::span!(tracing::Level::INFO, "libcnb-detect").entered();
 
-    #[cfg(not(feature = "trace"))]
-    let mut trace_error = |_: &dyn std::error::Error| {};
+    let trace_error = |error: &Error<<B as Buildpack>::Error>| {
```
edmorley, line 141: Sorry I missed that this would then cause unused variable lint warnings iff the trace feature isn't enabled. (Because when used in a workspace feature consolidation or whatever it's called means the example fixture enables the feature regardless.)

The warnings can be seen in this CI packaging test:
https://github.com/heroku/libcnb.rs/actions/runs/13639589720/job/38126210529#step:10:27

To avoid these we could either go back to using `#[cfg(not(feature = "trace"))]`, like so:

```rust
    #[cfg(feature = "trace")]
    let trace_error =
        |error: &Error<<B as Buildpack>::Error>| tracing::error!(?error, "libcnb-detect-error");
    #[cfg(not(feature = "trace"))]
    let trace_error = |_: &Error<<B as Buildpack>::Error>| {};
```

Or add a Clippy-allow (though this would turn off unused variable warnings for all of `trace_error`).

joshwlewis: Re-introduced the feature-conditional definition in [a108c18](https://github.com/heroku/libcnb.rs/pull/918/commits/a108c188662cadb679ff33d17987496f06c88b9f).

PR #925, opened 2025-03-18
Use `--locked` with Cargo where possible

edmorley:
The Cargo `--locked` argument ensures that Cargo will fail with an error if `Cargo.lock` is out of sync with `Cargo.toml`, rather than the lockfile being silently updated.

Since this repo is a library, we currently don't commit its `Cargo.lock` to Git (though we might want to revisit that for a few reasons in the future).

However, there are still some places where we should be using `--locked` that aren't already.

GUS-W-18062544.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - libcnb-package:
-  - The `cargo build` command used when packing the buildpack is now run using `--locked` when the `CI` env var is set. ([#924](https://github.com/heroku/libcnb.rs/pull/924))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- libcnb-package:
+  - The `cargo build` command used when packing the buildpack is now run using `--locked` when the `CI` env var is set. ([#924](https://github.com/heroku/libcnb.rs/pull/924))
```
Malax, line 13:
```suggestion
  - The `cargo build` command used when packing the buildpack is now run using `--locked` when the `CI` env var is set. ([#925](https://github.com/heroku/libcnb.rs/pull/925))
```

edmorley: Thank you!

Issue #899, opened 2025-01-10
Bug when prepending to PATH twice on the same layer


schneems:
## Expected

If I prepend PATH with `b` and then prepend it with `a` that the outcome will be `PATH=b:a:$PATH`.

## Actual

Only the last modification is preserved, it ends up being `PATH=b:$PATH`.

## Repro

```
$ cargo new env_delimiter_apply
$ cd env_delimiter_apply
$ cargo add libcnb
$ cargo add pretty_assertions
```

Stick [this code in your `main.rs`]() and run tests:

```
$ cargo test
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests src/main.rs (target/debug/deps/env_delimiter_apply-487764a010d26946)

running 2 tests
test test::test_simple_path ... ok
test test::test_double_path ... FAILED

failures:

---- test::test_double_path stdout ----
thread 'test::test_double_path' panicked at src/main.rs:46:9:
assertion failed: `(left == right)`

Diff < left / right > :
<PATH=b:a
>PATH=a
```

colincasey:
I've been hit by this as well (see #790). My workaround is [here](https://github.com/heroku/buildpacks-deb-packages/blob/main/src/install_packages.rs#L408-L425).

edmorley:
So I agree this is confusing, but it's currently working as designed.

libcnb.rs is a thin layer over the underlying CNB spec. The CNB spec doesn't have a way to say "I want to set multiple values for this delimited field, please concatenate them with the delimiter for me".

We could of course change the libcnb.rs API here (there are other parts of libcnb that try and smooth over usability issues in the underlying spec), though it may end up meaning we have to complicate the API or make things confusing in other ways.

schneems:
I disagree that the spec upstream has anything to say about the affordance over what goes into the output. i.e. it doesn't say if you should use it like:


```Bash
# Replace
echo bar > PATH.prepend
echo foo > PATH.prepend
```

Or

```bash
# Prepend
echo "foo:$(cat PATH.prepend)" > PATH.prepend
```

But also, whether we agree or disagree on this point is a bit of a tangent to my overall goals.

> We can change this behavior today without changing the current API. The only problem  (i can forsee) would be is it would be a breaking change for anyone using the prepend modification behavior to indicate "replace this" or "delete/clear" this.

[...]

PR #935, opened 2025-04-22
Failing test for #899

schneems:


edmorley:
This is currently working as designed in libcnb.rs. So this isn't a bug, but a request to change the API/design.

See:
https://github.com/heroku/libcnb.rs/issues/899#issuecomment-2820246034

schneems:
I responded in that other thread, restating here: If we cannot iterate on this interface directly, we should deprecate it in favor of one with fewer footguns.

I opened up this PR #935 and then another PR with another failing test #936 and didn't notice that I made another mistake: forgetting to add a delimiter. I just fixed it in: https://github.com/heroku/libcnb.rs/pull/936/commits/7a5ff485603e55e65619f72359cd2388bf1bb6a1. There is an RFC to change the upstream default https://github.com/buildpacks/rfcs/pull/326, however, I don't think we need to wait for it. I think we should prefer "correctness" of programmer intent and productivity over "correctness of spec" unless doing so would violate major expectations or cause clear "one way door" type pain.

Issue #534, opened 2022-12-01
Export `layer_env::PATH_LIST_SEPARATOR`


edmorley:
Currently when setting `ModificationBehavior::Delimiter` in a buildpack (eg for a path based env var), one has to either:
1. Hardcode `:` as the delimeter in the buildpack
2. Reimplement the platform specific definitions here: https://github.com/heroku/libcnb.rs/blob/dadd5b51fcc7c32e63bc0256132304a66ddff96f/libcnb/src/layer_env.rs#L921-L925

We should export them instead.

Also, those constants should probably be moved higher up in that file, since currently they are hidden after the tests.

Malax:
I like this idea! 👍🏻

edmorley:
I think we should maybe focus on fixing the default behaviour upstream instead of this - see:
https://github.com/buildpacks/spec/issues/285

PR #945, opened 2025-05-13
Change default PATH delimiter

schneems:
The default PATH delimiter is `` (empty string) rather than the OS delimiter. This has caused many issues and heartache https://github.com/buildpacks/spec/issues/285.

The upstream spec does not state the delimiter that should be used when none is provided. This change suggests that we should default to `:` for unix and `;` for windows. This can be seen as a dry-run for the spec proposal change https://github.com/buildpacks/rfcs/pull/326. If this change can be made with little to no reported problems, we can send that information upstream.

Other suggestions have been floated https://github.com/heroku/libcnb.rs/issues/534. I suggest that we prioritize the most common case, which for Heroku buildpacks is using the OS path separator.


libcnb/src/layer_env.rs
```diff
@@ -482,7 +483,7 @@ impl LayerEnvDelta {
         self.entries
             .get(&(ModificationBehavior::Delimiter, key.into()))
             .cloned()
-            .unwrap_or_default()
+            .unwrap_or(OsString::from(PATH_LIST_SEPARATOR))
```
schneems, line 486: Problem - This code is used in two ways: Writing env var changes to CNB layers and reading env var changes to CNB layers. 

We can make assumptions or change defaults about the writing because we "own" the interface and expectations there. However, we cannot change expectations about reading as we don't know who wrote it. We could work through/around this decoupling this logic somehow either on write or read. 

For example, on read if the delim doesn't exist, deserialize that as an explicit `""` (empty string) until the upstream spec is updated, and never write a layer without a `.delim` (by defaulting to `;`).

colincasey:
> what if we changed the type of ModificationBehavior::{Append,Prepend} so they accepted the delimiter instead?

Making the delimiter required and explicit seems like a reasonable compromise. 

It would also be nice if the APIs for append and prepend modifications took a list of values. Which, now that I think of it, what if we extended your idea @edmorley to lift the `name` and `value` parameters of the `insert` function into each `ModificationBehavior` enum?

```rust
pub enum ModificationBehavior {
    Append { delimiter: OsString, name: OsString, values: Vec<OsString> },
    Default  { name: OsString, value: OsString },
    Delimiter { value: OsString },
    Override { name: OsString, value: OsString },
    Prepend { delimiter: OsString, name: OsString, values: Vec<OsString> },
}
```

PR #955, opened 2025-08-04
Update otel crates 

schneems:
The dependabot failed to compile https://github.com/heroku/libcnb.rs/pull/932

### DIFF
--- a/libcnb/src/tracing.rs
+++ b/libcnb/src/tracing.rs
@@ -140,7 +140,7 @@ impl<W: Write + Send + Debug> FileExporter<W> {
 
 impl<W: Write + Send + Debug> SpanExporter for FileExporter<W> {
     fn export(
-        &mut self,
+        &self,
         batch: Vec<opentelemetry_sdk::trace::SpanData>,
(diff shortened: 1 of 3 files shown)


libcnb/src/tracing.rs
```diff
@@ -139,28 +138,21 @@ impl<W: Write + Send + Debug> FileExporter<W> {
 }
 
 impl<W: Write + Send + Debug> SpanExporter for FileExporter<W> {
-    fn export(
```
edmorley, line 142: I was curious about what changed upstream to need this change, and I found:
https://github.com/open-telemetry/opentelemetry-rust/commit/dc9a5c87d67453e59f274f80335b3c18e08f78de

Looking at that, it seems the implementation can be simplified further by switching to native `async` syntax, so I copied the migration patterns seen in the upstream PR and pushed a commit with those changes:
https://github.com/heroku/libcnb.rs/pull/955/commits/afd2fe17ebfac85c63402247c8316604dcb1c499

edmorley (approved):

edmorley:
Thank you for updating these!

PR #959, opened 2025-09-22
Change `libherokubuildpack::download::download_file` backend to `reqwest`, support `rustls-tls-native-roots`

Malax:
Adding `Platform certificate store support` from #958 without adding more than absolutely necessary. It's not a strict replacement of that PR. But platform certificate store support is something we immediately need now that can be implemented with a few lines of code. It's still not ideal to depend on `tokio` if there otherwise is no need for it. I consider this PR a stopgap and not a final solution.

Ref: GUS-W-19678535

### DIFF
--- a/libherokubuildpack/Cargo.toml
+++ b/libherokubuildpack/Cargo.toml
@@ -50,7 +50,7 @@ tar = { version = "0.4.44", default-features = false, optional = true }
 termcolor = { version = "1.4.1", optional = true }
 thiserror = { version = "2.0.12", optional = true }
 toml = { workspace = true, optional = true }
-reqwest = { version = "0.12.23", features = ["blocking", "rustls-tls-native-roots"], optional = true }
+reqwest = { version = "0.12.23", default-features = false, features = ["blocking", "rustls-tls-native-roots"], optional = true }
 
 [dev-dependencies]
 serde_test = "1.0.177"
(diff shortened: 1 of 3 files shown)


edmorley (approved):

Malax:
Thanks for the test @colincasey! I could not get it to work with `ureq` at this point. I moved this PR to use `reqwest`'s blocking API and ported the test over. Good catch! :)

I'm not super happy with this PR, but given that this download function should be removed sooner rather than later and we need a quick stopgap solution I think it's alright.

libherokubuildpack/src/download.rs
```diff
@@ -2,15 +2,18 @@ use std::{fs, io};
 
 #[derive(thiserror::Error, Debug)]
 pub enum DownloadError {
-    // Boxed to prevent `large_enum_variant` errors since `ureq::Error` is massive.
```
Malax, line 5: Note: no longer necessary since `reqwest` wraps their errors inside `reqwest::Error` in a `Box` already.

libherokubuildpack/Cargo.toml
```diff
@@ -50,8 +50,9 @@ tar = { version = "0.4.44", default-features = false, optional = true }
 termcolor = { version = "1.4.1", optional = true }
 thiserror = { version = "2.0.12", optional = true }
 toml = { workspace = true, optional = true }
-ureq = { version = "3.0.12", default-features = false, features = ["rustls"], optional = true }
+reqwest = { version = "0.12.23", features = ["blocking", "rustls-tls-native-roots"], optional = true }
```
edmorley, line 53: > I could not get it to work with `ureq` at this point.

Could you expand on this?

Was there a bug with ureq that meant native TLS wasn't being honoured? If so, should we file something upstream?

I'm fine with landing this as a stop-gap in the meantime prior to writing up a bug report - but we should definitely file a bug if there is one (and there's no explanation as to the rough reason so far in the PR?)

colincasey (approved):

PR #989, opened 2026-03-26
 Add `inventory::schedule` module for tracking release lifecycle

runesoerensen:
Buildpacks that ship an `Inventory` of versioned artifacts often need to track when those versions reach end-of-life, as well as other lifecycle metadata, so they can warn users. Today each buildpack has to roll its own implementation, or encode lifecycle data in `Artifact` metadata - which may not be ideal when the information spans a range of artifacts rather than belonging to any single one. This PR adds a small, generic `schedule` sub-module inside `inventory` that provides shared plumbing for it.

This is an evolution of the approach first explored in [heroku/buildpacks-nodejs#1332](https://github.com/heroku/buildpacks-nodejs/pull/1332), extracted into `libherokubuildpack` so it can be reused across buildpacks. A minimal "reference"/POC implementation in the .NET CNB can be found here: [heroku/buildpacks-dotnet#404](https://github.com/heroku/buildpacks-dotnet/pull/404)

This is just a proposal for a possible solution. Feedback is more than welcome!

GUS-W-21778596

### What it adds

`Schedule<R, E, M>` holds a list of `Release<R, E, M>` entries:

```rust
pub struct Release<R, E, M> {
    pub requirement: R,   // which versions this release covers
    pub end_of_life: E,   // when this release line reaches EOL
    pub metadata: M,      // arbitrary extra data
}
```

`Schedule::resolve(&version)` finds the first release whose `requirement` satisfies the given version, using the existing `VersionRequirement<V>` trait from `inventory::version`.

`Schedule` and `Release` derive `Serialize` and `Deserialize`. `Schedule::resolve()` requires `R: VersionRequirement<V>` for version matching - like the `semver::VersionReq` does when the `inventory-semver` feature is enabled. Because any `VersionRequirement` implementor automatically satisfies `ArtifactRequirement` (via a blanket impl), a release's `requirement` can also be passed directly to `Inventory::resolve` to find matching artifacts within that release line.

**`E`** (`end_of_life`) is also generic - the module doesn't pull in a date library or define what "end-of-life" means. Consumers choose their own type (e.g. `time::Date`, `String`, `bool`, a custom enum).

**`M`** (`metadata`) follows the same pattern as `Artifact` - use it for additional release information such as support tier (LTS/STS), channel name, support/release dates, etc. Use `Option<()>` to omit it from serialization entirely.

TOML serialization and deserialization is provided via `FromStr` and `Display`, following the same patterns as `Inventory`.

### Example

Building a schedule and serializing it to TOML:

```rust
let mut schedule = Schedule::<VersionReq, String, Option<()>>::new();
schedule.push(Release {
    requirement: VersionReq::parse("^1.0").unwrap(),
    end_of_life: "2025-01-01".to_string(),
    metadata: None,
});

let toml = schedule.to_string();
assert_eq!(toml, r#"[[releases]]
requirement = "^1.0"
end_of_life = "2025-01-01"
"#);
```

Parsing a schedule from TOML and resolving a release for a version:

```rust
let schedule: Schedule<VersionReq, String, Option<()>> = toml.parse().unwrap();
let release = schedule.resolve(&Version::new(1, 2, 3)).unwrap();
assert_eq!(release.end_of_life, "2025-01-01");
```

Using a schedule alongside an inventory:

```rust
// Resolve an artifact from the inventory as usual
let artifact = inventory.resolve(os, arch, &version_req).unwrap();

// Find the release that covers this artifact's version
let release = schedule.resolve(&artifact.version).unwrap();

// Use end_of_life to warn about EOL versions
if today >= release.end_of_life {
    warn!("{} has reached end-of-life", release.requirement);
}

// Use the release's requirement to find the latest artifact in the same release
if let Some(latest) = inventory.resolve(artifact.os, artifact.arch, &release.requirement) {
    if latest.version > artifact.version {
        info!(
            "Newer version available: {} (released {})",
            latest.version, latest.metadata.release_date
        );
    }
}
```

A working implementation using this module is available in [heroku/buildpacks-dotnet#404](https://github.com/heroku/buildpacks-dotnet/pull/404).

### Design notes and limitations

- **No ordering or uniqueness constraints.** `Schedule::resolve` returns the first `Release` whose requirement satisfies the version. Consumers are responsible for ensuring releases are in the desired order and that requirements don't overlap ambiguously.
- **Pre-release versions are not matched by default.** Common `VersionRequirement` implementations (including `semver::VersionReq`) do not match pre-release versions unless the specific pre-release is named in the requirement. This is generally desirable - pre-releases aren't considered supported releases. Consumers needing pre-release tracking can implement custom `VersionRequirement` types.

### Other changes

Moves the shared `VersionRequirement<String> for String` test impl from `artifact::tests` to `version::tests` so both `artifact` and `schedule` tests reuse it without duplication.

### DIFF
--- a/libherokubuildpack/src/inventory/schedule.rs
+++ b/libherokubuildpack/src/inventory/schedule.rs
@@ -1,0 +1,180 @@
+use crate::inventory::version::VersionRequirement;
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+/// Represents a release schedule for a set of version requirements.
+///
+/// A schedule maps version requirements to release lifecycle data such as end-of-life
+/// dates. It can be used by a buildpack to determine support status for a resolved version.
+///
+/// The schedule can be manipulated in-memory and then re-serialized to disk to facilitate
+/// both reading and writing schedule files.
+///
+/// # Example
+///
+/// ```rust
+/// use libherokubuildpack::inventory::schedule::{Schedule, Release};
+/// use semver::{Version, VersionReq};
+///
+/// // Create a release and add it to a schedule
+/// let new_release = Release {
+///     requirement: VersionReq::parse("^1.0").unwrap(),
+///     end_of_life: "2025-01-01".to_string(),
+///     metadata: None,
+/// };
+/// let mut schedule = Schedule::<VersionReq, String, Option<()>>::new();
+/// schedule.push(new_release.clone());
+///
+/// // Serialize the schedule to TOML
+/// let schedule_toml = schedule.to_string();
+/// assert_eq!(
+///     r#"[[releases]]
+/// requirement = "^1.0"
+/// end_of_life = "2025-01-01"
+/// "#,
+///     schedule_toml
+/// );
+///
+/// // Deserialize the schedule from TOML
+/// let parsed_schedule = schedule_toml
+///     .parse::<Schedule<VersionReq, String, Option<()>>>()
+///     .unwrap();
+///
+/// // Resolve a release for a given version
+/// let resolved_release = parsed_schedule.resolve(&Version::new(1, 2, 3)).unwrap();
+/// assert_eq!(resolved_release.end_of_life, "2025-01-01");
+/// ```
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Schedule<R, E, M> {
+    pub releases: Vec<Release<R, E, M>>,
+}
+
+/// A single entry in a [`Schedule`], covering versions that match `requirement`.
+///
+/// Each release carries an end-of-life value that can be used to determine
+/// when a release is no longer supported.
+///
+/// Metadata can be used to store additional information about the release.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Release<R, E, M> {
+    pub requirement: R,
+    pub end_of_life: E,
+    pub metadata: M,
+}
+
+impl<R, E, M> Default for Schedule<R, E, M> {
+    fn default() -> Self {
+        Self { releases: vec![] }
+    }
+}
+
+impl<R, E, M> Schedule<R, E, M> {
+    /// Creates a new empty schedule
+    #[must_use]
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Add a new release to the in-memory schedule
+    pub fn push(&mut self, release: Release<R, E, M>) {
+        self.releases.push(release);
+    }
+
+    /// Return a single release as the first match for the given version.
+    ///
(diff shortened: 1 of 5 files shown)


colincasey (commented):
I believe I can work with this. I'll try and integrate it into my stack of PRs to provide more direct feedback but I don't foresee any issues.

libherokubuildpack/src/inventory/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+/// let parsed_schedule = schedule_toml
+///     .parse::<Schedule<VersionReq, String, Option<()>>>()
+///     .unwrap();
+///
+/// // Resolve a release for a given version
+/// let resolved_release = parsed_schedule.resolve(&Version::new(1, 2, 3)).unwrap();
+/// assert_eq!(resolved_release.end_of_life, "2025-01-01");
+/// ```
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Schedule<R, E, M> {
```
colincasey, line 50: Found one minor issue here. If I want `E` to be a `time::Date` which may be useful since we can leverage TOML dates directly, I need to use a wrapper type with `toml_datetime_compat` as you need to use `#[serde(with =
  "toml_datetime_compat")]` on the field to use it.

runesoerensen: Ah yes you can enable the `serde-human-readable` feature on the `time` dependency to serialize the value as well -- that's what I did in the .NET implementation https://github.com/heroku/buildpacks-dotnet/pull/404/changes#diff-1442da3d00e03c0212733a9c70b2924b931685cffaffeec3a8039c9e6e423323R27

colincasey: Yeah, but the difference there is that you're not storing the date as actual TOML date in the output.

colincasey (approved):
This works for me. I say let's move forward with it and refine later (if necessary).

libherokubuildpack/src/inventory/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+
+impl<R, E, M> std::fmt::Display for Schedule<R, E, M>
+where
+    R: Serialize,
+    E: Serialize,
+    M: Serialize,
+{
+    #![allow(clippy::unwrap_used)]
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&toml::to_string(self).unwrap())
```
Malax: This should probably be an `expect` instead of `unwrap`, no?

runesoerensen: Replaced in c578003aa52d09cdecefab9e88f664acbc45bac1

runesoerensen:
Superseded by https://github.com/heroku/libcnb.rs/pull/991

PR #991, opened 2026-03-31
Add `inventory::warning` module for version-specific warnings

runesoerensen:
Buildpacks that ship an `Inventory` of versioned artifacts often need to warn users when those versions reach end-of-life, have security advisories, or are otherwise noteworthy. Today each buildpack has to roll its own implementation, or encode this in `Artifact` metadata - which may not be ideal when the information spans a range of artifacts rather than belonging to any single one. This PR adds a small `warning` sub-module inside `inventory` that provides shared plumbing for it.

This is an evolution of the approach proposed in #989, simplified based on feedback. It drops generic end-of-life types and metadata in favor of a plain `message: String` - letting each buildpack own its message format entirely.

A reference implementation in the .NET CNB can be found here: [heroku/buildpacks-dotnet#405](https://github.com/heroku/buildpacks-dotnet/pull/405)

GUS-W-21778596

### What it adds

`VersionWarnings<R>` holds a list of `VersionWarning<R>` entries:

```rust
pub struct VersionWarning<R> {
    pub requirement: R,   // which versions this warning covers
    pub message: String,  // the user-facing warning message
}
```

`VersionWarnings::resolve(&version)` finds **all** warnings whose `requirement` satisfies the given version, using the existing `VersionRequirement<V>` trait from `inventory::version`. This differs from `Inventory::resolve` (which returns the single highest match) — a version may match multiple warnings (e.g. an EOL notice and a security advisory).

`VersionWarnings` and `VersionWarning` derive `Serialize` and `Deserialize`. `resolve()` requires `R: VersionRequirement<V>` for version matching — like `semver::VersionReq` when the `inventory-semver` feature is enabled.

TOML serialization and deserialization is provided via `FromStr` and `Display`, following the same patterns as `Inventory`.

### Example

Building warnings and serializing to TOML:

```rust
let mut warnings = VersionWarnings::<VersionReq>::new();
warnings.push(VersionWarning {
    requirement: VersionReq::parse("^1.0").unwrap(),
    message: "Version 1.x reached end-of-life on 2025-01-01.".to_string(),
});

let toml = warnings.to_string();
assert_eq!(toml, r#"[[warnings]]
requirement = "^1.0"
message = "Version 1.x reached end-of-life on 2025-01-01."
"#);
```

Parsing warnings from TOML and resolving for a version:

```rust
let warnings: VersionWarnings<VersionReq> = toml.parse().unwrap();
let matched = warnings.resolve(&Version::new(1, 2, 3));
assert_eq!(matched.len(), 1);
assert_eq!(matched[0].message, "Version 1.x reached end-of-life on 2025-01-01.");
```

Using warnings alongside an inventory:

```rust
// Resolve an artifact from the inventory as usual
let artifact = inventory.resolve(os, arch, &version_req).unwrap();

// Find all warnings that apply to this artifact's version
for warning in warnings.resolve(&artifact.version) {
    warn!("{}", warning.message);
}
```

### Design decisions

- **`message: String` instead of generic `end_of_life: E` + `metadata: M`.** The previous `Schedule` approach had three generic parameters. In practice, consumers just format a user-facing string anyway. Letting the message author (typically an inventory-updater script) compose the full text is simpler and more flexible — warnings aren't limited to EOL; they can express security notices, deprecation hints, migration guides, etc.
- **Returns all matches, not just the first.** A version could match multiple warning rules. The caller can always take the first if they only want one, and an empty `Vec` cleanly represents "no warnings" without needing `Option`.
- **Single generic parameter `R`.** Consistent with how the rest of the inventory module is generic over version/requirement types, but down from 3 generics to 1.
- **No ordering or uniqueness constraints.** Consumers are responsible for ensuring warnings are in the desired order and that requirements don't overlap ambiguously.
- **Pre-release versions are not matched by default.** Common `VersionRequirement` implementations (including `semver::VersionReq`) do not match pre-release versions unless the specific pre-release is named in the requirement.

### Other changes

Moves the shared `VersionRequirement<String> for String` test impl from `artifact::tests` to `version::tests` so both `artifact` and `warning` tests reuse it without duplication.


Malax (approved):

colincasey:
I prefer #989 to this. That version not only provided a mechanism for resolving information on a release line but, due to it's API, it could be extended to:
- represent the Node.js release schedule in a form that allowed all release information to be leveraged
- at the buildpack level, control how to present messages to the user that are more informative

This PR feels like a step backwards in terms of functionality and flexible to the buildpack author.

PR #1002, opened 2026-05-14
Add LLVM coverage support for integration tests

colincasey:
## Summary

- Add `BuildConfig::enable_coverage()` and `LIBCNB_COVERAGE=1` env var support to `libcnb-test` for collecting LLVM source-based code coverage from integration tests running inside Docker containers
- When enabled, buildpack binaries are compiled with `-C instrument-coverage`, a volume mount extracts `.profraw` files from the container, and `LLVM_PROFILE_FILE` directs output to a known location on the host
- Add `VolumeMount` struct and `--volume` support to `PackBuildCommand`
- Includes integration test verifying profraw files are produced

W-22491168

### DIFF
--- a/libcnb-cargo/src/package/command.rs
+++ b/libcnb-cargo/src/package/command.rs
@@ -163,15 +163,16 @@ fn eprint_pack_command_hint(
 }
 
 fn eprint_compiled_buildpack_success(current_dir: &Path, target_dir: &Path) {
-    let size_string = calculate_dir_size(target_dir)
-        .map(|size_in_bytes| {
+    let size_string = calculate_dir_size(target_dir).map_or(
+        String::from("<unknown>"),
+        |size_in_bytes| {
             // Precision will only be lost for sizes bigger than 52 bits (~4 Petabytes), and even
             // then will only result in a less precise figure, so is not an issue.
             #[allow(clippy::cast_precision_loss)]
             let size_in_mib = size_in_bytes as f64 / (1024.0 * 1024.0);
             format!("{size_in_mib:.2}")
-        })
-        .unwrap_or(String::from("<unknown>"));
+        },
+    );
 
     let relative_output_path =
         pathdiff::diff_paths(target_dir, current_dir).unwrap_or_else(|| target_dir.to_path_buf());
--- a/libcnb-cargo/src/package/command.rs
+++ b/libcnb-cargo/src/package/command.rs
@@ -163,16 +163,14 @@ fn eprint_pack_command_hint(
 }
 
 fn eprint_compiled_buildpack_success(current_dir: &Path, target_dir: &Path) {
-    let size_string = calculate_dir_size(target_dir).map_or(
-        String::from("<unknown>"),
-        |size_in_bytes| {
+    let size_string =
+        calculate_dir_size(target_dir).map_or(String::from("<unknown>"), |size_in_bytes| {
             // Precision will only be lost for sizes bigger than 52 bits (~4 Petabytes), and even
             // then will only result in a less precise figure, so is not an issue.
             #[allow(clippy::cast_precision_loss)]
             let size_in_mib = size_in_bytes as f64 / (1024.0 * 1024.0);
             format!("{size_in_mib:.2}")
-        },
-    );
+        });
 
     let relative_output_path =
         pathdiff::diff_paths(target_dir, current_dir).unwrap_or_else(|| target_dir.to_path_buf());
--- a/libcnb-test/src/build.rs
+++ b/libcnb-test/src/build.rs
@@ -12,13 +12,29 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::{fs, io};
 
+/// Merges additional env vars into the cargo build env, appending to existing values with a space
+/// separator (e.g. for RUSTFLAGS) or inserting new entries.
+fn merge_cargo_env(
+    cargo_build_env: &mut Vec<(OsString, OsString)>,
+    additional: &[(OsString, OsString)],
+) {
+    for (key, value) in additional {
+        if let Some(existing) = cargo_build_env.iter_mut().find(|(k, _)| k == key) {
+            existing.1.push(" ");
+            existing.1.push(value);
+        } else {
+            cargo_build_env.push((key.clone(), value.clone()));
+        }
+    }
+}
+
 /// Packages the current crate as a buildpack into the provided directory.
 pub(crate) fn package_crate_buildpack(
     cargo_profile: CargoProfile,
     target_triple: impl AsRef<str>,
     cargo_manifest_dir: &Path,
     target_buildpack_dir: &Path,
-    coverage: bool,
+    additional_cargo_env: &[(OsString, OsString)],
 ) -> Result<PathBuf, PackageBuildpackError> {
     let buildpack_toml = cargo_manifest_dir.join("buildpack.toml");
 
@@ -37,7 +53,7 @@ pub(crate) fn package_crate_buildpack(
         target_triple,
         cargo_manifest_dir,
         target_buildpack_dir,
-        coverage,
+        additional_cargo_env,
     )
 }
 
@@ -47,7 +63,7 @@ pub(crate) fn package_buildpack(
     target_triple: impl AsRef<str>,
     cargo_manifest_dir: &Path,
     target_buildpack_dir: &Path,
-    coverage: bool,
+    additional_cargo_env: &[(OsString, OsString)],
 ) -> Result<PathBuf, PackageBuildpackError> {
     let mut cargo_build_env = match cross_compile_assistance(target_triple.as_ref()) {
         CrossCompileAssistance::HelpText(help_text) => {
@@ -59,15 +75,7 @@ pub(crate) fn package_buildpack(
         CrossCompileAssistance::Configuration { cargo_env } => cargo_env,
     };
 
-    if coverage {
-        let coverage_flag = OsString::from("-C instrument-coverage");
-        if let Some(existing) = cargo_build_env.iter_mut().find(|(k, _)| k == "RUSTFLAGS") {
-            existing.1.push(" ");
-            existing.1.push(&coverage_flag);
-        } else {
-            cargo_build_env.push((OsString::from("RUSTFLAGS"), coverage_flag));
-        }
-    }
+    merge_cargo_env(&mut cargo_build_env, additional_cargo_env);
 
     let workspace_root_path = find_cargo_workspace_root_dir(cargo_manifest_dir)
         .map_err(PackageBuildpackError::FindCargoWorkspaceRoot)?;
--- a/libcnb-test/src/build.rs
+++ b/libcnb-test/src/build.rs
@@ -12,27 +12,6 @@ use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::{fs, io};
 
-#[derive(Clone, Debug, PartialEq)]
-pub(crate) struct CargoEnvAddition {
-    pub(crate) key: OsString,
-    pub(crate) value: OsString,
-    pub(crate) separator: OsString,
-}
-
-fn apply_cargo_env_additions(
-    cargo_build_env: &mut Vec<(OsString, OsString)>,
-    additions: &[CargoEnvAddition],
-) {
-    for addition in additions {
-        if let Some(existing) = cargo_build_env.iter_mut().find(|(k, _)| k == &addition.key) {
-            existing.1.push(&addition.separator);
-            existing.1.push(&addition.value);
-        } else {
-            cargo_build_env.push((addition.key.clone(), addition.value.clone()));
-        }
-    }
-}
-
 /// Packages the current crate as a buildpack into the provided directory.
 pub(crate) fn package_crate_buildpack(
     cargo_profile: CargoProfile,
@@ -148,6 +127,27 @@ pub(crate) enum PackageBuildpackError {
     GetDependencies(GetDependenciesError<BuildpackId>),
     #[error(transparent)]
     PackageBuildpack(libcnb_package::package::PackageBuildpackError),
+}
+
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) struct CargoEnvAddition {
+    pub(crate) key: OsString,
+    pub(crate) value: OsString,
+    pub(crate) separator: OsString,
+}
+
+fn apply_cargo_env_additions(
+    cargo_build_env: &mut Vec<(OsString, OsString)>,
+    additions: &[CargoEnvAddition],
+) {
+    for addition in additions {
+        if let Some(existing) = cargo_build_env.iter_mut().find(|(k, _)| k == &addition.key) {
+            existing.1.push(&addition.separator);
+            existing.1.push(&addition.value);
+        } else {
+            cargo_build_env.push((addition.key.clone(), addition.value.clone()));
+        }
+    }
 }
 
 #[cfg(test)]
--- a/libcnb-test/src/build.rs
+++ b/libcnb-test/src/build.rs
@@ -144,3 +144,53 @@ pub(crate) enum PackageBuildpackError {
     #[error(transparent)]
     PackageBuildpack(libcnb_package::package::PackageBuildpackError),
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn merge_cargo_env_inserts_into_empty() {
+        let mut env = Vec::new();
+        merge_cargo_env(
+            &mut env,
+            &[(
+                OsString::from("RUSTFLAGS"),
+                OsString::from("-C opt-level=3"),
+            )],
+        );
+        assert_eq!(env.len(), 1);
+        assert_eq!(env[0].0, "RUSTFLAGS");
+        assert_eq!(env[0].1, "-C opt-level=3");
+    }
+
+    #[test]
+    fn merge_cargo_env_inserts_new_key_alongside_existing() {
+        let mut env = vec![(OsString::from("CC"), OsString::from("clang"))];
+        merge_cargo_env(
+            &mut env,
+            &[(OsString::from("RUSTFLAGS"), OsString::from("-C lto"))],
+        );
+        assert_eq!(env.len(), 2);
+        assert_eq!(env[0], (OsString::from("CC"), OsString::from("clang")));
+        assert_eq!(
+            env[1],
+            (OsString::from("RUSTFLAGS"), OsString::from("-C lto"))
+        );
+    }
+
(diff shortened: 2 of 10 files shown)


libcnb-test/src/pack.rs
```diff
@@ -1,7 +1,29 @@
...
+    fn from(mount: &VolumeMount) -> Self {
+        let mut s = OsString::from(&mount.source);
+        s.push(":");
+        s.push(&mount.target);
+        if let Some(ref opts) = mount.options {
+            s.push(":");
+            s.push(opts);
+        }
+        s
+    }
```
Malax: `impl From<&VolumeMount> for OsString` is a misleading abstraction. `OsString` has no canonical representation of a `VolumeMount` — the format here is specific to the `pack --volume` CLI grammar (`source:target[:options]`), not a general conversion. A reader who encounters `OsString::from(volume)` at the call site has to chase the impl to learn that it's pack-CLI-specific.

With exactly one caller, I'd suggest inlining the construction in `From<PackBuildCommand> for Command` and dropping this impl. The inline form ends up a few lines shorter as well.

libcnb-test/src/pack.rs
```diff
@@ -1,7 +1,29 @@
...
 use std::path::PathBuf;
 use std::process::Command;
 
+/// Represents a volume mount for Docker containers.
+#[derive(Clone, Debug)]
+pub(crate) struct VolumeMount {
+    pub(crate) source: PathBuf,
+    pub(crate) target: PathBuf,
+    pub(crate) options: Option<String>,
+}
```
Malax: `PackBuildCommand` is the central type in this file, but `VolumeMount` is defined above it. A reader skimming the file is led with a small support type before the thing it supports. I'd move `VolumeMount` (and its `From` impl, if it stays) below `PackBuildCommand` so the file reads top-down from the primary type.

libcnb-test/src/test_runner.rs
```diff
@@ -114,6 +121,12 @@ impl TestRunner {
             pack_command.env(key, value);
         });
 
+        let additional_cargo_env: Vec<CargoEnvUpdate> = if coverage_enabled {
```
Malax: The parameter/binding name `additional_cargo_env` and the type name `CargoEnvUpdate` describe the same concept under two different names. One calls them "env (additions)", the other calls them "updates". The annotation `Vec<CargoEnvUpdate>` reads as "a list of env updates", which clashes with the `_env` suffix on the binding.

I'd pick one vocabulary and use it everywhere. Either rename the type to `CargoEnv` or `CargoEnvAddition` to match the binding, or rename the binding to `cargo_env_updates` to match the type. The mismatch also shows up on the new `additional_cargo_env` parameter of `package_crate_buildpack` and `package_buildpack`.

libcnb-test/src/build.rs
```diff
@@ -8,15 +8,50 @@ use libcnb_package::dependency_graph::{GetDependenciesError, get_dependencies};
...
 
+/// Describes how to merge an env var into the cargo build environment.
+#[derive(Clone, Debug, PartialEq)]
+pub(crate) enum CargoEnvUpdate {
+    /// Append the value to any existing value using the given separator (e.g. `" "` for RUSTFLAGS).
+    Append {
+        key: OsString,
+        value: OsString,
+        separator: OsString,
+    },
```
Malax: `CargoEnvUpdate` is an enum with a single variant. The enum shape suggests there are multiple kinds of updates to dispatch on, but `apply_cargo_env_updates` has nothing to dispatch over. A struct with named fields would express the same data more directly:

```rust
pub(crate) struct CargoEnvAppend {
    pub key: OsString,
    pub value: OsString,
    pub separator: OsString,
}
```

If a second operation (e.g. overwrite, prepend) is added later, that's the right time to introduce the enum. Until then the variant adds a layer of pattern matching with no payoff.

libcnb-test/src/test_runner.rs
```diff
@@ -176,6 +191,42 @@ impl TestRunner {
...
+        std::fs::canonicalize(&dir).expect("Failed to canonicalize coverage output directory");
+
+    pack_command.volume(VolumeMount {
+        source: dir,
+        target: PathBuf::from("/tmp/llvm-cov"),
+        options: Some(String::from("rw")),
+    });
+    // %p = PID, %m = binary signature hash — prevents clobbering across concurrent runs.
+    // See: https://doc.rust-lang.org/rustc/instrument-coverage.html
+    pack_command.env("LLVM_PROFILE_FILE", "/tmp/llvm-cov/%p-%m.profraw");
```
Malax: The container path `/tmp/llvm-cov` is hardcoded twice: once as the `target` of the volume mount, once as the prefix of `LLVM_PROFILE_FILE`. The two values must agree for coverage output to land in the mounted directory, but nothing in the code expresses that dependency. A `const COVERAGE_CONTAINER_DIR: &str = "/tmp/llvm-cov";` (or similar) used in both places would make the relationship explicit and prevent silent drift if one site is later edited.

libcnb-test/src/build_config.rs
```diff
@@ -13,6 +13,7 @@ pub struct BuildConfig {
     pub(crate) target_triple: String,
     pub(crate) builder_name: String,
     pub(crate) buildpacks: Vec<BuildpackReference>,
+    pub(crate) coverage: bool,
```
Malax: Two things on the naming here:

1. The field is `coverage: bool`, the setter is `enable_coverage()`, and the local in `test_runner.rs` is `coverage_enabled`. Three names for the same boolean. `coverage` on its own reads as a noun (where are the coverage results?) rather than a flag.

2. More fundamentally, this PR doesn't produce coverage. It enables LLVM source-based instrumentation so that running the buildpack emits `.profraw` files. Coverage analysis happens later, off the data we collect here, with `llvm-profdata` / `grcov` / `cargo-llvm-cov`. Naming the field, env var, and method around "coverage" overstates what's being toggled.

I'd lean toward `instrumentation_enabled` / `enable_instrumentation()` / `LIBCNB_INSTRUMENTATION` (or `..._COVERAGE_INSTRUMENTATION` if you want to keep "coverage" discoverable). That matches the rustc flag name (`-C instrument-coverage`) and makes it accurate that this captures inputs, not results.

libcnb-cargo/src/package/command.rs
```diff
@@ -163,15 +163,14 @@ fn eprint_pack_command_hint(
...
+    let size_string =
+        calculate_dir_size(target_dir).map_or(String::from("<unknown>"), |size_in_bytes| {
             // Precision will only be lost for sizes bigger than 52 bits (~4 Petabytes), and even
             // then will only result in a less precise figure, so is not an issue.
             #[allow(clippy::cast_precision_loss)]
             let size_in_mib = size_in_bytes as f64 / (1024.0 * 1024.0);
             format!("{size_in_mib:.2}")
-        })
-        .unwrap_or(String::from("<unknown>"));
+        });
```
Malax, line 173: This change is unrelated to the PR's stated purpose (LLVM coverage support in `libcnb-test`). It's a pre-existing `clippy::map_unwrap_or` cleanup in `libcnb-cargo`. Ideally it would land as its own PR so the history stays bisectable and each PR has a single subject. Not blocking; just calling it out.

libcnb-test/src/test_runner.rs
```diff
@@ -176,6 +191,42 @@ impl TestRunner {
...
+
+    vec![CargoEnvUpdate::Append {
+        key: OsString::from("RUSTFLAGS"),
+        value: OsString::from("-C instrument-coverage"),
+        separator: OsString::from(" "),
+    }]
+}
```
Malax: This function is doing too many things in one place. Two issues:

1. **Three output channels.** It mutates `pack_command` (`.volume()` + `.env()`), performs filesystem side effects (mkdir, chmod, canonicalize), and returns `Vec<CargoEnvUpdate>`. A reader has to track all three to understand what the call site does. See the separate comment about returning data instead of taking `&mut PackBuildCommand`.

2. **Inconsistent error handling.** Rest of `test_runner.rs` uses `unwrap_or_else(|error| panic!("Error doing X: {error}"))`. Here it's a mix: `unwrap_or_else(panic!)` for the workspace root, `.expect("Failed to ...")` for the fs calls. Pick one style and apply it.

libcnb-test/src/test_runner.rs
```diff
@@ -176,6 +191,42 @@ impl TestRunner {
     }
 }
 
+fn configure_coverage(
+    cargo_manifest_dir: &Path,
+    pack_command: &mut PackBuildCommand,
+) -> Vec<CargoEnvUpdate> {
```
Malax: Taking `&mut PackBuildCommand` here is a code smell. The function uses it only to push a `VolumeMount` and an env var, while it already returns the `RUSTFLAGS` addition as data. That asymmetry is what makes the function read as "doing many things at once".

Consider returning all three pieces as data:

```rust
struct CoverageSetup {
    volume: VolumeMount,
    pack_env: (String, String),
    cargo_env: Vec<CargoEnvUpdate>, // or whichever name lands from the other thread
}

fn configure_coverage(cargo_manifest_dir: &Path) -> CoverageSetup { ... }
```

The caller then applies the mutations in one place. The function becomes pure-ish (still has the mkdir/chmod side effect, but that's it), and tests can assert on the returned struct instead of inspecting a mutated builder.

libcnb-test/src/build.rs
```diff
@@ -8,15 +8,50 @@ use libcnb_package::dependency_graph::{GetDependenciesError, get_dependencies};
...
+            } => {
+                if let Some(existing) = cargo_build_env.iter_mut().find(|(k, _)| k == key) {
+                    existing.1.push(separator);
+                    existing.1.push(value);
+                } else {
+                    cargo_build_env.push((key.clone(), value.clone()));
+                }
+            }
+        }
+    }
```
Malax: The file now opens with `CargoEnvUpdate` and `apply_cargo_env_updates` before the central API (`package_crate_buildpack` / `package_buildpack`). A reader skimming the file is led with a small support type and its private helper before the thing they support. Same shape as the `VolumeMount` ordering in `pack.rs`.

Two ways out, depending on how prominent this type stays:

1. Move the type and helper below the public packaging functions so the file reads top-down from the primary API.
2. If `CargoEnvUpdate` keeps growing (the unit tests already add a fair number of LoC), pull it into its own module (e.g. `cargo_env.rs`) so `build.rs` stays focused on packaging.

libcnb-test/src/test_runner.rs
```diff
@@ -67,6 +70,10 @@ impl TestRunner {
     ) {
         let config = config.borrow();
 
+        let coverage_enabled = config.coverage
+            || env::var("LIBCNB_COVERAGE")
+                .is_ok_and(|v| matches!(v.to_ascii_lowercase().as_str(), "1" | "true" | "yes"));
```
Malax: A few problems packed into three lines:

1. **Ad-hoc truthy vocabulary.** `matches!(... "1" | "true" | "yes")` invents its own set of accepted values. Why is `"yes"` in but `"on"` out? Why not `"y"`? Either pin to exactly `"1"`/`"true"` (rustc/cargo convention) or use `bool::from_str`. Beyond this one site, the parsing belongs in a central helper. As soon as a second feature reads a boolean env var, the two sites will drift on which strings are truthy and a developer will hit it. Lift it now (e.g. `fn parse_bool_env(name: &str) -> Option<bool>`) so the project has one canonical answer.

2. **Silent drop of malformed values.** `is_ok_and` collapses "var not set" and "var set to garbage" into the same `false`. A user typing `LIBCNB_COVERAGE=truee` gets a quietly broken test run with no signal. Malformed values should panic with a clear message.

3. **Untested logic buried in `build()`.** The parsing has nothing to do with running a build. Even before extracting a generic helper, lifting this into a small named function (e.g. `fn coverage_enabled(config: &BuildConfig) -> bool`) makes it unit-testable in isolation. The `build()` method is already long and gets more so with each feature like this.

See also the naming comment on `BuildConfig::coverage` for the broader "is this really 'coverage' or is it 'instrumentation'" question, which would also rename `LIBCNB_COVERAGE` if you take that path.

Malax:
A higher-level naming concern that runs through the whole PR: this work doesn't actually produce coverage. It enables LLVM source-based **instrumentation** so that running buildpacks emits `.profraw` files. Turning those files into coverage data happens later with `llvm-profdata` / `grcov` / `cargo-llvm-cov`, none of which is in scope here.

Naming the field, env var, method, fixture, and CHANGELOG entry around "coverage" overstates what the feature does and conflates capture with analysis. I'd lean toward "instrumentation" everywhere user-visible. If "coverage" is kept for discoverability, something like `coverage_instrumentation` would at least be accurate.

Malax (changes requested):

libcnb-test/src/pack.rs
```diff
@@ -1,7 +1,29 @@
...
+use std::ffi::OsString;
 use std::path::PathBuf;
 use std::process::Command;
 
+/// Represents a volume mount for Docker containers.
+#[derive(Clone, Debug)]
+pub(crate) struct VolumeMount {
+    pub(crate) source: PathBuf,
+    pub(crate) target: PathBuf,
+    pub(crate) options: Option<String>,
```
Malax: `options: Option<String>` punts the entire `pack --volume` options grammar onto the caller as a free-form string. Pack's volume options are a small, known set (`ro`, `rw`, plus mount-time flags), and they're joined with `,` when there are multiple. Encoding them as `Option<String>` means:

- Callers have to remember the comma-joining and the exact keywords.
- Typos like `Some(String::from("rww"))` compile fine and fail at runtime inside `pack`.
- The only current call site already passes `Some(String::from("rw"))` for what is conceptually "read-write", which is the default.

A typed enum captures both the constraint and the intent:

```rust
#[derive(Clone, Debug)]
pub(crate) enum VolumeMountMode {
    ReadOnly,
    ReadWrite,
}
```

Then `options: Option<String>` becomes `mode: VolumeMountMode` (with a sensible default), and the `From<&VolumeMount>` / inline construction renders the right keyword. If more options are needed later, extend the enum or split into a struct of typed flags. Either way, `String` is the wrong type here.

libcnb-test/src/pack.rs
colincasey: Fixed — inlined the construction in the `From<PackBuildCommand> for Command` impl: https://github.com/heroku/libcnb.rs/commit/d0fc777facd735008ebd918435f84c74b573acad

libcnb-test/src/pack.rs
colincasey: Fixed — moved VolumeMount below PackBuildCommand: https://github.com/heroku/libcnb.rs/commit/0328714201cc4c212d8b8b21628fa69d59ed670f

libcnb-test/src/test_runner.rs
colincasey: Fixed — unified vocabulary to `CargoEnvAddition` / `cargo_env_additions` everywhere: https://github.com/heroku/libcnb.rs/commit/dad8b961bd9b2c00f2df72ae2516a5ee266a63ee

libcnb-test/src/build.rs
colincasey: Fixed — converted to a struct with named fields: https://github.com/heroku/libcnb.rs/commit/dad8b961bd9b2c00f2df72ae2516a5ee266a63ee

libcnb-test/src/test_runner.rs
colincasey: Fixed — extracted `INSTRUMENTATION_CONTAINER_DIR` constant used in both places: https://github.com/heroku/libcnb.rs/commit/8a41142a5bbb2d34ba073e156b9b3cd28f14dd0a

libcnb-test/src/build_config.rs
colincasey: Fixed — renamed to `enable_instrumentation()` / `LIBCNB_INSTRUMENTATION` / `instrumentation_enabled` throughout: https://github.com/heroku/libcnb.rs/commit/0006bbb450cee3382bb2eb0e515e14e8fd836b67

libcnb-cargo/src/package/command.rs
colincasey: Agreed this is unrelated. It's only present here because CI clippy checks would fail without it. I'll extract it to a separate PR against main with just the clippy fix, and rebase this branch on top of that once merged.

libcnb-test/src/test_runner.rs
colincasey: Fixed — `configure_instrumentation` now returns an `InstrumentationSetup` struct; the caller applies the mutations. Also extracted `create_buildpack_references` to keep `build_internal` under the line limit, and standardized error handling to `unwrap_or_else(|error| panic!(...))`: https://github.com/heroku/libcnb.rs/commit/293b4d18299ec3fcf21b2d2bbe43e97d0dca6c82

libcnb-test/src/test_runner.rs
colincasey: Fixed in the same commit — function no longer takes `&mut PackBuildCommand`: https://github.com/heroku/libcnb.rs/commit/293b4d18299ec3fcf21b2d2bbe43e97d0dca6c82

libcnb-test/src/build.rs
colincasey: Fixed — moved CargoEnvAddition and apply_cargo_env_additions below the packaging functions: https://github.com/heroku/libcnb.rs/commit/70b8138cf9bda562a2851bbb0ab91b52cf0f3ba8

libcnb-test/src/test_runner.rs
colincasey: Fixed — pinned to `"1"/"true"` (rust/cargo convention), removed `"yes"`. Malformed values now return an `InvalidInstrumentationEnvVar` error (caller panics with the Display message). Parsing is extracted to `instrumentation_enabled_via_env()` for testability: https://github.com/heroku/libcnb.rs/commit/55ab256b37512eefe8f8b128605148d3256ca97c

libcnb-test/src/pack.rs
colincasey: Fixed — replaced `options: Option<String>` with a typed `VolumeMountMode` enum (`ReadOnly`/`ReadWrite`). Testing confirmed the mode must be explicit — omitting it caused profraw files not to be written: https://github.com/heroku/libcnb.rs/commit/6bc26f6

libcnb-cargo/src/package/command.rs
edmorley: This was fixed in #1004

edmorley:
Cancelling review to stop it appearing in the Slack reminders list. If this PR is still wanted, it needs a rebase so the unrelated Clippy changes can be removed etc.

